PR #1, opened 2018-02-02
First version of the heat exchanger OF-CCX tutorial.

MakisH:
At the moment, I use the case prepared with SimScale. In order to not polute the repository, I kept the large files (fluid meshes and the complete solid case) in my LRZ Sync+Share account and I provide a script to download them from there.

Eventually I want to recreate the case, maybe with another geometry, if needed. Currently we don't have the original geometry files.

### DIFF
--- a/CHT/heat_exchanger/buoyantSimpleFoam-CalculiX/Allclean
+++ b/CHT/heat_exchanger/buoyantSimpleFoam-CalculiX/Allclean
@@ -1,0 +1,59 @@
+#!/bin/sh
+cd ${0%/*} || exit 1    # Run from this directory
+
+echo "Cleaning..."
+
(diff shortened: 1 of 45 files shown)


uekerman:
@MakisH I just also gave you collaboration rights on the repo. 
Please put the existing SU2-CCX tutorial in another subfolder and update the readme. Thanks!

@shkodm @kirmartynov @sltmyr Please then check if all your links (wiki) still work

MakisH:
Done! I don't see any links from the wiki to specific files, so everything looks good.

Could you please add a nice description, keywords, and license, as you have for the rest of the preCICE repositories? A link to precice.org or the main repository would be very good.

PR #10, opened 2018-09-25
FEniCS tutorial + adapter draft

BenjaminRueth:
In this example I coupled a partitioned heat equation via preCICE. The discretization is realized via two instances of FEniCS. A draft for a FEniCS adapter is provided in this example, as well.


BenjaminRueth:
We decided to move this contribution into a repository for a FEniCS adapter (https://github.com/precice/fenics-adapter). The tutorial will be part of the adapter repository as an example use case.

uekerman:
A few consistency remarks (Not sure if this here is the best place):
 *  `exchange_data` should better be called `advance` and should also include the functionality from `check_convergence`
 * I don't like the package name `Coupling` too much. Wouldn't `Adapter` or anything with preCICE and adapter be better?
 * The following code could be moved into a `initialize` function:
```
coupling.set_coupling_mesh(mesh, coupling_boundary, coupling_mesh_name)
    coupling.set_read_field(read_data_name, u_D_function)
    coupling.set_write_field(write_data_name, f_N_function)
elif problem is ProblemType.NEUMANN:
    coupling.set_read_field(read_data_name, f_N_function)
    coupling.set_write_field(write_data_name, u_D_function)
```

PR #14, opened 2018-12-04
Openfoam-calculix perp_flap added

derekrisseeuw:
The perpendicular flap tutorial


derekrisseeuw:
Added the cylinder with a flap tutorial in the second commit.

MakisH:
Thank you very much for the rigorous work on these cases, Derek!

I am trying to run these (either using the `Allrun` or the `runSolid`) script and in both cases I get the following error from CalculiX (2.13):

```
*ERROR in readinput: cannot open file Solid/all.msh
```

Am I missing some file or step?

derekrisseeuw:
Hi Makis, 
Is the Allrun file already working for you?
The first step of the solid part in the Allrun file is to prepare the mesh, which is done by this line:
`cgx -bg pre_flap.fbd > prepare_flap.log 2>&1`
Could you state the log file please? I guess there is something going wrong in the preprocessing. 

What is your version of ccx, I tested with 2.13?

MakisH:
Apparently I have CGX built as `cgx_2.13`, while the script tries to find `cgx`. I will document this. Both cases seem to be running normally for me! I will get into the details tomorrow.

derekrisseeuw:
Ah, that's indeed a good rermark, I had not thought of that!
Maybe it's good to add a comment of this in the Allrun file.

MakisH:
When trying to run any of the two tutorials with the (older) OpenFOAM 4.1, I get the following error:

```
Selecting dynamicFvMesh dynamicMotionSolverFvMesh
Selecting motion solver: displacementLaplacian


--> FOAM FATAL IO ERROR: 
keyword displacementLaplacianCoeffs is undefined in dictionary "./Fluid/constant/dynamicMeshDict"

file: ./Fluid/constant/dynamicMeshDict from line 19 to line 36.

    From function const Foam::dictionary& Foam::dictionary::subDict(const Foam::word&) const
    in file db/dictionary/dictionary.C at line 642.
```

Is this a hint that we need something more in the `constant/dynamicMeshDict` or is it only a compatibility issue?

MakisH:
Seems to work for all versions with this, thanks! I hope the syntax will not become deprecated very soon.

derekrisseeuw:
I share your concern! However, I think we're in the clear here. 
There are also motionsolvers which allow multiple motions. For these the `coeffs` always will be used in the dictionary.

MakisH:
Side-note for the future reader: the `cylinderFlap` case contributed here is almost but not exactly the Turek&Hron FSI3 benchmark case. For this, one would need a parabolic inlet profile, for which one would need to use groovyBC, an extension provided by [swak4Foam](https://openfoamwiki.net/index.php/Contrib/swak4Foam).

MakisH:
Accidental merge, more improvements to come. But it's mostly ready.

PR #19, opened 2019-01-21
SSI tutorial with CalculiX

Alex-tru:
This is a ready-to-run tutorial for a simple case of structure-structure interaction within a single beam.


KyleDavisSA:
The simulation runs and seems to provide the same results when run on my side. I just can't run it with the runTutorial.sh file (or I am using it wrong?). I just used my own Allrun script (see attached).

[Allrun.txt](https://github.com/precice/tutorials/files/2866737/Allrun.txt)

PR #24, opened 2019-03-18
OpenFOAM-Nutils CHT tutorial

uekerman:
A first OpenFOAM-Nutils tutorial.

I copied the "flow over a heated plate" tutorial from the OpenFOAM-OpenFOAM variant and adjusted a few things:
* The width in z direction is now much smaller (0.05 instead of 0.4). The big aspect ratio of the original case made me feel uncomfortable. You still might see small oscillation in the temperature coming from the mapping. My tests showed that Nutils needs at least two layers of elements in z-direction such that the mapping works properly.  
* There is no `Allrun` script as I think that this does not help to understand how preCICE works (it still might be a valid standard for pure OpenFOAM cases). 
* Both solvers are moved into sub-folders. Only the coupling happens in the root directory.    
* The heat equation has only one physical constant: the heat conductivity, for which I use `k` and which is set to 100. `rho` and `c_p` are both 1. We, Gertjan an me, had problems to understand why two different variables are used for the diffusivity and the flux computation in, e.g., the FEniCS variant of this tutorial. See also [wikipedia](https://en.wikipedia.org/wiki/Heat_equation). 

I will move the content of the README.md to the precice wiki, once we are settled with this PR (but before merging). I just left it here for the moment that everybody can also comment on the description.

Reviewers:
@BenjaminRueth Please test the case and look at the physics. Any idea why things are different in FEniCS and here?
@MakisH Anything that should be added in the description?
@gertjanvanzwieten Please check if we can still improve anything in the Nutils script. 

Thanks already everybody!

I will merge in the end.


CHT/flow-over-plate/buoyantPimpleFoam-nutils/README.md
```diff
@@ -0,0 +1,66 @@
...
+Terminal 1:
+```
+$ cd OpenFOAM
+$ ./runFOAM
+```
+
+Terminal 2:
+```
+$ cd Nutils
+$ python3 cht.py
```
MakisH: Again, for consistency, I would create a script `runSolid` that executes these commands.

Even though it may be more educative to ask the user to do all the steps, repeating the tutorial (as a test case) with just one script, instead of looking at the documentation, is very very helpful for us.

CHT/flow-over-plate/buoyantPimpleFoam-nutils/README.md
```diff
@@ -0,0 +1,66 @@
...
+For running this tutorial, you further have to install
+
+* **preCICE**, see [preCICE wiki](https://github.com/precice/precice/wiki/Building).
+* **Python bindings**, see [bindings readme](https://github.com/precice/precice/tree/develop/src/precice/bindings/python)
+* **OpenFOAM**, see [Notes on OpenFOAM](https://github.com/precice/openfoam-adapter/wiki/Notes-on-OpenFOAM).
+* **OpenFOAM adapter**, see [OpenFOAM adapter wiki](https://github.com/precice/openfoam-adapter/wiki/Building). If you have problems compiling, see the [troubleshooting section](https://github.com/precice/precice/wiki/CHT-with-OpenFOAM-and-FEniCS#troubleshooting) below.
+
+### Testing your installation
+
+* **OpenFOAM and OpenFOAM adapter:** To make sure that everything is working properly, you should run the following tutorial case: https://github.com/precice/openfoam-adapter/wiki/Tutorial-for-CHT:-Flow-over-a-heated-plate.
```
MakisH:
```suggestion
* **OpenFOAM and OpenFOAM adapter:** To make sure that everything is working properly, you should run the [similar OpenFOAM-OpenFOAM tutorial case](https://github.com/precice/openfoam-adapter/wiki/Tutorial-for-CHT:-Flow-over-a-heated-plate).
```

CHT/flow-over-plate/buoyantPimpleFoam-nutils/precice-config.xml
```diff
@@ -0,0 +1,61 @@
...
+
+  <data:scalar name="Temperature"/>
+  <data:scalar name="Heat-Flux"/>
+
+  <mesh name="OpenFOAM-Mesh">
+    <use-data name="Temperature"/>
+    <use-data name="Heat-Flux"/>
+  </mesh>
+
+  <mesh name="MeshNutils-GP">
```
MakisH: Maybe rename this (and the following) to `Nutils-Mesh-GP`? I understand why you want to use the solver names as mesh names here, but let's make it consistent inside the file at least.

CHT/flow-over-plate/buoyantPimpleFoam-nutils/README.md
uekerman: I'll add the script and put both variants in the README.md

uekerman:
After reconsideration, I will also add an `Allrun` script, but not make it the first option in the documentation. But for testing, it makes a lot of sense.

BenjaminRueth:
I created a [branch](https://github.com/precice/tutorials/tree/validate_flux_computation), where we can investigate the flux computation issue. I would propose to not care about the issue (#25) in this pull request.

uekerman:
> I am not so happy with the code duplication of /tutorials/CHT/flow-over-plate/buoyantPimpleFoam-nutils/OpenFOAM. In the FEniCS example I request the user to simply copy the code from the OpenFOAM repository. This is a bit more complicated for the user, but more clean as well from my perspective.

I agree that there is a problem, but the best solution is not obvious to me. @MakisH and me wanted to have a detailed look into this next week.
 
 > In the existing flow-over-plate tutorials, we use Solid and Fluid for naming of the folders and participants (in precice-config.xml). Here, you use Nutils and OpenFOAM. Is there a reason?

Yes. Also already discussed with @MakisH. I had the use case in mind that somebody familiar with Nutils wants to get to know preCICE. Then, naming the participants after the solvers and not after the physics is the better choice. I agree though that consistency is a problem. Also on the agenda for next week's discussion.

BenjaminRueth:
> > I am not so happy with the code duplication of /tutorials/CHT/flow-over-plate/buoyantPimpleFoam-nutils/OpenFOAM. In the FEniCS example I request the user to simply copy the code from the OpenFOAM repository. This is a bit more complicated for the user, but more clean as well from my perspective.
> 
> I agree that there is a problem, but the best solution is not obvious to me. @MakisH and me wanted to have a detailed look into this next week.

This is related to #28.

PR #27, opened 2019-04-10
Tutorial for Structure/Structure coupling with FEniCS

BenjaminRodenberg:
# What do we want in the end?

* we want a tutorial similar to the already existing [heat-heat coupling with FEniCS](https://github.com/precice/tutorials/tree/master/HT/partitioned-heat/fenics-fenics). Here, we want to split a beam into two partitions to solve the linear elasticity problem (more information in the [FEniCS tutorial book, p.50ff](https://fenicsproject.org/pub/tutorial/pdf/fenics-tutorial-vol1.pdf))
* this tutorial can be used to test the exchange of vector values quantities via the fenics adapter. See https://github.com/precice/fenics-adapter/issues/26

# Current State

* [ ] Implement (pseudo-)time loop to solve static problem. See Comment below
* [ ] Use advance in (pseudo-)time loop.
* [ ] Resolve https://github.com/precice/fenics-adapter/issues/31, we need this for applying zero traction on the outer boundary while applying non-zero traction on the coupling boundary.

# Pseudo Time Loop 

Linear elasticity is a static problem, which is generally not handled by preCICE, but can be circumvented by reformulating the problem as `laplacian(u) - f = du/dt`. When `du/dt = 0` (= convergence), the static problem `laplacian(u) = f` is solved


uekerman:
As discussed with @BenjaminRueth on the phone: 

I think FSI would be the better choice here than SSI:
* FSI is much more relevant to our users. 
* SSI has still some numerical issues that we have not yet fully resolved / understood. See the master thesis of @Alex-tru (not online afaik, I can send an offline copy if somebody needs it). My guess is that it is a similar issue than the flux computation for heat-heat coupling. 
* For FSI, you only have "half of the work" as you "only" need to read forces and write displacements, but not the other way around as well.   

So, I would recommend to approach FSI first and maybe SSI afterwards if needed.

BenjaminRueth:
I added the `wontfix` label and we can leave this PR open to indicate that this PR is `postponed`.

BenjaminRodenberg:
This PR is very outdated. Rewriting the setup based on https://github.com/precice/tutorials/tree/master/partitioned-elastic-beam and adapting https://github.com/precice/tutorials/tree/master/perpendicular-flap/solid-fenics makes more sense.

PR #38, opened 2019-07-09
FEniCS OpenFOAM perp-flap and cyl-flap

richahert:
I implemented the perpendicular flap with FEniCS and OpenFOAM. 
It comes with a new Fluid Case that is almost the same as for OF-CcX but since FEniCS is in 2D I changed the thickness to 1.

There is still some code in perp-flap.py that is not relevant for a tutorial that can be deleted.

The displacements are smaller than expected for unknown reasons. 

It makes sense to merge this PR together with https://github.com/precice/fenics-adapter/pull/38, the PR brings FSI functionality to the fenics-adapter.

# ToDos: perp-flap

- [x] make fluid (in XZ plane) and solid case (in XY plane) consistent (solved by 2837129)
- [x] fix VTK output (in the end we want to obtain identical number of samples for both simulation and a nice visualization in paraview), see https://github.com/precice/tutorials/pull/38#issuecomment-517323896, opened issue https://github.com/precice/openfoam-adapter/issues/95

# ToDos: cyl-flap

- [ ] ~~test case runs with OpenFOAM 5 or older~~
- [x] test case runs with OpenFOAM 6 or newer (see #40)


BenjaminRueth:
With the version of the fenics-adapter on [this branch](https://github.com/precice/fenics-adapter/tree/richahert-fsi-pr) I can run the example and the results look reasonable. There are some open points, though:

* The 2D-3D coupling is not handled properly. While the Fluid "lives" in the XZ plane, the Solid "lives" in XY plane. The coupling seems to work properly, therefore this is mainly a visualization issue.
* Some information in the `README.md` on expected output and hints for visualization would be good. You can also use the wiki for this. I already created a new page [here](https://github.com/precice/precice/wiki/Tutorial-for-FSI-with-FEniCS-and-OpenFOAM).

BenjaminRueth:
> 
> The 2D-3D coupling is not handled properly. While the Fluid "lives" in the XZ plane, the Solid "lives" in XY plane. The coupling seems to work properly, therefore this is mainly a visualization issue.
> 

Regarding this I added https://github.com/richahert/fenics-adapter/commit/2c9c2a0baff6d71823d3c76afe7d4b8164027bb6. This removes the entire swapping of coordinates from the adapter. I think we should try to fix the test cases instead. This means that quasi-2D OpenFOAM simulations should not live in the XZ plane, but in the XY plane. This makes the adapter much cleaner and also solves the visualization issue mentioned above.

BenjaminRueth:
There are a few minor visualization issues:

---

If I open `Solid/FSI_S/u_fsi.pvd` instead of `Solid/FSI_S/u_fsi*` in paraview, the output does not match. Probably there is some inconsistency between the time in FEniCS and the time in OpenFOAM.

---

If look at the VTK output of the OpenFOAM simulation vs. the FEniCS Simulation I get 50 samples on the FEniCS side (`u_fsi000000.vtu`, `u_fsi000001.vtu`, ... , ``u_fsi000049.vtu`), but 52 samples on the OpenFOAM side (`Fluid_0.vtk`, `Fluid_1.vtk`, `Fluid_10.vtk`, `Fluid_20.vtk`, ..., `Fluid_500.vtk`).

I created the VTK files for the Fluid by running `foamToVTK` inside `OpenFOAM-FEniCS/Fluid`.

Did I do something wrong with respect to visualization? Why is there a File called `Fluid_1.vtk`? This does not look right...

This issue can be solved manually by:

* delete `Fluid_1.vtk` 
* add a line `displacement_out << (u_np1,t)` after `precice.finalize()` in `Solid/perp-flap.py` (I am not sure that this will generally work, but I think in the special case of the tutorial the very last step is being output by OpenFOAM, but FEniCS does not, because `n % 10 == 0:` is evaluated to `false` in the `if` statement)

Then the results look very nice:

last timestep **without** `displacement_out << (u_np1,t)`:  

Screenshot from 2019-08-01 16-51-14

last timestep **with** `displacement_out << (u_np1,t)`:

Screenshot from 2019-08-01 16-54-37

richahert:
Now the physics should be correct. The results for the FSI3 look very much like the deal.ii-OpenFOAM example. deformations are bigger than in the benchmark because of the linear elastic material instead of St. Venant Kirchhoff material in the original FSI3.
fsi3

BenjaminRueth:
>     * delete `Fluid_1.vtk`

see https://github.com/precice/tutorials/pull/38#issuecomment-517323896

is still necessary. I think this is an issue of the adapter. Therefore, let's not worry anymore about this here. I opened an issue: https://github.com/precice/openfoam-adapter/issues/95

The rest of the flap tutorial looks very nice!

BenjaminRueth:
I cannot run the OpenFOAM Fluid case for the FSI3 benchmark using OpenFOAM 6. @richahert did you use OpenFOAM 5 or older? Did this work?

I will try to find out how to fix it for OpenFOAM 6.

BenjaminRueth:
Both tutorials are now running on my machine. See #40 regarding compatibility problems with OpenFOAM6.

FSI/cylinderFlap/OpenFOAM-FEniCS/Fluid/system/controlDict
```diff
@@ -10,8 +10,8 @@ FoamFile
 
 // Note: With OpenFOAM v1806 and OpenFOAM 6, the DyM solvers
 // were marked deprecated and merged into their respective standard solvers.
-application         pimpleFoam;    // OpenFOAM v1806, OpenFOAM 6, or newer
-// application         pimpleDyMFoam; // OpenFOAM v1712, OpenFOAM 5.x, or older
+//application         pimpleFoam;    // OpenFOAM v1806, OpenFOAM 6, or newer
+application         pimpleDyMFoam; // OpenFOAM v1712, OpenFOAM 5.x, or older
```
MakisH: Please always default to the latest supported version (so, here `pimpleFoam`). This would also be consistent with the rest of our tutorials. ;-)

BenjaminRueth: Done: 5c625d859c66a43740cad6bfb16a2d25f4e07d41

PR #46, opened 2019-09-23
Adding complex interface case in HT fenics-fenics tutorial

IshaanDesai:
The current heat transfer fenics-fenics tutorial consists of one case in which the interface is a straight edge parallel to the y-axis. A more generalised case of an interface of circular shape is being added in the same tutorial. User can select between two options `simple-interface` and `complex-interface` to compute the respective cases. The aim of this PR is also to verify the nearest_projection mapping implemented in this PR: https://github.com/precice/fenics-adapter/pull/42

### DIFF
--- a/HT/partitioned-heat/fenics-fenics/heat.py
+++ b/HT/partitioned-heat/fenics-fenics/heat.py
@@ -51,17 +51,13 @@ def determine_gradient(V_g, u, flux):
     solve(a == L, flux)
 
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(description='Solving heat equation for simple or complex interface case')
 parser.add_argument("-d", "--dirichlet", help="create a dirichlet problem", dest='dirichlet', action='store_true')
 parser.add_argument("-n", "--neumann", help="create a neumann problem", dest='neumann', action='store_true')
 parser.add_argument("-g", "--gamma", help="parameter gamma to set temporal dependence of heat flux", default=0.0, type=float)
 parser.add_argument("-a", "--arbitrary-coupling-interface", help="uses more general, but less exact method for interpolation on coupling interface, see https://github.com/precice/fenics-adapter/milestone/1", dest='arbitrary_coupling_interface', action='store_true')
-parser.add_argument("-dl", "--domain-left", help="right part of the domain is being computed", dest='domain_left', action='store_true')
-parser.add_argument("-dr", "--domain-right", help="left part of the domain is being computed", dest='domain_right', action='store_true')
-parser.add_argument("-is", "--simple_interface", help="Interface defined is a straight line", dest='simple_interface', action='store_true')
-parser.add_argument("-ic", "--complex_interface", help="Interface defined is circular", dest='complex_interface', action='store_true')
-parser.add_argument("-dc", "--domain_circular", help="domain inside circular part is being computed", dest='domain_circular', action='store_true')
-parser.add_argument("-dnc", "--domain_rest", help="domain excluding circular part is being computed", dest='domain_rest', action='store_true')
+parser.add_argument("-interface", metavar="interface_type string", type=str, nargs=1, help="Type of geometric interface to be solved: simple or complex", dest='interface')
+parser.add_argument("-domain", metavar='domain_type string', type=str, nargs=1, help="Part of domain being solved. For simple interface the options are left and right, for complex interface the options are circular and rest", dest='domain')
 
 args = parser.parse_args()
 
--- a/HT/partitioned-heat/fenics-fenics/heat.py
+++ b/HT/partitioned-heat/fenics-fenics/heat.py
@@ -32,7 +32,6 @@ from errorcomputation import compute_errors
 from my_enums import ProblemType, DomainPart, Subcyling
 import argparse
 import numpy as np
-import mshr
 from problem_setup import OuterBoundary, CouplingBoundary, get_geometry, get_problem_setup
 
 
@@ -79,9 +78,6 @@ if not (args.dirichlet or args.neumann):
     raise Exception("you have to choose either a dirichlet problem (option -d) or a neumann problem (option -n)")
 
 
-# coupling parameters
-
-
 # Create mesh and define function space
 if problem is ProblemType.DIRICHLET:
     adapter_config_filename = "precice-adapter-config-D.json"
@@ -113,7 +109,9 @@ alpha = 3  # parameter alpha
 beta = 1.3  # parameter beta
 gamma = args.gamma  # parameter gamma, dependence of heat flux on time
 
+print("problem = {}".format(problem))
 domain_part = get_problem_setup(args, problem)
+print("domain_part = {}".format(domain_part))
 mesh = get_geometry(args, domain_part)
 
 V = FunctionSpace(mesh, 'P', 2)
--- a/HT/partitioned-heat/fenics-fenics/heat.py
+++ b/HT/partitioned-heat/fenics-fenics/heat.py
@@ -124,7 +124,10 @@ f_N = Expression('2 * gamma*t*x[0] + 2 * (1-gamma)*x[0] ', degree=1, gamma=gamma
 f_N_function = interpolate(f_N, V)
 
 coupling_boundary = CouplingBoundary()
+coupling_boundary.get_user_input_args(args)
+
 remaining_boundary = OuterBoundary()
+remaining_boundary.get_user_input_args(args)
 
 bcs = [DirichletBC(V, u_D, remaining_boundary)]
 # Define initial value
--- a/HT/partitioned-heat/fenics-fenics/heat.py
+++ b/HT/partitioned-heat/fenics-fenics/heat.py
@@ -33,6 +33,7 @@ from fenicsadapter import Adapter, ExactInterpolationExpression, GeneralInterpol
 from errorcomputation import compute_errors
 import argparse
 import numpy as np
+import mshr
 
 
 class ProblemType(Enum):
@@ -47,8 +48,10 @@ class DomainPart(Enum):
     """
     Enum defines which part of the domain [x_left, x_right] x [y_bottom, y_top] we compute.
     """
-    LEFT = 1  # left part of domain
-    RIGHT = 2  # right part of domain
+    LEFT = 1  # left part of domain in simple interface case
+    RIGHT = 2  # right part of domain in simple interface case
+    CIRCULAR = 3 # circular part of domain in complex interface case
+    REST = 4 # domain excluding circular part of complex interface case
 
 
 class Subcyling(Enum):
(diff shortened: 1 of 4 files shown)


HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -104,10 +58,13 @@ def determine_gradient(V_g, u, flux):
 parser.add_argument("-a", "--arbitrary-coupling-interface", help="uses more general, but less exact method for interpolation on coupling interface, see https://github.com/precice/fenics-adapter/milestone/1", dest='arbitrary_coupling_interface', action='store_true')
 parser.add_argument("-dl", "--domain-left", help="right part of the domain is being computed", dest='domain_left', action='store_true')
```
BenjaminRueth: Replace `-dl`, `-dr`, `-dc`, `-dnc` with `--domain DOMAINTYPE`. Otherwise, all four arguments could be given at the same time. Ideally use the optional `choices=[]` argument of `add_argument` together with a list of names for the different domain parts (might be`Enum.name`).

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -104,10 +58,13 @@ def determine_gradient(V_g, u, flux):
 parser.add_argument("-a", "--arbitrary-coupling-interface", help="uses more general, but less exact method for interpolation on coupling interface, see https://github.com/precice/fenics-adapter/milestone/1", dest='arbitrary_coupling_interface', action='store_true')
```
BenjaminRueth: `-a` should not be needed anymore in this PR: new strategy allows to use the same strategy for all types of coupling boundaries.

This will also allow us to get rid of `interpolation_strategy=interpolation_strategy` below.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -104,10 +58,13 @@ def determine_gradient(V_g, u, flux):
...
 parser.add_argument("-dl", "--domain-left", help="right part of the domain is being computed", dest='domain_left', action='store_true')
 parser.add_argument("-dr", "--domain-right", help="left part of the domain is being computed", dest='domain_right', action='store_true')
+parser.add_argument("-is", "--simple_interface", help="Interface defined is a straight line", dest='simple_interface', action='store_true')
+parser.add_argument("-ic", "--complex_interface", help="Interface defined is circular", dest='complex_interface', action='store_true')
```
BenjaminRueth: Is already determined by `--domain` and can be removed: If `--domain` is `LEFT`or `RIGHT`, our coupling boundary is a `VerticalEdge` and therefore we have a `--simple-interface`. If `--domain` is `CIRCULAR` or `REST`, our coupling boundary is a `CircularBoundary` and therefore a `--complex-interface`.

HT/partitioned-heat/fenics-fenics/my_enums.py
```diff
@@ -0,0 +1,30 @@
...
+
+class ProblemType(Enum):
+    """
+    Enum defines problem type. Details see above.
+    """
+    DIRICHLET = 1  # Dirichlet problem
+    NEUMANN = 2  # Neumann problem
+
+
+class DomainPart(Enum):
```
BenjaminRueth: List of all `DomainPart`s can be used as valid choices for `--domain`above.

HT/partitioned-heat/fenics-fenics/problem_setup.py
```diff
@@ -0,0 +1,122 @@
...
+                return False
+        if self._complex_interface:
+            point = Point(x[0], x[1])
+            if on_boundary and point.distance(midpoint)**2 - radius**2 < tol:
+                return True
+            else:
+                return False
+
+
+def get_problem_setup(args, problem):
```
BenjaminRueth: I am not 100% sure about the difference between `get_geometry`and `get_problem_setup`. Can you combine both functions into a single one? Or is there a good reason for splitting up functionality? A comment might help.

`problem` can be obtained from `args`. Move the corresponding lines from above (70-78 from `heat.py`) into this function.

HT/partitioned-heat/fenics-fenics/my_enums.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Enum defines which part of the domain [x_left, x_right] x [y_bottom, y_top] we compute.
+    """
+    LEFT = 1  # left part of domain in simple interface case
+    RIGHT = 2  # right part of domain in simple interface case
+    CIRCULAR = 3  # circular part of domain in complex interface case
+    REST = 4  # domain excluding circular part of complex interface case
```
BenjaminRueth: `REST` is not really an exact name. Something like `RECTANGLE_WITH_HOLE`? Shorter would be better.

HT/partitioned-heat/fenics-fenics/problem_setup.py
```diff
@@ -0,0 +1,122 @@
...
+            raise Exception("you can only choose to either compute the circular part of the domain (option -dc) or the residual part (option -dnc)")
+        if not (args.domain_circular or args.domain_rest):
+            print("Default domain partitioning for complex interface is used: Circular part of domain is a Dirichlet-type problem; Rest of the domain is a Neumann-type problem")
+            if problem is ProblemType.DIRICHLET:
+                return DomainPart.CIRCULAR
+            elif problem is ProblemType.NEUMANN:
+                return DomainPart.REST
+
+
+def get_geometry(args, domain_part):
```
BenjaminRueth: We actually do not need `args`, here. Then the difference to `get_problem_setup` is also clear: `get_geometry` returns a FEniCS mesh for a given `domain_part, while `get_problem_setup` parses given `args` into a description of the problem.

BenjaminRueth:
https://github.com/precice/tutorials/blob/bf13fc21c422d9dd1634f4cbfea31bbd7740f1b5/HT/partitioned-heat/fenics-fenics/heat.py#L202-L204

This part of the code is a very dirty hack for computing the normal heat flux (normal times heat flux). Of course, this only works for the case with vertical coupling interface. Before these lines of code are replaced with a more general routine, we cannot expect correct results for the complex interface case.

I created an issue: #47.

HT/partitioned-heat/fenics-fenics/my_enums.py
IshaanDesai: Changed the name to RECTANGLE. This is also not perfect but definitely better than REST.

HT/partitioned-heat/fenics-fenics/heat.py
IshaanDesai: So we now you the `GeneralInterpolationExpression` for solving both the simple_interface and complex_interface case ?

HT/partitioned-heat/fenics-fenics/problem_setup.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class OuterBoundary(SubDomain):
+    def get_user_input_args(self, args):
+        self._simple_interface = args.simple_interface
+        self._complex_interface = args.complex_interface
+
+    def inside(self, x, on_boundary):
+        tol = 1E-14
+        if self._simple_interface:
```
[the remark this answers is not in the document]
IshaanDesai: I do not understand how this `if...else` can go into another function definition. We are overloading the FEniCS function `inside` right ? And for that we need a class definition with the parameter `SubDomain` is how I understood it.

HT/partitioned-heat/fenics-fenics/problem_setup.py
IshaanDesai: I changed the if conditions  in `get_geometry` here: https://github.com/precice/tutorials/pull/46/commits/e60b71e7e4680e4a8ff7949b5f26199b1d2d1312 but then I moved the the two functions `OuterBoundary` and `CouplingBoundary` inside `get_geometry` [here](https://github.com/precice/tutorials/pull/46/commits/09ef90ad0f329c40ce566609ce8d0c826ecea730#diff-d4ebb41e76de027cf50f69fbdee263d8R130-R134). For these two functions I need the parameter `args.interface` because that is how I decide how to calculate the boundaries. Can I somehow do this without passing `args` at all ?

BenjaminRueth:
23e450b requires an updated version of the adapter (https://github.com/precice/fenics-adapter/commit/1eed43064ba4695140564bde851a451384ff7b75). The convergence of this updated version is worse than before (see https://travis-ci.org/precice/systemtests/jobs/596793593). This is not expected, since no algorithmic changes were introduced. The tests use the latest version of the tutorials on master (https://github.com/precice/tutorials/commit/3252ea73ef74877a6a8a75bc65bcc348e089c942) - without the new features that are introduced in this PR. We should fix this first.

After that we can continue finalizing/merging this PR. The features introduced in this PR are expected to introduce algorithmic changes and therefore also the convergence behavior can change.

HT/partitioned-heat/fenics-fenics/problem_setup.py
```diff
@@ -0,0 +1,122 @@
...
+import mshr
+
+y_bottom, y_top = 0, 1
+x_left, x_right = 0, 2
+x_coupling = 1.5  # x coordinate of coupling interface
+radius = 0.2
+midpoint = Point(0.5, 0.5)
+
+
+class OuterBoundary(SubDomain):
```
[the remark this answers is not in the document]
BenjaminRueth: Resolved in https://github.com/IshaanDesai/tutorials/commit/94e6b4c35311e3f4b2aae36b00f674942c439b38#diff-0a9e55bd0a94ac0b4956fd2770c6e4cf. Naming is different than suggested above.

HT/partitioned-heat/fenics-fenics/problem_setup.py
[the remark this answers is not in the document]
BenjaminRueth: Resolved in https://github.com/IshaanDesai/tutorials/commit/94e6b4c35311e3f4b2aae36b00f674942c439b38#diff-0a9e55bd0a94ac0b4956fd2770c6e4cf.

HT/partitioned-heat/fenics-fenics/problem_setup.py
BenjaminRueth: Resolved in https://github.com/IshaanDesai/tutorials/commit/94e6b4c35311e3f4b2aae36b00f674942c439b38#diff-0a9e55bd0a94ac0b4956fd2770c6e4cf

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -196,7 +124,10 @@ def determine_gradient(V_g, u, flux):
 f_N_function = interpolate(f_N, V)
 
 coupling_boundary = CouplingBoundary()
+coupling_boundary.get_user_input_args(args)
+
 remaining_boundary = OuterBoundary()
+remaining_boundary.get_user_input_args(args)
```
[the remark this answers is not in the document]
BenjaminRueth: Resolved in https://github.com/IshaanDesai/tutorials/commit/94e6b4c35311e3f4b2aae36b00f674942c439b38#diff-0a9e55bd0a94ac0b4956fd2770c6e4cf

IshaanDesai:
I was unable to reproduce the `nearest-projection` based error reported by @BenjaminRueth. For the time being I have assumed that `nearest-projection` mapping works. A convergence study for the simple and complex interface case however shows that NP mapping shows first-order behaviour. One possibility is the Rbf interpolation used in `GeneralInterpolationExpression` contributes to this error

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -97,69 +51,29 @@ def determine_gradient(V_g, u, flux):
...
-if problem is ProblemType.DIRICHLET:
-    adapter_config_filename = "precice-adapter-config-D.json"
-elif problem is ProblemType.NEUMANN:
-    adapter_config_filename = "precice-adapter-config-N.json"
-
 # for all scenarios, we assume precice_dt == .1
 if subcycle is Subcyling.NONE and not args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
-    error_tol = 10 ** -7  # Error is bounded by coupling accuracy. In theory we would obtain the analytical solution.
+    error_tol = 10 ** -2  # Error is bounded by coupling accuracy. In theory we would obtain the analytical solution.
```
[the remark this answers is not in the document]
IshaanDesai: Resetting the tolerances to original values here: https://github.com/precice/tutorials/pull/46/commits/3eb1632f9c130ae2e52e4de699bdbab34d148554. I had eased the tolerances to see how the `GeneralInterpolationExpression` performs for the `simple_interface` case

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -97,69 +51,29 @@ def determine_gradient(V_g, u, flux):
...
+    # interpolation_strategy = ExactInterpolationExpression
     interpolation_strategy = ExactInterpolationExpression
 elif subcycle is Subcyling.NONE and args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
-    error_tol = 10 ** -3 # error low, if we do not subcycle. In theory we would obtain the analytical solution.
+    error_tol = 10 ** -1  # error low, if we do not subcycle. In theory we would obtain the analytical solution.
```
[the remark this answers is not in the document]
IshaanDesai: Resetting the tolerances to original value here: https://github.com/precice/tutorials/pull/46/commits/3eb1632f9c130ae2e52e4de699bdbab34d148554

PR #55, opened 2019-11-11
Remove extrapolation

uekerman:
Using an extrapolation for implicit coupling has no guaranteed benefit. Sometimes it gives better results, sometimes not. I would not advise beginners to use it. That's why I suggest to remove it from all tutorials. 

**Please note** This might break the system tests as the results now might differ.

@Reviewers: What do you think? How big is the pain to update the reference results for all system tests?


SSI/loaded_beam/CalculiX-CalculiX/precice-config.xml
```diff
@@ -40,37 +40,27 @@
       <m2n:sockets from="Calculix1" to="Calculix2"  exchange-directory="../" distribution-type="gather-scatter"/>
       
       <coupling-scheme:parallel-implicit>
-<!--      <coupling-scheme:parallel-explicit>-->
```
MakisH: This was kind of intentionally left here, so that one can easily see what the differences between explicit and implicit are (and aitken vs IQN). But I see that this is inconsistent and dead code, which could eventually lead to trouble.

Eder-K:
I agree with @MakisH in that it should be easy but a little tedious. I will check how the results are altered using this version.

BenjaminRueth:
In case our systemtests now go crazy: this is good, because we want to get rid of the extrapolation and then we will know where action is needed.

PR #59, opened 2019-12-11
Add parallel run to fe-fe tutorial

BenjaminRueth:
Ideally, we should not add another tutorial, but avoid duplication and just have the parallel run as a new option of the already existing tutorial.

Related to fenics adapter version of https://github.com/precice/fenics-adapter/tree/parallelFEniCS


BenjaminRueth:
There is an update on this PR https://github.com/precice/fenics-adapter/pull/71. Closing this PR, since it is outdated.

PR #62, opened 2019-12-27
WIP: Using mid-level interface design of fenics-adapter

IshaanDesai:
This PR implements design changes to the `HT/paritioned-heat/fenics-fenics` tutorial according to the new mid-level interface proposed in https://github.com/precice/fenics-adapter/pull/59
The main aim of this PR is that the new interface design in the `fenics-adapter` should be reflected in the tutorial by making the workflow more understandable.

### DIFF
--- a/HT/partitioned-heat/fenics-fenics/heat.py
+++ b/HT/partitioned-heat/fenics-fenics/heat.py
@@ -26,13 +26,13 @@ Heat equation with mixed boundary conditions. (Neumann problem)
 
 from __future__ import print_function, division
 from fenics import Function, FunctionSpace, Expression, Constant, DirichletBC, TrialFunction, TestFunction, \
-    File, solve, lhs, rhs, grad, inner, dot, dx, ds, interpolate, VectorFunctionSpace, FacetNormal
-from fenicsadapter import Adapter, ExactInterpolationExpression, GeneralInterpolationExpression
+    File, solve, lhs, rhs, grad, inner, dot, dx, ds, interpolate, VectorFunctionSpace
+from fenicsadapter import Adapter
 from errorcomputation import compute_errors
-from my_enums import ProblemType, Subcyling
+from my_enums import ProblemType, Subcycling
 import argparse
 import numpy as np
-from problem_setup import get_geometry, get_problem_setup, DomainPart
+from problem_setup import get_geometry, get_problem_setup
 
 
 def determine_gradient(V_g, u, flux):
@@ -63,24 +63,25 @@ args = parser.parse_args()
 
 config_file_name = "precice-config.xml"  # TODO should be moved into config, see https://github.com/precice/fenics-adapter/issues/5 , in case file doesnt not exsist open will fail
 
-subcycle = Subcyling.NONE
+subcycle = Subcycling.NONE
+fenics_dt = 0.0
 
 # for all scenarios, we assume precice_dt == .1
-if subcycle is Subcyling.NONE and not args.arbitrary_coupling_interface:
+if subcycle is Subcycling.NONE and not args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
     error_tol = 10 ** -7  # Error is bounded by coupling accuracy. In theory we would obtain the analytical solution.
     # interpolation_strategy = ExactInterpolationExpression
-    interpolation_strategy = ExactInterpolationExpression
-elif subcycle is Subcyling.NONE and args.arbitrary_coupling_interface:
+    # interpolation_strategy = ExactInterpolationExpression
+elif subcycle is Subcycling.NONE and args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
     error_tol = 10 ** -3  # error low, if we do not subcycle. In theory we would obtain the analytical solution.
-    interpolation_strategy = GeneralInterpolationExpression
+    # interpolation_strategy = GeneralInterpolationExpression
     # TODO For reasons, why we currently still have a relatively high error, see milestone https://github.com/precice/fenics-adapter/milestone/1
-elif subcycle is Subcyling.MATCHING:
+elif subcycle is Subcycling.MATCHING:
     fenics_dt = .01  # time step size
     error_tol = 10 ** -2  # error increases. If we use subcycling, we cannot assume that we still get the exact solution.
     # TODO Using waveform relaxation, we should be able to obtain the exact solution here, as well.
-elif subcycle is Subcyling.NONMATCHING:
+elif subcycle is Subcycling.NONMATCHING:
     fenics_dt = .03  # time step size
     error_tol = 10 ** -1  # error increases. If we use subcycling, we cannot assume that we still get the exact solution.
     # TODO Using waveform relaxation, we should be able to obtain the exact solution here, as well.
@@ -91,6 +92,8 @@ gamma = args.gamma  # parameter gamma, dependence of heat flux on time
 
 domain_part, problem = get_problem_setup(args)
 mesh, coupling_boundary, remaining_boundary = get_geometry(domain_part)
+
+adapter_config_filename = None
 
 # Create mesh and define function space
 if problem is ProblemType.DIRICHLET:
@@ -113,14 +116,9 @@ bcs = [DirichletBC(V, u_D, remaining_boundary)]
 u_n = interpolate(u_D, V)
 u_n.rename("Temperature", "")
 
-precice = Adapter(adapter_config_filename, interpolation_strategy=interpolation_strategy)
+precice = Adapter(adapter_config_filename)
 
-if problem is ProblemType.DIRICHLET:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=u_D_function,
-                                    write_field=f_N_function, u_n=u_n)
-elif problem is ProblemType.NEUMANN:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=f_N_function,
-                                    write_field=u_D_function, u_n=u_n)
+precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, u_n=u_n)
 
 dt = Constant(0)
 dt.assign(np.min([fenics_dt, precice_dt]))
@@ -174,15 +172,40 @@ while precice.is_coupling_ongoing():
     # Compute solution u^n+1, use bcs u_D^n+1, u^n and coupling bcs
     solve(a == L, u_np1, bcs)
 
+    state = precice.initialize_solver_state(u_n, t, n)
+
     if problem is ProblemType.DIRICHLET:
         # Dirichlet problem obtains flux from solution and sends flux on boundary to Neumann problem
         determine_gradient(V_g, u_np1, flux)
-        t, n, precice_timestep_complete, precice_dt = precice.advance(flux, u_np1, u_n, t, dt(0), n)
+        precice.write(flux)
     elif problem is ProblemType.NEUMANN:
         # Neumann problem obtains sends temperature on boundary to Dirichlet problem
-        t, n, precice_timestep_complete, precice_dt = precice.advance(u_np1, u_np1, u_n, t, dt(0), n)
+        precice.write(u_np1)
 
-    dt.assign(np.min([fenics_dt, precice_dt]))  # todo we could also consider deciding on time stepping size inside the adapter
+    precice_dt = precice.advance_coupling(dt(0), fenics_dt)
+    dt.assign(precice_dt)
+
+    precice.read()
+
+    precice.update_boundary_condition()
+
+    solver_state_has_been_restored = False
+
+    if precice.is_action_required(precice.action_read_iteration_checkpoint()):
+        assert (not precice.is_timestep_complete())  # avoids invalid control flow
+        precice.restore_solver_state_from_checkpoint(state)
+        solver_state_has_been_restored = True
+    else:
+        precice.advance_solver_state(state, u_np1, dt)
+
+    if precice.is_action_required(precice.action_write_iteration_checkpoint()):
+        assert (not solver_state_has_been_restored)  # avoids invalid control flow
+        assert (precice.is_timestep_complete())  # avoids invalid control flow
+        precice.save_solver_state_to_checkpoint(state)
+
+    precice_timestep_complete = precice.is_timestep_complete()
+
+    _, t, n = state.get_state()
 
     if precice_timestep_complete:
         u_ref = interpolate(u_D, V)
--- a/HT/partitioned-heat/fenics-fenics/my_enums.py
+++ b/HT/partitioned-heat/fenics-fenics/my_enums.py
@@ -18,7 +18,7 @@ class DomainPart(Enum):
     RECTANGLE = 4  # domain excluding circular part of complex interface case
 
 
-class Subcyling(Enum):
+class Subcycling(Enum):
     """
     Enum defines which kind of subcycling is used
     """


HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
         determine_gradient(V_g, u_np1, flux)
-        t, n, precice_timestep_complete, precice_dt = precice.advance(flux, u_np1, u_n, t, dt(0), n)
+        precice.write(flux)
     elif problem is ProblemType.NEUMANN:
         # Neumann problem obtains sends temperature on boundary to Dirichlet problem
-        t, n, precice_timestep_complete, precice_dt = precice.advance(u_np1, u_np1, u_n, t, dt(0), n)
+        precice.write(u_np1)
+
+    # API call to advance coupling, also returns the optimum time step value
+    precice_dt = precice.advance_coupling(dt(0))
```
uekerman: Why not simply `precice.advance(dt(0))`?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
+    dt.assign(np.min([fenics_dt, precice_dt]))
+
+    # read data
+    precice.read()
+
+    precice.update_boundary_condition()
+
+    solver_state_has_been_restored = False
+
+    if precice.is_action_required(precice.restore_to_checkpoint()):
```
uekerman: I would advise to stick to the preCICE naming here, so `read_checkpoint()`. Maybe it's not the very best name, but let's stay consistent, otherwise it's getting confusing.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
+        assert (not precice.is_timestep_complete())  # avoids invalid control flow
+        precice.restore_solver_state_from_checkpoint(state)
+        solver_state_has_been_restored = True
+    else:
+        precice.advance_solver_state(state, u_np1, dt)
```
uekerman: Not sure if it's the best design decision to put this functionality in the adapter. Isn't that a functionality that every FEniCS code already has?

IshaanDesai: Referring to the file: https://github.com/precice/fenics-adapter/blob/develop/fenicsadapter/solverstate.py my understanding is that this functionality was implemented in the adapter to simplify the three types of assignments in FEniCS, using the sign `=`, using `copy` functionality and using `assign` functionality. There is confusion about how these three assigning types work in the sense that some affect the variable only in the local scope and some affect the variable in the global scope.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
+t_new = 0
+n_new = 0
+
 while precice.is_coupling_ongoing():
 
     # Compute solution u^n+1, use bcs u_D^n+1, u^n and coupling bcs
     solve(a == L, u_np1, bcs)
 
+    # Intializing solver state
+    state = precice.initialize_solver_state(u_n, t, n)
```
BenjaminRueth: Putting the word `initialization` inside the `is_coupling_ongoing` loop locks wrong to me. Maybe it is just a matter of rephrasing: Does the name `get_solver_state` also fit? Or can we put the initialization of the state before the loop and rather `update` than `get` the state for every timestep?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
-        t, n, precice_timestep_complete, precice_dt = precice.advance(u_np1, u_np1, u_n, t, dt(0), n)
+        precice.write(u_np1)
+
+    # API call to advance coupling, also returns the optimum time step value
+    precice_dt = precice.advance(dt(0))
+
+    dt.assign(np.min([fenics_dt, precice_dt]))
+
+    # read data
+    precice.read()
```
BenjaminRueth: Currently, the update of the boundary conditions happens very implicitly through `read`. Something like `coupling_bc = precice.read()` would be more explicit. In the end the code would be more verbose and include some duplication, but one would also see what's going on:

```
coupling_function = precice.read()

    bcs = [DirichletBC(V, coupling_function, coupling_boundary), non_coupling_bc]
elif problem is ProblemType.NEUMANN:
    F = non_coupling_F + coupling_function * ds
```

This will require further changes for the [initialization of `F`](https://github.com/precice/tutorials/pull/62/files#diff-333cee8321c434253c5640a5a2ecd185R133), the [initialization of `bc`](https://github.com/precice/tutorials/pull/62/files#diff-333cee8321c434253c5640a5a2ecd185R112) and [how the adapter provides the coupling boundary conditions](https://github.com/precice/tutorials/pull/62/files#diff-333cee8321c434253c5640a5a2ecd185R135-R140). But (based on past experience with the prior adapter design) I think that a more explicit (and verbose) approach will be much more user friendly.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
+    if precice.is_action_required(precice.read_checkpoint()):
+        assert (not precice.is_timestep_complete())  # avoids invalid control flow
+        precice.restore_solver_state_from_checkpoint(state)
+        solver_state_has_been_restored = True
+    else:
+        precice.advance_solver_state(state, u_np1, dt)
+
+    if precice.is_action_required(precice.write_checkpoint()):
+        assert (not solver_state_has_been_restored)  # avoids invalid control flow
+        assert (precice.is_timestep_complete())  # avoids invalid control flow
```
BenjaminRueth: Can we hide these assertions inside the adapter? In the end I do not expect a user to put these kind of security checks into the user code. Additionally, it will make the code more verbose. If we can perform these checks inside the adapter, we can help the user by avoiding invalid control flow.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
 flux = Function(V_g)
 flux.rename("Flux", "")
 
+t_new = 0
+n_new = 0
```
BenjaminRueth: Unused variables?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -169,20 +170,56 @@ def determine_gradient(V_g, u, flux):
...
+
+    # API call to advance coupling, also returns the optimum time step value
+    precice_dt = precice.advance(dt(0))
+
+    dt.assign(np.min([fenics_dt, precice_dt]))
+
+    # read data
+    precice.read()
+
+    precice.update_boundary_condition()
```
BenjaminRueth: See my comment above. I think this is something the user should do explicitly in order to stay in control of the process. Better provide something like a `coupling_function` that the user can use to define the boundary conditions than forcing the user to use the boundary conditions that we support.

HT/partitioned-heat/fenics-fenics/heat.py
BenjaminRueth: Referring to [boundary condition initialization](https://github.com/precice/tutorials/pull/62/files#diff-333cee8321c434253c5640a5a2ecd185R135-R140): I think defining a Dirichlet or Neumann boundary condition should not be a responsibility of the adapter, but of the user. Writing `bcs.append(precice.create_coupling_dirichlet_boundary_condition(V))` or `bcs.append(DirichletBC(V, precice.get_initial_coupling_function(), coupling_boundary))` does not really make a difference, while the second approach is more flexible. Example: A user might use the `coupling_function` to implement a different kind of BC we do not even think of now.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -63,24 +62,24 @@ def determine_gradient(V_g, u, flux):
...
 
 # for all scenarios, we assume precice_dt == .1
-if subcycle is Subcyling.NONE and not args.arbitrary_coupling_interface:
+if subcycle is Subcycling.NONE and not args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
     error_tol = 10 ** -7  # Error is bounded by coupling accuracy. In theory we would obtain the analytical solution.
     # interpolation_strategy = ExactInterpolationExpression
-    interpolation_strategy = ExactInterpolationExpression
-elif subcycle is Subcyling.NONE and args.arbitrary_coupling_interface:
+    # interpolation_strategy = ExactInterpolationExpression
```
BenjaminRueth: Dead code.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -63,24 +62,24 @@ def determine_gradient(V_g, u, flux):
...
+elif subcycle is Subcycling.NONE and args.arbitrary_coupling_interface:
     fenics_dt = .1  # time step size
     error_tol = 10 ** -3  # error low, if we do not subcycle. In theory we would obtain the analytical solution.
-    interpolation_strategy = GeneralInterpolationExpression
+    # interpolation_strategy = GeneralInterpolationExpression
```
BenjaminRueth: Dead code.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -113,14 +114,14 @@ def determine_gradient(V_g, u, flux):
...
 if problem is ProblemType.DIRICHLET:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=u_D_function,
-                                    write_field=f_N_function, u_n=u_n)
+    precice_dt = precice.initialize(coupling_boundary, mesh, u_n, u_D_function, f_N_function)
 elif problem is ProblemType.NEUMANN:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=f_N_function,
-                                    write_field=u_D_function, u_n=u_n)
+    precice_dt = precice.initialize(coupling_boundary, mesh, u_n, f_N_function, u_D_function)
+
+precice.set_interpolation_type("cubic_spline")
```
BenjaminRueth: Can we use an enum instead of a string?

HT/partitioned-heat/fenics-fenics/heat.py
IshaanDesai: This function is now changed to `get_solver_state`.

HT/partitioned-heat/fenics-fenics/heat.py
BenjaminRueth: Let's put this explanation of what we want to achieve with the `state` object and why we need it in a wiki or similar. Then we have some persistent documentation.

HT/partitioned-heat/fenics-fenics/heat.py
BenjaminRueth: More explicit: Don't let the user worry about our `state` object at all (we can still use it internally for checkpointing). Better just have two functions `precice.store_checkpoint(u_n, t, n)` and `u_cp, t, n = precice.retreive_checkpoint()`. The user then has to independently update the state variables like it is common practice in FEniCS, e.g.:
```
if precice.is_action_required(precice.read_checkpoint):
    u_cp, t, n = precice.retreive_checkpoint()
    u_n.assign(u_cp)
else:
    u_n.assign(u_np1)
    t += dt
    n += 1
```
For reference see [Adapter example, implicit coupling](https://github.com/precice/precice/wiki/Adapter-Example#implicit-coupling).

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -171,20 +191,41 @@ def determine_gradient(V_g, u, flux):
 
 while precice.is_coupling_ongoing():
 
+    if precice.is_action_required(precice.action_write_checkpoint()):
+        precice.store_checkpoint(u_n, t, n)
+
+    # read data
+    precice.read()
```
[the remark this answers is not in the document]
IshaanDesai: I have tried to implement the idea of handling boundary conditions explicitly here: https://github.com/precice/tutorials/pull/62/commits/68c439212a59bc13562cea2f724530cbcdb58752 I am still running into some errors but does the basic idea seem okay?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -195,9 +250,12 @@ def determine_gradient(V_g, u, flux):
         ref_out << u_ref
         error_out << error_pointwise
 
-    # Update dirichlet BC
+    # Update Dirichlet BC
     u_D.t = t + dt(0)
     f.t = t + dt(0)
 
+    # Trial: Reset coupling function
+    coupling_function = None
```
BenjaminRueth, line 258: This line could be removed. Or do we have to use `coupling_expression` instead?

IshaanDesai:
The following tutorials have been modified:
* `CHT/flow-over-plate/buoyantPimpleFoam-fenics/`
* `FSI/cylinderFlap/OpenFOAM-FEniCS/`
* `FSI/flap_perp/OpenFOAM-FEniCS/`
* `HT/partitioned-heat/fenics-fenics/`
to use the new design of the `fenics-adapter` here: https://github.com/precice/fenics-adapter/pull/59

The correctness of the tutorial modifications are tested by:
* Comparing the QN iterations on both participants for each tutorial with the results from the old `fenics-adapter` related tutorial setups. **Found to be identical**
* For the `HT/partitioned-heat/fenics-fenics/` tutorial a direct comparison to the analytical solution via the method of manufactured solutions was used to test accuracy.
* For FSI and CHT tutorials physical values of simulation results were compared and were found to be in near perfect agreement with old results.

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
```diff
@@ -12,44 +12,43 @@
...
 
-
-#create Function Space
+# create Function Space
 V = VectorFunctionSpace(mesh, 'P', 2)
 
-#BCs
-tol=1E-14
+# BCs
+tol = 1E-14
```
BenjaminRueth, line 51: Here, we are using `1E-14` instead of `10**-14` might be a nice solution for our whitespace "problem" above. Like I said: I am not aware of any guidelines which one of the two notations is preferrable.

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
```diff
@@ -63,176 +62,188 @@ def Neumann_Boundary(x, on_boundary):
...
-#alpha method parameters
+bc = DirichletBC(V, Constant((0, 0)), clamped_boundary)
 
+# alpha method parameters
 alpha_m = Constant(0.2)
 alpha_f = Constant(0.4)
-gamma   = Constant(0.5+alpha_f-alpha_m)
-beta    = Constant((gamma+0.5)**2/4.)
+gamma = Constant(0.5 + alpha_f - alpha_m)
+beta = Constant((gamma + 0.5) ** 2 / 4.)
```
BenjaminRueth, line 91:
```suggestion
beta = Constant((gamma + 0.5)**2 / 4.)
```

Whitespace around `**` looks a bit odd.

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
```diff
@@ -63,176 +62,188 @@ def Neumann_Boundary(x, on_boundary):
...
-    return alpha*x_old + (1-alpha)*x_new
+    return alpha * x_old + (1 - alpha) * x_new
+
 
 # residual
 a_np1 = update_a(du, u_n, v_n, a_n, ufl=True)
 v_np1 = update_v(a_np1, u_n, v_n, a_n, ufl=True)
 
-res = m(avg(a_n,a_np1,alpha_m), v) + k(avg(u_n,du, alpha_f), v)  #TODO: Wext(v) needs to be replaced by coupling
+res = m(avg(a_n, a_np1, alpha_m), v) + k(avg(u_n, du, alpha_f), v)  # TODO: Next(v) needs to be replaced by coupling
```
BenjaminRueth, line 169: I am not really sure what this `TODO` means. What is `Next(v)`? An Issue for this `TODO` would be helpful.

Note: I could not find out where the `v` comes from. Am I overseeing something? Does this code run?

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
```diff
@@ -1,8 +1,8 @@
-#Import required libs
+# Import required libs
 from fenics import *
```
BenjaminRueth, line 2: We should avoid `from ... import *` statements. This is a quick fix that we can just add in this PR.

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
```diff
@@ -63,176 +62,188 @@ def Neumann_Boundary(x, on_boundary):
...
-    t, n, precice_timestep_complete, precice_dt, Forces_x, Forces_y = precice.advance(u_np1, u_np1, u_n, t, float(dt), n)
-    
-    
-    if precice_timestep_complete:
-        
+
+    dt = Constant(np.min([precice_dt, fenics_dt]))
+
+    # Write new displacements to preCICE
+    precice.write_data(u_np1.copy())
```
BenjaminRueth, line 217: Again: Do we need the `copy()`? See above.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -26,21 +26,23 @@
...
+from problem_setup import get_geometry, get_problem_setup
+import dolfin
+from dolfin import FacetNormal, dot
 
 
 def determine_gradient(V_g, u, flux):
     """
     compute flux following http://hplgit.github.io/INF5620/doc/pub/fenics_tutorial1.1/tu2.html#tut-poisson-gradu
-    :param mesh
+    :param V_g:
```
BenjaminRueth, line 43:
```suggestion
    :param V_g: Vector function space
```

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -26,21 +26,23 @@
...
     :param u: solution where gradient is to be determined
-    :return:
+    :param flux:
```
BenjaminRueth, line 45:
```suggestion
    :param flux: returns computed flux into this value
```

Might be worth considering changing the signature of the whole function to make it more pythonic:

```
def determine_gradient(V_g, u):
    ...
    return flux
```

Does this work or am I missing something?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 args = parser.parse_args()
 
-config_file_name = "precice-config.xml"  # TODO should be moved into config, see https://github.com/precice/fenics-adapter/issues/5 , in case file doesnt not exsist open will fail
+config_file_name = "precice-config.xml"  # TODO should be moved into config, see https://github.com/precice/fenics-adapter/issues/5 , in case file doesnt not exist open will fail
 
-subcycle = Subcyling.NONE
+subcycle = Subcycling.NONE
+
+fenics_dt = None
+error_tol = None
```
BenjaminRueth, line 77: Is this required?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 alpha = 3  # parameter alpha
 beta = 1.3  # parameter beta
 gamma = args.gamma  # parameter gamma, dependence of heat flux on time
 
+# Create mesh and separate mesh components for grid, boundary and coupling interface
 domain_part, problem = get_problem_setup(args)
 mesh, coupling_boundary, remaining_boundary = get_geometry(domain_part)
 
-# Create mesh and define function space
+adapter_config_filename = None
```
BenjaminRueth, line 108: Is this required?

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 if problem is ProblemType.DIRICHLET:
     adapter_config_filename = "precice-adapter-config-D.json"
 elif problem is ProblemType.NEUMANN:
     adapter_config_filename = "precice-adapter-config-N.json"
```
BenjaminRueth, line 112:
```suggestion
```
Move this part down.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 
-precice = Adapter(adapter_config_filename, interpolation_strategy=interpolation_strategy)
+# Adapter definition and initialization
+precice = Adapter(adapter_config_filename)
 
+# Initialize the adapter according to the specific participant
 if problem is ProblemType.DIRICHLET:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=u_D_function,
-                                    write_field=f_N_function, u_n=u_n)
+    precice_dt = precice.initialize(coupling_boundary, mesh, u_D_function, V)
```
BenjaminRueth, line 135:
```suggestion
    precice = Adapter(adapter_config_filename="precice-adapter-config-D.json")
    precice_dt = precice.initialize(coupling_boundary, mesh, u_D_function, V)
```

If you remove the part from above, you can create the `Adapter` here.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 elif problem is ProblemType.NEUMANN:
-    precice_dt = precice.initialize(coupling_subdomain=coupling_boundary, mesh=mesh, read_field=f_N_function,
-                                    write_field=u_D_function, u_n=u_n)
+    precice_dt = precice.initialize(coupling_boundary, mesh, f_N_function, V_g)
```
BenjaminRueth, line 137:
```suggestion
    precice = Adapter(adapter_config_filename="precice-adapter-config-N.json")
    precice_dt = precice.initialize(coupling_boundary, mesh, f_N_function, V_g)
```

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
 
 dt = Constant(0)
 dt.assign(np.min([fenics_dt, precice_dt]))
 
 # Define variational problem
 u = TrialFunction(V)
 v = TestFunction(V)
-f = Expression('beta + gamma * x[0] * x[0] - 2 * gamma * t - 2 * (1-gamma) - 2 * alpha', degree=2, alpha=alpha, beta=beta, gamma=gamma, t=0)
+f = Expression('beta + gamma * x[0] * x[0] - 2 * gamma * t - 2 * (1-gamma) - 2 * alpha', degree=2, alpha=alpha,
+               beta=beta, gamma=gamma, t=0)
```
BenjaminRueth, line 150:
```suggestion
f = Expression('beta + gamma*x[0]*x[0] - 2*gamma*t - 2*(1-gamma) - 2*alpha', degree=2, alpha=alpha, beta=beta, gamma=gamma, t=0)
```
Does this improve readability? I am unsure...

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -171,22 +205,45 @@ def determine_gradient(V_g, u, flux):
...
     # Compute solution u^n+1, use bcs u_D^n+1, u^n and coupling bcs
     solve(a == L, u_np1, bcs)
 
+    # Write data to preCICE according to which problem is being solved
     if problem is ProblemType.DIRICHLET:
-        # Dirichlet problem obtains flux from solution and sends flux on boundary to Neumann problem
+        # Dirichlet problem reads temperature and writes flux on boundary to Neumann problem
         determine_gradient(V_g, u_np1, flux)
-        t, n, precice_timestep_complete, precice_dt = precice.advance(flux, u_np1, u_n, t, dt(0), n)
+        precice.write_data(flux.copy())
```
BenjaminRueth, line 226:
```suggestion
        precice.write_data(flux)
```
See above.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -171,22 +205,45 @@ def determine_gradient(V_g, u, flux):
...
     elif problem is ProblemType.NEUMANN:
-        # Neumann problem obtains sends temperature on boundary to Dirichlet problem
-        t, n, precice_timestep_complete, precice_dt = precice.advance(u_np1, u_np1, u_n, t, dt(0), n)
-
-    dt.assign(np.min([fenics_dt, precice_dt]))  # todo we could also consider deciding on time stepping size inside the adapter
-
-    if precice_timestep_complete:
+        # Neumann problem reads flux and writes temperature on boundary to Dirichlet problem
+        precice.write_data(u_np1.copy())
```
BenjaminRueth, line 229:
```suggestion
        precice.write_data(u_np1)
```
See above.

HT/partitioned-heat/fenics-fenics/precice-config.xml
```diff
@@ -45,8 +45,8 @@
          <max-iterations value="100"/>
          <exchange data="Flux"        mesh="NeumannNodes" from="HeatDirichlet" to="HeatNeumann" />
          <exchange data="Temperature" mesh="NeumannNodes" from="HeatNeumann"   to="HeatDirichlet" initialize="true"/>
-         <relative-convergence-measure data="Flux"        mesh="NeumannNodes" limit="1e-5"/>
-         <relative-convergence-measure data="Temperature" mesh="NeumannNodes" limit="1e-5"/>
+         <relative-convergence-measure data="Flux"        mesh="NeumannNodes" limit="1e-4"/>
+         <relative-convergence-measure data="Temperature" mesh="NeumannNodes" limit="1e-4"/>
```
BenjaminRueth, line 49: Why did you change the accuracy limit? I would suggest to not do it, if it is not absolutely necessary.

Even though it is a bit surprising that you require less accuracy than before and still you reach the same error? This either means that we were too strict before or that the accuracy has been improved.

HT/partitioned-heat/fenics-fenics/heat.py
```diff
@@ -54,89 +56,120 @@ def determine_gradient(V_g, u, flux):
...
+parser.add_argument("-i", "--interface", metavar="interface_type string", type=str, choices=['simple', 'complex'],
+                    help="Type of coupling interface case to be solved. Options: simple, complex", default="simple")
+parser.add_argument("-dom", "--domain", metavar='domain_type string', type=str,
+                    choices=['left', 'right', 'circular', 'rectangle'],
+                    help="Specifying part of the domain being solved. For simple interface the options are left, right, for complex interface the options are circular, rest")
 
 args = parser.parse_args()
 
-config_file_name = "precice-config.xml"  # TODO should be moved into config, see https://github.com/precice/fenics-adapter/issues/5 , in case file doesnt not exsist open will fail
+config_file_name = "precice-config.xml"  # TODO should be moved into config, see https://github.com/precice/fenics-adapter/issues/5 , in case file doesnt not exist open will fail
```
BenjaminRueth, line 72: I think this is also not required anymore.

FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py
IshaanDesai: The `TODO` is not longer relevant from the new adapter design and has been removed. Here `v` is the `TestFunction` based on the vector function space V (`V = VectorFunctionSpace(mesh, 'P', 2)`).
This perpendicular flap FSI example runs and the results are reasonable.

HT/partitioned-heat/fenics-fenics/heat.py
IshaanDesai: From one of our earlier discussions we had concluded that we need to directly send a FEniCS Function into the `solve(a == L, flux)` command, and then this function will be updated by reference. Hence we need to stick to a non-pythonic `return` mechanism here.

HT/partitioned-heat/fenics-fenics/heat.py
IshaanDesai: The code still runs if I remove this but according to PEP-8 nomenclature variables defined inside `if` statements need to have an initial definition before the conditional definition within the `if` loop.

PR #67, opened 2020-01-29
Updates config files and fixes API compatibility

KyleDavisSA:
This PR changes all of the tutorial config files. `timestep-length` is changed to `time-window-size` and `timestep` to `time-window` (according to  https://github.com/precice/precice/pull/619#issuecomment-578889853) from PR https://github.com/precice/precice/pull/619


BenjaminRueth:
We will also deal with the `v2.0.0` compatibility problems here.

PR #78, opened 2020-05-22
cylinderFlap OpenFOAM: Remove residualControl, version-specific

### LINKED ISSUE #40: Some FSI tutorials are not working with OpenFOAM6
### What's the problem?

I am using OpenFOAM6 on my machine and the tutorial cases `tutorials/FSI/cylinderFlap/*` do not work. I explicitly tested the case `tutorials/FSI/cylinderFlap/OpenFOAM-FEniCS` while checking #38. For the other cases I tried executing `runFluid` and it always resulted in an error thrown by OpenFOAM (OF), even before the coupling was initialized.

### My proposed workaround

Commenting out some lines in `Fluid/system/fvSolution` does the trick and it works:

```
PIMPLE
{
    nCorrectors 2;
    nNonOrthogonalCorrectors 0;
    tolerance 1.0e-14;
    relTol 5e-3;
    pisoTol 1e-6;
    consistent  true;

    nOuterCorrectors 50;

    // residualControl
    // {
    //        U
    //    {
    //        tolerance 5e-5;
    //        relTol      0;
    //    }
    //    p
    //    {
   //         tolerance 5e-4;
   //         relTol      0;
   //     }
   // }
}
```

Note again: I only explicitly tested the case with OpenFOAM and FEniCS. For OpenFOAM and deal.II, OpenFOAM and CalculiX respectively, I only ran `runFluid` and with the changes mentioned above OpenFOAM does not exit with an error, but executes until coupling is initialized and it is waiting for the second participant.

### What I already did

* The FEniCS test cases are working (until the very end) without changes on my machine with OF6 using the revision provided on master.
* I needed to apply some changed for the other cases (CalculiX, deal.II). Using the revision on branch https://github.com/precice/tutorials/tree/OpenFOAM6 the cases are working without changes, but I only validated until the coupling is initialized and OpenFOAM is waiting for the Solid solver. https://github.com/precice/tutorials/commit/7f3344b40bbbd2f408049ec49d83884bf9917673. 

### What should we do now?

The fix I am proposing above seems to work. However, there are some more things we should do in order to improve the compatibility of the tutorials with different version of OpenFOAM (**bold** for important ones and *italics* for less important):

- [ ] **Check whether the proposed fix worsens the performance** Especially the first few iterations are very expensive. This might be a property of the test case or the performance has been worsened by removing the `residualControl`. I am not an OpenFOAM expert and I did not compare with the performance using a different OpenFOAM version that accepts `residualControl`.
- [ ] **Test tutorials and OF-adapter for different versions of OF.** I think we should open this issue in https://github.com/precice/systemtests. Currently, we only test with OF4 (see [here](https://github.com/precice/systemtests/blob/master/Dockerfile.openfoam-adapter#L18)).
- [ ] **Test cylinderFlap.** Again, this is an issue for https://github.com/precice/systemtests. Currently, we only test `flap_perp`. Here, `residualControl` is not provided in `fvSolution` (see [here](https://github.com/precice/tutorials/blob/master/FSI/flap_perp/OpenFOAM-deal.II/Fluid/system/fvSolution)). Therefore, I also did not observe any problems when running `flap_perp` under OF6 and I did not expect any problems to show up running `cylinderFlap`.
- [ ] *follow up:* there are quite some differences in the tolerances provided in `fvSolution` of the both cases. Why?
- [ ] *automatically choose `pimpleFoam` or `pimpleDyMFoam`* In `Fluid/system/controlDict` one has to manually choose the fitting solver. This straightforward, but still has to be done manually.

MakisH:
Instead, offer commented-out ways to get the residualControl in the various OpenFOAM versions, with the current values.

This may affect convergence, but the purpose of the tutorial is to demonstrate how to configure preCICE and the adapters.

This fixes #40.

### DIFF
--- a/FSI/3D_Tube/OpenFOAM-CalculiX/Fluid/system/fvSolution
+++ b/FSI/3D_Tube/OpenFOAM-CalculiX/Fluid/system/fvSolution
@@ -7,206 +7,214 @@ FoamFile
     object fvSolution;
 }
 
-    PISO
+PISO
+{
+    nCorrectors 3;
+    nNonOrthogonalCorrectors 2;
+    tolerance 1e-8;
+    relTol 0;
+    maxIter 30;
+    minIter 1;
+    pisoTol 1e-1;
+}
+
+PIMPLE
+{
+    nCorrectors 3;
+    nNonOrthogonalCorrectors 2;
+    tolerance 1e-5;
+    relTol 0;
+    maxIter 30;
+    minIter 1;
+    pisoTol 1e-1;
+}
+
+SIMPLE
+{
+    nCorrectors 3;
+    nNonOrthogonalCorrectors 2;
+    tolerance 1e-5;
+    relTol 0;
+    maxIter 30;
+    minIter 1;
+    pisoTol 1e-1;
+}
+
+blockSolver
+{
+    nCorrectors 3;
+    nNonOrthogonalCorrectors 2;
+    tolerance 1e-5;
+    relTol 0;
+    maxIter 50;
+    nOuterCorrectors 50;
+    minIter 1;
+    pisoTol 1e-1;
+    convergenceTolerance 1e-6;
+}
+
+solvers
+{
+    p
     {
-        nCorrectors 3;
-        nNonOrthogonalCorrectors 2;
-    	tolerance 1e-8;
-    	relTol 0;
-    	maxIter 30;
-    	minIter 1;
-    	pisoTol 1e-1;
+        solver GAMG;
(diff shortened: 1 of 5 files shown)


MakisH:
@BenjaminRueth Good point, I found and updated more cases, including the 3D_Tube.

In the diff view, please set to "ignore whitespace", as I fixed a few formatting bits in these files as well.

Note that the `.gitignore` file of the FEniCS cases ignores the `Fluid/` directory and then un-ignores the important things, which confused my editor and did not see it in the beginning. (#87)

PR #79, opened 2020-06-05
Update dealii tutorials for new solver

### LINKED ISSUE #77: Inappropriate setup for FSI OpenFOAM tutorials
During my work on precice/dealii-adapter#24, I struggled heavily with added mass instabilities. From the structural participant side, there is not really that much you can do to tackle this issue, since displacement and all its derivatives are zero in the beginning i.e. no inertia. 
Choosing a different setup (fvSchemes/fvSolutions) in the Fluid participant solves the issue (at least in my case), improves it in any case.

Since these are official tutorials, we should provide a good setup, especially considering the instabilities in FSI cases. Many users probably just copy these configurations and I think there have already been some discussions on discourse with users, facing similar problems.

DavidSCN:
Updates the tutorials according to https://github.com/precice/dealii-adapter/pull/26.

The solver apply now stress data instead of force as provided in https://github.com/precice/openfoam-adapter/pull/125.

#77 Becomes a separate PR, since it might take some time to investigate the appropriate parameters.


DavidSCN:
@uekerman @MakisH I think we should remove the 3D (quasi 2D) examples here. They are a great example of how it shouldn't be done, since the amount of work in the solid solver is much larger. Example: For a simple mesh with 150 cells and a polynomial degree of 4 I obtain 5 082 unknowns. The same mesh has in 3D 38 115 unknowns.

Also, the cases are a duplication. Opinions?

MakisH:
These are the only 2D cases we have and with these tutorials we mainly want to teach users how to configure preCICE for 2D (and showcase that the OpenFOAM adapter can also work in quasi-2D).

You write
> They are a great example of how it shouldn't be done, since the amount of work in the solid solver is much larger.
I would not worry about this, as it is clear that our tutorials are mostly toy/skeleton problems.

Do you have any alternative solution in mind so that we can still offer a (better, if possible) 2D-2D tutorial case?

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Solid/parameters.prm
```diff
@@ -3,64 +3,79 @@
...
+  # Force residual tolerance
+  set Tolerance force               = 1.0e-9
 end
 
 subsection precice configuration
     # Cases: FSI3 or PF for perpendicular flap
     set Scenario            = FSI3
 
-    # Whether preCICE is used for coupling to another solver
-    set Enable precice      = true
```
MakisH, line 44: Is this now default / automatically deduced? If yes, let's document it somewhere (maybe we should start using changelogs in the adapters).

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Solid/parameters.prm
```diff
@@ -3,64 +3,79 @@
...
-	
-    # Name of the node based coupling mesh in the precice-config.xml file
-    set Node mesh name      = Solid_nodes
-	
-    # Name of the face based coupling mesh in the precice-config.xml file
-    set Face mesh name      = Solid_faces
+    set Participant name    = Solid
+
+    # Name of the coupling mesh in the precice-config.xml file
+    set Mesh name           = Solid_mesh
```
MakisH, line 74: Do we now only have one possible mesh name? I may be overseeing some documentation here.

DavidSCN:
Sorry, I was a bit unclear.
I propose to remove the (in FE space 3d) `OpenFoam-deal.II` tutorials located in `flap_perp` and `cylinderFlap` and keep only the `flap_per_2D` and `cylinderFlap_2D` cases.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/precice-config.xml
```diff
@@ -8,65 +8,62 @@
...
-        <use-mesh name="Solid_nodes" from="Solid"/>
-        <use-mesh name="Solid_faces" from="Solid"/>
+        <use-mesh name="Solid_mesh" from="Solid"/>
         <read-data name="Displacement" mesh="Fluid-Mesh-Nodes"/>
-        <write-data name="Force" mesh="Fluid-Mesh-Centers"/>
-        <!--export:vtk directory="coupling_data" /-->
-        <mapping:rbf-thin-plate-splines direction="write" from="Fluid-Mesh-Centers" to="Solid_faces" constraint="conservative" />
-        <mapping:rbf-thin-plate-splines direction="read" from="Solid_nodes" to="Fluid-Mesh-Nodes" constraint="consistent"/>
+        <write-data name="Stress" mesh="Fluid-Mesh-Centers"/>
+        <export:vtk directory="coupling_data" />
```
MakisH, line 32: Do we want this enabled by default for users?

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Fluid/system/controlDict
```diff
@@ -19,13 +19,13 @@ startTime           0;
 
 stopAt              endTime;
 
-endTime             10;
+endTime             15;
```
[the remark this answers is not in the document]
MakisH: Same for `parameters.prm` and `precice-config.xml`.

MakisH:
> Sorry, I was a bit unclear.
> I propose to remove the (in FE space 3d) `OpenFoam-deal.II` tutorials located in `flap_perp` and `cylinderFlap` and keep only the `flap_per_2D` and `cylinderFlap_2D` cases.

Oh, I see, I misunderstood it as "remove the 2D". I don't see any strong reason not to remove the 3D deal.ii cases, but also not a strong reason to remove them. I guess that the 2D is the case adding value for our tutorials, and the 3D is maybe indeed duplicate from the deal.ii side. We should just make sure to document that 3D is also possible and how.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Fluid/system/controlDict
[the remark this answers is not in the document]
DavidSCN: Parameters changed in `controlDict` are not relevant for this. Increasing the total time was just an intention from my side, since the benchmark runs much longer than `t=10`.
The 3D should be removed in my opinion.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Solid/parameters.prm
DavidSCN: No, I removed the option, it is not really useful for the examples. The examples have preconfigured cases. If you disable preCICE, nothing happens (apart from the case, where you specify body forces in the parameter file). But there is not really a benefit from this case. Also, since the preCICE constructor needs the config file, I need to provide it even in the case, where preCICE is disabled.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Solid/parameters.prm
DavidSCN: Yes, there is now only one mesh. Be aware, that this is not the adapter documentation. I still need to add it there though, but I wouldn't document it in the tutorial anywhere.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/precice-config.xml
DavidSCN: Good question. I would enable it maybe for the `perfFlap`, but not for this case, since it produces a lot of data files.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Fluid/system/controlDict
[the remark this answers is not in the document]
MakisH: Then let's keep this open for now: if we decide to drop the 3D variant, let's also keep these changes. If we decide to kep the 3D, let's make both cases consistent.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/Solid/parameters.prm
MakisH: Yes, I mean document it in the wiki and then resolve this.

DavidSCN:
This is now ready to run with the current deal adapter. I also added a `runSolid` script, which can be executed with `-linear` or `-nonlinear`.
Waiting now for https://github.com/precice/systemtests/pull/228 to be merged.

FSI/cylinderFlap_2D/OpenFOAM-deal.II/precice-config.xml
DavidSCN: Disabled it now for cylinder case

uekerman:
I also don't see any strong reason why to remove the 3D case. I would see it as an example how to best approach these cases, but how to work with the dealii solver in 3D. And we have it tested then.

Eventually, we will need to reorganize our tutorials anyway, but that's the story for another day :)

DavidSCN:
Ok, then I need to update the nonlinear solver to cope with 3d and update the 3d tutorial cases here as well!

uekerman:
> Ok, then I need to update the nonlinear solver to cope with 3d and update the 3d tutorial cases here as well!

So, currently only the linear solver supports 3D, not the nonlinear solver, correct?

uekerman:
One more thing: if we use stresses we currently need the develop version of the OpenFOAM adapter, but this PR goes to the master branch. 
@MakisH Can we merge that feature in the OpenFOAM adapter to the master? Or should this PR better target develop?

MakisH:
> @MakisH Can we merge that feature in the OpenFOAM adapter to the master? Or should this PR better target develop?

I just now merged it.

DavidSCN:
> The non-linear solver already for the flap is extremely slow. Any simple ideas how to speed this up? Maybe a coarser mesh?

Do you run the 2D case? Running it in release is considerably faster. I should set it as the default build option.

> So, currently only the linear solver supports 3D, not the nonlinear solver, correct?

That's more a matter of the setup. I have already tested it once for three 3D, but discarded it.

>  Can we merge that feature in the OpenFOAM adapter to the master? Or should this PR better target develop?

Please merge the OpenFOAM version to master, there is also an open PR for a systemtest dev->master update and the deal.II adapter needs it in master as well (which is also the target).

MakisH:
> Please merge the OpenFOAM version to master, there is also an open PR for a systemtest dev->master update and the deal.II adapter needs it in master as well (which is also the target).

For the system tests, I would prefer to leave the decision to @Eder_K

DavidSCN:
> > Please merge the OpenFOAM version to master, there is also an open PR for a systemtest dev->master update and the deal.II adapter needs it in master as well (which is also the target).
> 
> For the system tests, I would prefer to leave the decision to @Eder-K

That's the case, see https://github.com/precice/systemtests/pull/230

> I just now merged it.

Perfect, this will probably fix the failing build there as well.

DavidSCN:
There were some unstaged changes. I added now a .gitignore in these cases in order to avoid it. 
We should maybe add a gitignore for the whole repo.

uekerman:
> > The non-linear solver already for the flap is extremely slow. Any simple ideas how to speed this up? Maybe a coarser mesh?
> 
> Do you run the 2D case? Running it in release is considerably faster. I should set it as the default build option.

If you mean for dealii then yes :+1:

DavidSCN:
You run it in `CMAKE_BUILD_TYPE`  release and it is slowly? This should not be the case.

> ./runFluid -parallel does not work due to the write-consistent mapping. We need to move the mapping to the other participant if we use stresses.

I updated now everything. Only the `config` is erroneous ([flap_perp case](https://github.com/precice/tutorials/blob/dealii_update/FSI/flap_perp_2D/OpenFOAM-deal.II/precice-config.xml)) and leads to a failure after switching the consistent mapping. I get a segfault and I can't figure out the reason right now.

uekerman:
> You run it in `CMAKE_BUILD_TYPE` release and it is slowly? This should not be the case.
> 

Just talked with @DavidSCN 
dealii is by default built in debug and release. If the adapter is built in release it also links against the dealii release version. Then, it's indeed much faster and fast enough for the perpendicular flap. Let's make this the default for the adapter documentation.

DavidSCN:
@uekerman I forgot one thing: Are you able to open up the generated `vtk` files with your paraView version?

uekerman:
> @uekerman I forgot one thing: Are you able to open up the generated `vtk` files with your paraView version?

I can open them, but I only get point data (which I can visualize with the glyph filter), but no cell data.
paraview version 5.4.1

Btw, `Allclean` does not yet remove those files.

DavidSCN:
Ok, no need to hurry, let's discuss it during the next meeting. Your version would be too old. 
ParaView/ vtk has a new feature for higher order methods, which is use in deal.II.

PR #95, opened 2020-07-21
Remove CGX pre-processing step from OpenFOAM-CalculiX FSI tutorials

### LINKED ISSUE #35: FSI with CalculiX requires CGX (can we get rid of it?)
For the FSI tutorials with OpenFOAM and CalculiX (and potentially other), we require that the user already has CGX installed. This is not very difficult (it is, I find, easier than CalculiX, there are even [prebuilt binaries](http://www.dhondt.de/cgx_2.15.bz2)), but it still is one additional piece of software.

Can we already provide the case prepared, so that the user does not have to install CGX as well?

MakisH:
This adds the prepared meshes in the repository. These are only additional files that can be deleted if needed.

Notes:
- The sources to generate a new mesh are still there.
- These files are not overwritten during the simulation.
- The command to prepare the case remains commented-out in the `Allrun` scripts. It is removed from the `runSolid` scripts, as we there aim for as short scripts as possible.
- We don't use CGX in the `FSI/3D_Tube`, `FSI/flap_perp/OpenFOAM-CalculiX`, or in the `SSI/loaded_beam` tutorials.
- It also adds `flap.12d` in the list of files that the `Allclean` scripts should delete (somehow missed before)

Post-merge:
- [x] Update the tutorial pages in the wiki (I will do)
- [ ] Update the system tests for consistency (should not break) @Eder-K
    - [ ] Remove [the now bundled files](https://github.com/precice/systemtests/tree/develop/tests/TestCompose_of-ccx_fsi.Ubuntu1604.home.PETSc) `all.msh`, `fix1_beam.nam`, `interface_beam.nam`
    - [ ]  Remove [the Docker command to copy them](https://github.com/precice/systemtests/blob/6c98e6126c886d9e29a654d6c88a6c9a880a9e55/tests/TestCompose_of-ccx_fsi.Ubuntu1604.home.PETSc/Dockerfile.tutorial_data#L10).

Review requests:
- Please check if the diff makes sense. I have already ran the cases, but you may also want to run one of them.

Closes #35.


MakisH:
I checked the diff again and corrected a small problem in a README file. Everything else looks good and I have ran the cases. I am merging this to `develop` to test everything together. Please comment here if anything looks wrong.

PR #100, opened 2020-08-03
Add run/clean scripts for SU2-CalculiX and restructure files

### LINKED ISSUE #89: FSI SU2/CalculiX: Restructure run scripts to remove tmux
We are currently triggering tmux from inside the tutorial's run scripts, as e.g. here:
https://github.com/precice/tutorials/blob/master/FSI/flap_perp/SU2-CalculiX/runTutorial_parallel.sh

This is inconsistent with the rest of our cases and probably difficult for most new users to follow. We should restructure the tutorial to have a `runFluid` (also in parallel) and a `runSolid` script that can be started in separate terminals and are easier to follow.

We should also add an `Allrun` and an `Allclean` script, [currently missing](https://github.com/precice/tutorials/tree/master/FSI/flap_perp/SU2-CalculiX).

MakisH:
The `FSI/flap_perp/SU2-CalculiX` was one of the first tutorial cases we got, as part of the BGCE project that resulted into http://run.precice.org/. Since then, we have many more cases that follow a few patterns, which this case currently ignores.

This PR:
- Adds a `runFluid` and a `runSolid` script, which start the two participants separately and use both the screen and `Fluid.log`/`Solid.log` for logging (using `tee`, which we could also apply into more tutorials). These files were previously called `Su2.log` and `Calculix.log`.
- Adds an `Allrun` script, similar to the `OpenFOAM-CalculiX` one.
- Adds an `Allclean` script.
- Changes the configuration of SU2 to write result files only into the `Fluid/` directory (and not the root directory).
- Changes the `exchange-directory` from `..` (`flap_perp/`) to the default (`.`, i.e. `SU2-CalculiX`).
- Removes the duplicate configuration file `euler_config_coupled_mergeSolution.cfg` which was used in case of parallel simulations. This only had trivial changes, irrelevant to the parallelization (`UNST_TIME`, `SOLUTION_FLOW_FILENAME`)

## Todo

- [x] Remove the `runTutorial_serial.sh` and `runTutorial_parallel.sh`.
- [x] Update the `README.md`.
- [x] @Eder-K this will require changes in the system tests scripts, but probably not in the reference results.
- [x] Update the wiki.

I also tried [a few options](https://github.com/su2code/SU2/blob/v6.0.0/config_template.cfg) to reduce the verbosity of the screen output in SU2 (e.g. not print the residuals for all the iterations), but I could not succeed.

Closes #89.


MakisH:
This is now ready for review.

I have tested the intended functionality (starting/cleaning) and it works.

I cannot test the complete simulation, as the Solid simulation diverges in my system with `ERROR: Too many cutbacks`, but this is the case also with the state before this PR, as well as with preCICE v2.1 and v2.0.2. However, I don't think we need to run a complete simulation for this PR.

MakisH:
@Eder-K please try to apply these changes to the system tests and approve when ready.

FSI/flap_perp/SU2-CalculiX/Allrun
```diff
@@ -0,0 +1,80 @@
...
+    # echo "Preparing the ${Participant2} participant..."
+    # cd ${Participant2}
+    #     echo "  Executing cgx (provided by CalculiX, make sure this exists)..."
+    #     cgx -bg pre_flap.fbd > prepare_flap.log 2>&1
+    # cd ..
+
+    # Run
+    echo "  Starting the ${Participant2} participant..."
+    ${Solver2} -i ${Participant2}/flap -precice-participant Calculix > ${Participant2}.log  2>&1 &
+    PIDParticipant2=$!
```
uekerman, line 54:
```suggestion
echo "  Starting the ${Participant2} participant..."
${Solver2} -i ${Participant2}/flap -precice-participant Calculix > ${Participant2}.log  2>&1 &
PIDParticipant2=$!
```

FSI/flap_perp/SU2-CalculiX/Allrun
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+cd ${0%/*} || exit 1                        # Run from this directory
+. $WM_PROJECT_DIR/bin/tools/RunFunctions    # Tutorial run functions
+
```
uekerman, line 5: That's probably a copy-paste leftover
```suggestion
```

FSI/flap_perp/SU2-CalculiX/runFluid
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+cd ${0%/*} || exit 1    		    		# Run from this directory
+. $WM_PROJECT_DIR/bin/tools/RunFunctions    # Tutorial run functions
```
uekerman, line 3: That's probably a copy-paste leftover
```suggestion
```

FSI/flap_perp/SU2-CalculiX/Allrun
MakisH: I am confused from the diff. Is this a formatting change?

MakisH: The indentation indeed does not make much sense here, I fixed it.

Eder-K (approved):
[The systests are updated](https://github.com/precice/systemtests/pull/251), the test is now launched by executing the `runFluid`/`runSolid` scripts in the respective adapters. 

However, the test does still fail during finalization due to a [known double free error](https://github.com/precice/su2-adapter/issues/15) in the SU2 adapter: https://travis-ci.org/github/precice/systemtests/jobs/722485626#L30853
This is not exclusive to this new test structure though, as the current cron jobs on systemtest regularly fail with the same problem: https://travis-ci.org/github/precice/systemtests/jobs/722480420#L30824

MakisH:
I also fixed the watchpoint name in the `plotDisplacement.sh` (somehow I had not yet committed this change).

PR #101, opened 2020-08-11
Update FEniCS cases data initialization

BenjaminRueth:
Ensures compatibility with changes introduced in https://github.com/precice/fenics-adapter/pull/81.

# Update

- [x] `heat.py` Note: Other tutorials do not use data initialization
- [x] `cyl-flap.py` uses `PointSources`
- [x] `perp-flap.py` uses `PointSources`

# Check for regressions

- [x] `heat.py`
- [x] `flap_perp`
- [x] `cylinderFlap`
- [x] `flow-over-plate`


BenjaminRueth:
I ran all cases and checked the `precice-*-iterations.log` files for the FEniCS participants. Cases worked nicely and no regression was visible.

PR #102, opened 2020-08-13
Use SegregatedRBFInterpolationExpression for partitioned heat equation tutorial.

BenjaminRueth:
Tutorial now uses new SegregatedRBFInterpolationExpression from https://github.com/precice/fenics-adapter/pull/83


BenjaminRueth:
I also tried running the complex geometry case. However, results were not as good as reported in https://github.com/precice/fenics-adapter/pull/82#issuecomment-672896370. We should inverstigate this furtherly, but for this PR everything is ok.

PR #105, opened 2020-09-08
FSI3 deal.II-OpenFOAM

DavidSCN:
Results and configuration of the deal.II-OpenFOAM FSI3 tutorial. Reference data has been taken from the [original paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.550.1689&rep=rep1&type=pdf) of Hron and Turek. I used groovyBC (swak4foam) in order to apply the correct parabolic velocity profile and ramp it via consine function.

I uploaded in this PR also some result data such as the watchpoint, the forces, and the Solid-iteration.log file. The simulation was performed in serial and until a final time of 10 sec. The overall computational time was around two days. Since the reference comprises only qualitative data above 19 sec. the results of our simulation have been shifted by approximately 15 sec. in order to compare magnitude and frequency. I think the precice configuration could still be improved (cf. precice-Solid-iterations)

We see that x- and y-displacement results are in a good agreement, magnitude of lift and drag are still slightly different.
It would be interesting to use this setup with other structure codes in order to compare the results.


FSI/fsi3/precice-config.xml
```diff
@@ -0,0 +1,93 @@
...
+	 <data name="Displacement" mesh="Solid_mesh" scaling="1.0"/>
+	 <data name="Stress" mesh="Fluid-Mesh-Centers" scaling="100"/>
+  	<initial-relaxation value="0.1"/>
+  	<preconditioner type="residual-sum"/>
+  	<filter type="QR2" limit="1e-2"/>
+  	<max-used-iterations value="60"/>
+  	<time-windows-reused value="0"/>
+	</acceleration:IQN-IMVJ-->
+        <acceleration:IQN-ILS>
+            <data name="Displacement" mesh="Solid_mesh"/>
```
uekerman, line 71: A parallel-implicit coupling scheme needs both data vectors in the acceleration:
```suggestion
            <data name="Displacement" mesh="Solid_mesh"/>
            <data name="Stress" mesh="Fluid-Mesh-Centers"/>
```

FSI/fsi3/precice-config.xml
```diff
@@ -0,0 +1,93 @@
...
+        <max-time value="10"/>
+        <participants first="Fluid" second="Solid"/>
+        <exchange data="Stress" mesh="Fluid-Mesh-Centers" from="Fluid" to="Solid"/>
+        <exchange data="Displacement" mesh="Solid_mesh" from="Solid" to="Fluid" initialize="0"/>
+
+        <max-iterations value="150"/>
+	<relative-convergence-measure limit="5e-5" data="Stress" mesh="Fluid-Mesh-Centers"/>
+        <!--relative-convergence-measure limit="5e-5" data="Displacement" mesh="Solid_mesh"/-->
+        <absolute-convergence-measure limit="1e-7" data="Displacement" mesh="Solid_mesh"/>
+        <extrapolation-order value="2"/>
```
uekerman, line 59: Extrapolation does not always help. Try:
```suggestion
        <extrapolation-order value="0"/>
```

FSI/fsi3/Solid/nonlinear_elasticity.prm
```diff
@@ -0,0 +1,81 @@
...
+
+subsection Nonlinear solver
+  # Number of Newton-Raphson iterations allowed
+  set Max iterations Newton-Raphson = 10
+
+  # Displacement error tolerance
+  set Tolerance displacement        = 1.0e-6
+
+  # Force residual tolerance
+  set Tolerance force               = 1.0e-9
```
uekerman, line 60: The nonlinear tolerances are for absolute residuals, right? And the linear for a relative residual.
The linear residual should be tighter (by one or two orders of magnitude) than the nonlinear one. This is easier to specify if all use the same kind. I recommend to use only relative residuals. And why do we need the force residual?

FSI/fsi3/precice-config.xml
```diff
@@ -0,0 +1,93 @@
...
+        <time-window-size value="0.001"/>
+        <max-time value="10"/>
+        <participants first="Fluid" second="Solid"/>
+        <exchange data="Stress" mesh="Fluid-Mesh-Centers" from="Fluid" to="Solid"/>
+        <exchange data="Displacement" mesh="Solid_mesh" from="Solid" to="Fluid" initialize="0"/>
+
+        <max-iterations value="150"/>
+	<relative-convergence-measure limit="5e-5" data="Stress" mesh="Fluid-Mesh-Centers"/>
+        <!--relative-convergence-measure limit="5e-5" data="Displacement" mesh="Solid_mesh"/-->
+        <absolute-convergence-measure limit="1e-7" data="Displacement" mesh="Solid_mesh"/>
```
uekerman, line 58: Those appear too tight compared to the solvers' internal criteria. The internal ones should be 1 to 2 orders of magnitude tighter. I recommend to use only relative criteria.

FSI/fsi3/precice-config.xml
```diff
@@ -0,0 +1,93 @@
...
+  	<initial-relaxation value="0.1"/>
+  	<preconditioner type="residual-sum"/>
+  	<filter type="QR2" limit="1e-2"/>
+  	<max-used-iterations value="60"/>
+  	<time-windows-reused value="0"/>
+	</acceleration:IQN-IMVJ-->
+        <acceleration:IQN-ILS>
+            <data name="Displacement" mesh="Solid_mesh"/>
+            <preconditioner type="residual-sum"/>
+            <filter type="QR2" limit="1.2e-2"/>
```
uekerman, line 73: The filter appears to be too tight. Too many columns get deleted. But this could be due to the tight convergence criteria and the missing data field in the acceleration.

uekerman (commented):
The results look already pretty good. Really going until 19s is normally not necessary. So, what you do there is pretty standard.

However, the coupling iterations are indeed too high.
Please try first to add the missing data field in the acceleration. This should have a huge effect. Maybe also on the lift and drag. Afterwards please try the other suggestions one by one.

FSI/fsi3/Fluid/system/fvSchemes
```diff
@@ -0,0 +1,42 @@
...
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+    ddtSchemes
+    {
+        default backward;
+    }
+
+    gradSchemes
+    {
+        default cellLimited Gauss linear 1;
```
DavidSCN, line 18: In order to track this:
Removing here the `cellLimited` keyword results in similar amplitudes, but the result is phase shifted. Thus, it would be indeed a good idea to perform the simulation until 20 sec to see, where the oscillations are located in time as compared to the reference.

uekerman: AFAIK, the phase shift is rarely analyzed in publications using the FSI3. Normally, people only look at the frequency and the amplitude. Depends on the ramp and the initial conditions (e.g. precomputed flow field), of course. So, yes we can look at this, but should not have the highest priority.

PR #114, opened 2020-11-03
Use same method for flux computation like in HT example.

BenjaminRueth:
This PR updates CHT/flow-over-plat to use the same method for flux computataion like https://github.com/precice/tutorials/tree/master/HT/partitioned-heat/fenics-fenics.


IshaanDesai (approved):
Running the tutorial with `fenics-adapter` shows normal results. Everything looks good :+1:

PR #118, opened 2020-11-23
Fix wrong visualization in nutils cht.py

uekerman:
If the system tests only use the coupling vtks this should have no effect. It's a pure visualization problem ... results were just lagging by one timestep. 
@Eder-K  Anything we need to adjust in the system tests or can I just merge?


Eder-K:
No, looks good :+1:

PR #119, opened 2020-11-24
Change FEniCS tutorials to handle new initialization signature of FEniCS-Adapter

IshaanDesai:
The initialization signature of the FEniCS-Adapter is changed in https://github.com/precice/fenics-adapter/pull/101. Correspondingly all FEniCS tutorials need to use the new initialization.

Not to be merged before https://github.com/precice/fenics-adapter/pull/101

### DIFF
--- a/FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py
+++ b/FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py
@@ -88,6 +88,7 @@ clamped_boundary_domain = AutoSubDomain(left_boundary)
 force_boundary = AutoSubDomain(remaining_boundary)
 
 # Initialize the coupling interface
+# Function space V is passed twice as both read and write functions are defined using the same space
 precice_dt = precice.initialize(coupling_boundary, read_function_space=V, write_object=V,
                                 fixed_boundary=clamped_boundary_domain)
 
--- a/FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py
+++ b/FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py
@@ -88,7 +88,7 @@ clamped_boundary_domain = AutoSubDomain(left_boundary)
 force_boundary = AutoSubDomain(remaining_boundary)
 
 # Initialize the coupling interface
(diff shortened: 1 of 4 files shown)


FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py
```diff
@@ -88,7 +88,7 @@ def remaining_boundary(x, on_boundary):
 force_boundary = AutoSubDomain(remaining_boundary)
 
 # Initialize the coupling interface
-precice_dt = precice.initialize(coupling_boundary, mesh, V, dim, fixed_boundary=clamped_boundary_domain)
+precice_dt = precice.initialize(coupling_boundary, V, V, clamped_boundary_domain)
```
uekerman, line 91: Why do we need to pass "V" twice?

IshaanDesai: The new proposed initialization of the adapter is: `initialize(self, coupling_subdomain, read_function_space=None, write_object=None, fixed_boundary=None)`. The function space `V` is passed twice because both read and write functions live on the same function space in this case. It is necessary to pass the function space twice, once as `read_function_space` and again as `write_object` so that the adapter knows it is a bi-directional coupling and the user wants to *read* and *write* data.
Initializing like: `precice.initialize(coupling_boundary, read_function_space=V, fixed_boundary=clamped_bounday_domain)` would mean the user wants to perform a read-only uni-directional coupling.

This new initialization basically provides a way to define uni-directional coupling. Does this make sense?

uekerman: OK, thanks for clarification. I think it's not perfect, but good enough for the moment. Let's maybe add a quick comment in the tutorials code.

BenjaminRueth:
```suggestion
precice_dt = precice.initialize(coupling_boundary, read_function_space=V, write_object=V, fixed_boundary=clamped_boundary_domain)
```
I would suggest to use the keyword to make the actual usage of `V` more explicit. This is optional, but from my point of view improves readability of the tutorial.

BenjaminRueth (approved):
Looks good to me :+1: I ran the HT and CHT cases.

BenjaminRueth:
Correction: Will be fixed in #127

PR #123, opened 2020-12-02
Add a quickstart rigid flap tutorial

### LINKED ISSUE #117: Create quickstart tutorial: rotating rigid body in a channel flow
For the [new docs](https://www.precice.org/precice.github.io_future/quickstart.html), we need a simple, yet interesting enough first testcase. Simple means minimal dependencies, easy to run, easy to understand the physics. Interesting means close to real solvers, sth moving etc.

A good candidate could be an FSI coupling between OpenFOAM and a C++ rigid body solver. The following could work:

quickstart

The tutorial could go to `FSI/rigid_flap/OpenFOAM-RigidBody`.

**Setup**:

* The channel flow uses OpenFOAM.
* A rigid beam is fixed at a rotation point and therefore only has one degree of freedom: the deflection angle.
* The flow is instationary and laminar, a similar regime as the Turek and Hron FSI3 for example. We should get an oscillatory movement of the rigid beam (at least no convergence to a steady state). If necessary we could vertically dislocate the beam (similarly to the Turek and Hron benchmarks)
* The beam should be 2D, we want to use the 2D-3D mapping within OpenFOAM (the preCICE dimension is 2). 
* The angle should get big enough to be easily visible, but still small enough to give a robust mesh movement. 
* The added-mass effect should be large such that implicit coupling and IQN is necessary. Let's use a serial coupling.
* Let's call the participants `Fluid` and `Solid`.

**Rigid body solver**:

* I do not recommend to start from the [existing 6dof solver](https://github.com/precice/6dof-rigid-body-solver), but from scratch.
* It should be one C++ file only, directly hosted in the tutorial. Building with cmake.
* The code should be C++-14 using smart pointers and vectors. No external dependencies (beyond the std libs). 
* We don't need a configuration file. All options can be hard-coded variables.
* Options are the width and height of the beam and the number of boundary vertices along the width and height. Moreover, the mass of the beam. This is probably the best way to fine-tune the added-mass effect in the end. 
* The read data are forces, the write data displacements.
* From the forces, the total angular momentum can be computed and from there the angular acceleration, the angular velocity and then the angle itself. The beam should have inertia. First order time integration is enough. We should document these formulas directly in the quickstart.
* The code should be short and readable. The internal physics could be separated from the coupling in a separate function. But otherwise, no extensive software engineering should be used. Basically one main + one or two separate functions. 
* Mesh connectivity should be defined. We could then use two nearest projection mappings.
* For visualization we can directly use the preCICE export. The beam code does not need an own visualization. It's enough to write the angle, the angular velocity, and the angular acceleration to the screen.
* A watchpoint could track the end of the beam. Let's also provide a gnuplot script to visualize the watchpoint.

DavidSCN:
Resolves #117 
Setup as described in #117 . We decided to remove the `nearest-projection` capability and use a RBF mappings instead.


FSI/rigid_flap/OpenFOAM-RigidBody/Fluid/system/fvSolution
```diff
@@ -0,0 +1,110 @@
...
+    nCorrectors 2;
+    nNonOrthogonalCorrectors 1;
+    tolerance 1.0e-8;
+
+    correctPhi          yes;
+    relTol 1e-4;
+    pisoTol 1e-6;
+    consistent  true;
+
+    // OpenFOAM (.com), OpenFOAM 5 or older (.org)
```
DavidSCN, line 74: I will check if we can remove this block completely since it is an unnecessary complication for newcomer.

uekerman (changes requested):
Nice :+1: 

* Let's also add a `export:vtk` to the Solid participant and explain in the docs how to visualize the point cloud.
* Could you please format the C++ file with the same clang-format options as used in preCICE?

The IQN convergence is still a bit bad. I'll try whether better settings are possible.

FSI/rigid_flap/OpenFOAM-RigidBody/precice-config.xml
```diff
@@ -0,0 +1,75 @@
...
+  <log>
+    <sink
+      filter="%Severity% > debug and %Rank% = 0"
+      format="---[precice] %ColorizedSeverity% %Message%"
+      enabled="true" />
+  </log>
+
+  <solver-interface dimensions="2">
+    <data:vector name="Force" />
+    <data:vector name="Displacement" />
```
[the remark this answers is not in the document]
MakisH: I don't think we really have a consistent scheme right now. I took a note for our tutorials restructuring.

I also think this is not a trivial to argue decision: one could see it as "what is the quantity I couple" (thus singular), but when we speak we usually use plural, as it somehow sounds better. But I think that the singular (used here) is easier to keep consistent (works even on one-point interfaces) and more correct: when writing the mapping equations, one does not see a plural there. Moreover, we could use the plural to refer to the two sides of the interface (`Forces: Force_Fluid, Force_Solid`).

But overall, I would not address this right now, but only at the batch-editing phase.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,252 @@
...
+  std::vector<int> vertex_ids(n_nodes);
+  double           theta_dot = 0.0;
+  double           theta     = 0.0;
+
+  {
+    // Define a boundary mesh
+    std::cout << "Rigid body: defining mesh: ";
+    std::cout << n_nodes << " nodes " << std::endl;
+    const double delta_y = height / (n_vertical_nodes - 1);
+    const double delta_x = length / (n_horizontal_nodes - 1);
```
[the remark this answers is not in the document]
DavidSCN: They are here declared in a local scope, but I moved them up.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,252 @@
...
+  {
+    // Compute total moment m = x^{n} x f^{n+1}
+    double moment = 0;
+    for (uint i = 0; i < forces.size() / 2; ++i)
+      moment += vertices[2 * i] * forces[2 * i + 1] - vertices[2 * i + 1] * forces[2 * i];
+
+    // Store rigid body angle theta^{n}
+    const double theta_old = theta;
+    // Update angle to theta^{n+1} according to forward Euler method (simplified moment
+    // computation, which does not depend on the updated configuration)
```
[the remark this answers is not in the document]
DavidSCN: Done, please check again if this is helpful.

DavidSCN:
> Could you please format the C++ file with the same clang-format options as used in preCICE?

I used now the main repo .clang-format file and it should be formatted properly. I'm still a bit confused why the comments are not formatted (linebreaks), but I guess this is not configured in preCICE itself.

uekerman (approved):
Looks good. Let me only still check the IQN performance. I can merge then.

FSI/rigid_flap/OpenFOAM-RigidBody/Fluid/system/decomposeParDict
```diff
@@ -0,0 +1,25 @@
...
+{
+    version 2.0;
+    format ascii;
+    class dictionary;
+    location system;
+    object decomposeParDict;
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+    numberOfSubdomains 25;
```
MakisH, line 11: Maybe use a more modest number here, so that it can easily be used on a laptop?

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
+#include <iostream>
+#include <math.h>
```
MakisH, line 2: Better use the C++ headers:
```suggestion
#include <cmath>
```

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
...
+                      const double &theta_dot)
+  {
+    old_vertices  = vertices;
+    old_theta     = theta;
+    old_theta_dot = theta_dot;
+  }
+
+  void reload_old_state(Vector &vertices,
+                        double &theta,
+                        double &theta_dot) const
```
MakisH, line 19: Why declare this a const method? It should not work with const objects anyway.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
...
+  {
+  }
+
+  void
+  solve(const Vector &forces,
+        const Vector &initial_vertices,
+        Vector &      vertices,
+        double &      theta,
+        double &      theta_dot,
+        const double  delta_t) const
```
MakisH, line 44: same here for `const`

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
...
+  {
+    // Compute total moment M = x^{n} x f^{n+1}
+    double moment = 0;
+    for (uint i = 0; i < forces.size() / 2; ++i)
```
MakisH, line 48: Just use `auto` here (where does actually `uint` come from?)

```suggestion
    for (auto i = 0; i < forces.size() / 2; ++i)
```

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
+#include <iostream>
```
MakisH, line 1: We should also explicitly include `string` here, even if `iostream` includes some parts:

```suggestion
#include <string>
```

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
```diff
@@ -0,0 +1,246 @@
...
+  const Vector initial_vertices = vertices;
+
+  // Pass the vertices to preCICE
+  precice.setMeshVertices(mesh_id,
+                          n_nodes,
+                          vertices.data(),
+                          vertex_ids.data());
+
+  // initialize the Solverinterface
+  double dt = precice.initialize();
```
MakisH, line 177: In such cases where what is returned already has a type, you can use `auto` to avoid implicit conversions (although here this is clear).

MakisH (commented):
I quickly went through the code (did not try to run it) and I have a few minor suggestions. Feel free to integrate as many as you want. I can also try running it, if you want.

FSI/rigid_flap/OpenFOAM-RigidBody/Fluid/system/decomposeParDict
DavidSCN: Oh I haven't noticed the file at all. I changed it to two.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
DavidSCN: `unsigned int` ?

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
DavidSCN: `iostream` is still required.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
DavidSCN: Why not? I pulled the `markActionFulfilled` statement out in the main loop. So, it is actually a `const` method.

DavidSCN:
> Looks good. Let me only still check the IQN performance. I can merge then.

Where should the `README` go? I guess I should add it before merging this?

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
MakisH: You are right, I quickly read it and understood `reload_old_state` as modifying the DataContainer (i.e. `old_vertices = vertices`). This is totally fine.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
MakisH: It looks like `uint` is compiler-specific: https://stackoverflow.com/questions/3552094/c-uint-unsigned-int-int

MakisH:
> > Looks good. Let me only still check the IQN performance. I can merge then.
> 
> Where should the `README` go? I guess I should add it before merging this?

Yes, please. Even if we keep the instructions in a separate place (already now or later), we need a minimal README here as well.

MakisH (commented):
- I tried running it (serial & parallel) and it works nicely. The Fluid results in ParaView look fine, the watchpoint plot was clear. The motion feels a bit too intense, but good that preCICE converges in such conditions.
- The code looks clean & nice :+1:

Two small todos:
- [ ] We still need a README, but you have this already in mind.
- [x] We still need to add the script `removeObsoleteFolders.sh`, similarly to other OpenFOAM tutorials.

MakisH:
> Looks good. Let me only still check the IQN performance. I can merge then.

Here are the convergence and iteration files for the parallel case:

[precice-Solid-convergence.log](https://github.com/precice/tutorials/files/5657953/precice-Solid-convergence.log)
[precice-Solid-iterations.log](https://github.com/precice/tutorials/files/5657954/precice-Solid-iterations.log)

It starts converging in 3-4 iterations, goes up to 7 iterations in the given time frame.

FSI/rigid_flap/OpenFOAM-RigidBody/Solid/rigid_body_solver.cpp
DavidSCN: Oh, I was not aware, good to know

DavidSCN:
> The motion feels a bit too intense

It's indeed near an (even with implicit coupling) unstable configuration. We could reduce the motion by, e.g.,  modifying the Fluid density.

MakisH:
> We could reduce the motion by, e.g., modifying the Fluid density.

Or maybe reducing the (output) time step. We mainly want to change the visual effect here.

FSI/rigid_flap/OpenFOAM-RigidBody/README.md
```diff
@@ -0,0 +1,27 @@
+# FSI tutorial of a rigid body flap in a channel flow
+
+This tutorial deals with a fluid-structure-interaction problem. The fluid part of the simulation is computed using OpenFOAM and the rigid body motion is a rigid body model written in `c++` with only a single degree of freedom, namely the deflection angle of the flap in the channel. The rigid body is fixed in the origin at (0,0) and the force exerted by the fluid on the rigid body structure causes an oscillatory rotation of the body.
+
+placeholder for a geometry overview
```
MakisH, line 5: TODO

DavidSCN:
I added now a README and the tutorial is ready to merge from my side. 

> Or maybe reducing the (output) time step. We mainly want to change the visual effect here.

I like the visual effect :D

MakisH:
> I added now a README and the tutorial is ready to merge from my side.
> 
> > Or maybe reducing the (output) time step. We mainly want to change the visual effect here.
> 
> I like the visual effect :D

For me it feels that the flap moves too sharply with every plotting step. Usually we have more smooth transitions. But this can indeed be a matter of preference.

uekerman:
I also think we need a smaller timestep size:
* The displacement jumps are too big and, thus, accuracy is probably weak. Might be a strange "first impression".
* The visual effect.  
* Quasi-Newton coupling needs more similarity of the coupled system between the current and the last timestep.

Furthermore, the fluid side is rather unstable. Many inner iterations are needed. Small changes in the parameters can let the case crash (e.g. smaller dt + parallel coupling). We need a setup that is rather insensitive to changes, robustness-wise. I guess users would want to play around with such things.

What worked well for me was sth like:
* density fluid 1000
* nu 0.001
* density solid 10000
* dt = 0.005
* T= 2.5
... but then the displacement gets too big. I guess we will need some sort of rotational spring that works against the fluid force.

DavidSCN:
I added now an implicit rotational spring. After (too) many test runs, I conclude the following: The spring doesn't change the general problem. We have either stable setups or unstable setups. I guess there is even an analytic solution for some characteristic numbers which could tell us the stability region. If we want to see a visual effect in the results, the setups are (after a certain amount of time) unstable, for stable setups, the effect is 'only' visible in the watchpoint. Therefore, I added now a 'switch' in the middle of the simulation, which stiffens the spring so that the unstable initial setup is stabilized. The advantage would be, that you can really play around with the parameters and the simulation is rather stable. If it is not stable, the reason should be clear enough for the user. 
As always, feel free to try it yourself. I just realized that I need to update the given formulas.

uekerman:
I think the stiffness switch is a good solution. With all that, however, the simulation became a little bit too slow maybe. I'll try more things.

uekerman:
With a few tweaks, we are now done to 30s compute time for adapter + preCICE in debug mode. :tada:

DavidSCN:
> With a few tweaks, we are now done to 30s compute time for adapter + preCICE in debug mode. :tada:

Sounds good. I still had some points on my list for the README, in particular the formulas and an explanation for the rigid body model as well as the the explanation of the additional spring. Should this go in a separate PR or are you fine with the current state?

uekerman:
Feel free to open another PR or directly push to develop if the changes are pure extensions.

Issue #112, opened 2020-10-27
Restructure pseudo-2D cases to use preCICE in 2D


MakisH:
We are currently defining a few tutorial cases (more prominently the [FSI/flap_perp](https://github.com/precice/tutorials/tree/master/FSI/flap_perp)) in the following way:
- the solver meshes are 3D, providing a fake additional layer in one direction, and
- the preCICE solver interface is defined as 3D (in the `precice-config.xml`).

After a series of cases that the RBF mapping fails on different systems (see https://github.com/precice/calculix-adapter/issues/37, https://github.com/precice/calculix-adapter/issues/43), @uekerman suggested that these cases are ill-defined and it is a coincidence that they succeed in some systems.

The reason that we were using 3D solver interfaces until now was that some adapters could not handle 2D meshes & data. Since https://github.com/precice/openfoam-adapter/pull/96, the OpenFOAM adapter can automatically work in a "2D coupling mode" and a nice example case is [deal.II-OpenFOAM (2D variant)](https://github.com/precice/tutorials/tree/master/FSI/flap_perp_2D/OpenFOAM-deal.II).

If we had the same functionality for CalculiX, FEniCS, and SU2, we could restructure the pseudo-2D cases to actually use 2D solver interfaces, even if the solvers themselves keep working in 3D internally. Maybe @KyleDavisSA and @IshaanDesai know more on how to implement this in these adapters.

uekerman:
Problem is that the RBF mapping is ill-defined. If we have a case where a 2D mesh is extruded by one layer in span-wise direction, there is, first, not a lot of information to build up an interpolant in span-wise direction. Depending on how the layers on both sides look like this can lead to a ill-conditioned problem. If we mark the span-wise direction dead, though, we get duplicated rows in the RBF system matrix ... so a singular matrix. The QR decomposition directly detects this. If we use GMRES it could still converge if the right-hand side also has identical entries. If those differ, however (system-dependent differences of the order of machine precision might be enough), then also GMRES fails. 
The features "marking axis dead" is not meant for such situations. It is only meant for a single layer of cells that is parallel to an axis.


> If we had the same functionality for CalculiX, FEniCS, and SU2, we could restructure the pseudo-2D cases to actually use 2D solver interfaces, even if the solvers themselves keep working in 3D internally. Maybe @KyleDavisSA and @IshaanDesai know more on how to implement this in these adapters.

[...]

PR #120, opened 2020-11-30
Modify FEniCS tutorials to enable parallel runs

IshaanDesai:
The FEniCS-Adapter now supports parallel runs: https://github.com/precice/fenics-adapter/pull/71.
Correspondingly minor modifications to FEniCS scripts and preCICE configurations for all FEniCS tutorials are necessary to enable parallel runs.

PR #110, opened 2020-10-22
two perpendicular flap case

carme-hp:
Motivation for case **tutorials/FSI/two_flap_perp_2D** is to test multi-coupling. Instead of one single flap, now we have two flaps, which means that the case involves three participants: Fluid, Solid1 and Solid2. For the first one we use OpenFOAM solver pimpleFoam and for the second we use the deal.II Adapter. In this link you can check the different possible approaches for multi-coupling: https://github.com/precice/precice/wiki/Multi-Coupling-Configuration
In this case we use a fully-implicit multi-coupling scheme.

The geometry of the flaps is defined directly in the deal.II Adapter solver. Because the flaps are located at different positions we need different executables for the solid participants. To obtain them we need to compile https://github.com/precice/dealii-adapter/blob/c332050b5770bf99459b8be596fd6e41f3401a26/linear_elasticity/linear_elasticity.cc#L208-L220 changing the parameters regarding the position of the flap. 

- For **Solid1**:

        point_bottom =
   	    dim == 3 ? Point<dim>(-1.05, 0, 0) : Point<dim>(-1.05, 0);  
        point_tip = dim == 3 ? Point<dim>(-0.95, 1, 0.3) : Point<dim>(-0.95, 1); //flap1 

- For **Solid2**: 

        point_bottom =
          dim == 3 ? Point<dim>(0.95, 0, 0) : Point<dim>(0.95, 0);  
        point_tip = dim == 3 ? Point<dim>(1.05, 1, 0.3) : Point<dim>(1.05, 1); // flap2

The executables are named linear_elasticity1 and linear_elasticity2 respectively. 

In the precice-config.xml we define the communication between the solid participants and the Fluid participant. Here we can select the mapping (see https://github.com/precice/precice/wiki/Mapping-Configuration ).

**Current status**

- [x] mapping:nearest-neighbor

- [x]  mapping:rbf-thin-plate-splines

- [x] linear case

- [x] non-linear case

- [x] parallel run

The current priority is to find out why the solver crashes, which happens at t=0.55. If we use the rbf-thin-plate-splines mapping instead, it crashes immediately.

PR #125, opened 2020-12-08
Restructure tutorials

fsimonis:
This PR is a common communication channel for restructuring the tutorials, which should simplify the integration with the website.


quickstart/README.md
```diff
@@ -2,7 +2,7 @@
 title: Quickstart Guide
 permalink: quickstart.html
 keywords: tutorial, quickstart
-summary:
+summary: "Install preCICE on Linux (e.g. via a Debian package), and then couple two OpenFOAM solvers with the OpenFOAM-preCICE adapter."
```
DavidSCN, line 5: .. a little bit unsure, since @MakisH the description again and no description is provided anywhere.: the quickstart consists of the `rigid-body` case, right? Not the heat exchanger as described here.

MakisH: @DavidSCN this branch does not yet have the quickstart case you developed, so we have the OpenFOAM-OpenFOAM tutorial as a placeholder for now. I only "beautified" the placeholder for now.

DavidSCN:
@fsimonis I think we have some merge conflicts. May I ask you to resolve them. Otherwise I cannot choose develop as base branch for restructuring.

quickstart/README.md
DavidSCN: Got it

MakisH:
This also closes #112.

fsimonis:
Can someone with all adapters installed please try to merge this?
There are tons of conflicts due to the diverging history and I cannot verify if the merge is successful or not.

DavidSCN:
Not sure about all available merge strategies. I see a lot of conflicts in the FSI cases: we can safely delete all `flap_perp*` and `cylinder_flap` cases. Maybe a forced `develop` update is in the end a better solution (for us, not the user).

MakisH:
I am a bit concerned that we will probably lose a lot of the history when merging this actually. What is a good way to prevent this?

BenjaminRodenberg:
> I am a bit concerned that we will probably lose a lot of the history when merging this actually. What is a good way to prevent this?

One solution that might help us is splitting this PR into two PRs:

1. Make all relevant changes to prepare the restructuring: change parameters, make all cases consistent, simplify solvers, add new cases, if this is necessary. Here, we should mainly see small, but important changes.
2. Do the actual restructuring: only move files (don't change them!) delete duplicates. Here, it will be difficult to understand all changes in detail, since there will be a lot of code removals and additions, but we are only moving around files. So nothing complicated happens inside the files.

Currently, I see that both points are mixed up in this PR and this will make it very hard to understand what this PR changes from a future perspective. The important changes are hidden in a huge diff.

I think we can still continue using this PR for preparing (and discovering) all the relevant changes for the restructuring. However, if we want to keep a sane history, we need to invest some time into de-spaghettiing this PR at the end. Maybe, by going back to develop and reapplying all changes step-by-step in a manual fashion.

uekerman:
> One solution that might help us is splitting this PR into two PRs:

I am afraid that this will not be possible (or not that easy). Many file changes are directly related to the flattening.

BenjaminRodenberg:
I took care of resolving the merge conflicts. It's a quite big merge, so might be a good idea to merge this into your respective PRs. Some information on what has changed:

- [ ] new case: `FSI/rigid_flap/OpenFOAM-RigidBody/runSolid* (from #123, **needs restructuring**)
- [ ] new case: `FSI/two_flap_perp_2D/OpenFOAM-deal.II` (from #110, **needs restructuring**)
- [x] major changes: `partitioned-heat-conduction/fenics-fenics/heat.py` (#120 and #144)
- [x] new case: `perpendicular-flap/OpenFOAM-FEniCS` (quite a lot was missing here, but now it looks ok, I tested this case, adds changes from #141 and #142)
- [ ] new case: `perpendicular-flap/SU2-FEniCS` (**needs to be added to #146**)
- [ ] new case: `turek-hron-fsi3/solid-fenics` (**needs to be tested**, since `turek-hron-fsi3/fluid-openfoam` is currently broken)

I added the tick-boxes to indicate open todos that I see from the merge. The things that have to be done are marked in **bold**.

DavidSCN:
@BenjaminRodenberg the rigid body restructuring is covered in #145, the two flap case is covered in #151 and the Su2-Fenics perpendicular flap is already included in #146. Please don't ask me why, but `solid-fenics` is not part of `turek-hron-fsi3` (have a look at the project description). So, you could check all of your open boxes.

uekerman:
> Please don't ask me why, but `solid-fenics` is not part of `turek-hron-fsi3`

Because FEniCS only has a linear solid solver and therefore gives too bad results for the FSI3.

BenjaminRodenberg:
> > Please don't ask me why, but `solid-fenics` is not part of `turek-hron-fsi3`
> 
> Because FEniCS only has a linear solid solver and therefore gives too bad results for the FSI3.

I agree that we should not have it in `turek-hron-fsi3` anymore, if it does not meet the quality criteria. Completely throwing it away and/or forgetting about it would also be bad, since it took a lot of work to set up this case.

I can take the care of putting it onto a branch and create a (draft) PR. This will avoid that we totally forget about the case and it might be a nice starting point for a thesis or similar. The currently missing steps are 1) to implement a non-linear model and 2) make sure that results agree with the other cases.

Alternatively, we can also create `turek-hron-fsi3-linear`, but this does not really contribute a meaningful case and looks to me like a dead-end with a lot of unnecessary maintainment.

BenjaminRodenberg:
> Adding the namespaces to the tutorials, documentation etc. (over some time) to slowly teach users a new best practice

from https://github.com/precice/precice/issues/928#issuecomment-776590346 could also fit into this PR, since we are touching all the configs anyway. But it also makes this PR even more gigantic.

PR #126, opened 2020-12-09
Provide vector valued function for initialization

BenjaminRueth:
I explain the bug here: https://github.com/precice/tutorials/commit/41b1a46f08163cd371a22603df7677b45d4c08eb#commitcomment-44959091


IshaanDesai (changes requested):
The changes made are correct but there is another conflict elsewhere regarding passing a vector valued function to the adapter.

IshaanDesai:
In the time loop, only the y-component of the flux is provided to the `write_data()` function. 
```python   
# Dirichlet problem obtains flux from solution and sends flux on boundary to Neumann problem
determine_heat_flux(V_g, u_np1, k, fluxes)
fluxes_y = fluxes.sub(1)  # only exchange y component of flux.
precice.write_data(fluxes_y)

```
This entity is detected as a scalar entity by the adapter. This is a problem because the original function related to the flux is defined on a VectorFunctionSpace `V_g`.

BenjaminRueth:
I just realized that OpenFOAM requires a scalar quantity. So the current approach is fine (at least for develop), if we don't want to change the OpenFOAM-side.

PR #127, opened 2020-12-10
Initialize with V_g.sub(1) as write_object

BenjaminRueth:
This is the actually correct fix for the bug mentioned in #126.

With this fix we do not have to change the OpenFOAM side of the simulation, which expects a scalar valued flux. We now provide a subspace of the vector valued function space `V_g` as `write_object`. `V_g.sub(1)` is scalar and we can still write the scalar valued flux `flux_y`.


BenjaminRueth:
Verified correct functionality of this case with:

* `fenicsprecice-1.0.0rc2` (`develop`), runs and results look good (but I cannot plot contour lines anymore)
* `fenicsprecice-1.0.0rc2+109.ge3878dc` (`parallelMidLevel`), does not work due to problem with dimensions. Needs `precice-config.xml` from #120. However, then still says

```
Rank 0: Nothing to Receive
Rank 0: Nothing to Send
```

PR #128, opened 2020-12-11
Rename all tutorial cases according to new scheme

uekerman:
Following the scheme @MakisH and me came up with.
Note that in this PR the second level is not yet flattened.

@MakisH Please have a quick look if everything seems right. I recommend to not overdo the review here. The diff will be hard to read anyway. We can still change things later (before merging to develop) and this is kind of breaking for everything else. So, we should merge it soon.
@fsimonis Could you please merge this one alongside the necessary changes on the website.

```
    - tools/
    - flow-over-heated-plate/
    - flow-over-heated-plate-steady-state/
    - heat-exchanger/
    - elastic-tube-1d (eventually)       
    - elastic-tube-3d   
    - turek-hron-fsi3 (we only keep the 2D case)
    - perpendicular-flap (we only keep the 2D case)      
    - multiple-perpendicular-flaps
    - partitioned-heat-conduction
    - partitioned-elastic-beam
```


MakisH (approved):
I quickly browsed the files and the names look reasonable. I understand that this also already deletes the pseudo-3D variants of `flap_perp` and `cylinderFlap`, making the "real 2D" the main ones.

uekerman:
> I understand that this also already deletes the pseudo-3D variants of flap_perp and cylinderFlap, making the "real 2D" the main ones.

Yes, exactly.

fsimonis:
@uekerman I have the feeling that this will become a nightmare to merge.
Could you repeat the renames directly on the `tutorials` branch?

uekerman:
Let me try the merge. I know what you did and I know what I did. Should be doable.

PR #130, opened 2020-12-14
Apply new conventions on Turek Hron FSI3 case

uekerman:
This PR applies all main changes from @MakisH and my restructuring discussion on the FSI3 case. This case should serve as a role model to copy from when porting other cases. 

**Summary**:

* I deleted the alternatives `OpenFOAM-CalculiX` and `OpenFOAM-FEniCS` and only kept `OpenFOAM-dealii`. The purpose of the FSI3 case is to have one proper verification case, not to show flexibility of preCICE. So, it's enough to have only one version. OpenFOAM-FEniCS never made really sense here as the structure model in FEniCS is only linear. OpenFOAM-CCX is complicated to port as the CCX adapter currently doesn't support stresses as coupling data. We wanted to keep one tutorial with stresses, however.
* The case is flattened. 
* Both solvers are called from the subdirectories, i.e. `exchange-directory=".."`. Run and clean scripts adjusted (I kept the clean script for the moment).
* Basic information is given in the README.
* Consistent names: `Stress` and `Displacement`. 

@MakisH Please have a detailed look at the README.
@DavidSCN Please try to run the case and please add a displacement plot of the actual case (with this exact mesh). You can directly push to my branch.


turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
```diff
@@ -7,29 +7,30 @@ FoamFile
 }
 // * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
 
-convertToMeters 1;
+scale 1;
```
MakisH, line 10: I just now noticed this. @DavidSCN does this run also with openfoam.org? In the rest of the tutorials we use `convertToMeters`, which is older and (hopefully) still supported. Is the `scale` also available in more versions?

turek-hron-fsi3/fluid-openfoam/system/decomposeParDict
```diff
@@ -8,12 +8,12 @@ FoamFile
...
 // * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
 
-    numberOfSubdomains 6;
+    numberOfSubdomains 25;
     
     method hierarchical;
     hierarchicalCoeffs
     {
-        n (3 2 1);
+        n (5 5 1);
```
MakisH, line 16: I know this is about the restructuring, but I now saw this: we probably need a more modest number here. @DavidSCN

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+summary: The Turek-Hron FSI cases are well-established numerical benchmarks and, therefore, well suited for verification of preCICE itself and the used adapters. In this tutorial, we focus on the FSI3 case, which presents the most challenging case in terms of added mass. Please note that the meshes of this case are significantly finer than for other tutorials. Running the simulation might take a few hours. We do not recommend to run this tutorials as your first preCICE tutorial.  
 ---
 
-"Lorem Ipsum" is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.
+## Setup
+
+The setup is shown schematically here:
+
+![FSI3 setup](images/tutorials-turek-hron-fsi3-setup.png)
```
MakisH, line 12: These are the results from the previous cylinderFlap case. I guess we need a scheme with just the setup, or an image of the actual results.

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+For more information please refer to the original publication of the benchmark [1].
+
+## Available solvers
+
+hier auch wo mann sie herbekommt
```
MakisH, line 18: TODO (but do we want this here?)

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+Fluid participant:
+
+* OpenFOAM. For more information, have a look at the [OpenFOAM adapter documentation](adapter-openfoam-overview.html). 
+
+{% include important.html content="For the parabolic inflow profile, this tutorial requires groovyBC. groovyBC is part of swak4Foam, which can be cloned from [this repository](https://github.com/Unofficial-Extend-Project-Mirror/openfoam-extend-swak4Foam-dev.git). Please follow the building instructions there." %}
```
MakisH, line 24: Bernhard Gscheider (the maintainer of swak4Foam) recommends to not use "the Git repository that is going around". I am not sure why and I cannot find the official repository at the moment (which should be a Mercurial repository on BitBucket). I guess the most official resource for now is https://openfoamwiki.net/index.php/Contrib/swak4Foam (he also maintains the OpenFOAMWiki, but apparently has not updated this page for a while, I pinged him).

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+Solid participant:
+
+* deal.II. For more information, have a look at the [deal.II adapter documentation](adapter-dealii-overview.html). This tutorial requires the nonlinear solid solver. Please copy the nonlinear solver executable to this folder or make it discoverable at runtime.
```
MakisH, line 28:
```suggestion
* deal.II. For more information, have a look at the [deal.II adapter documentation](adapter-dealii-overview.html). This tutorial requires the nonlinear solid solver. Please copy the nonlinear solver executable to this folder or make it discoverable at runtime and update the `solid-dealii/run.sh` script.
```

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+./run.sh
+```
+and
+```
+cd solid-dealii
+./run.sh
+```
+
+You can also run OpenFOAM in parallel by `./run.sh -parallel`.
+Moreover, the name of your solver might differ, depending on your OpenFOAM version. Have a look in the `fluid-openfoam/system/controlDict` file and set the appropriate solver name.
```
MakisH, line 46:
```suggestion
For older OpenFOAM versions, the solver name will differ. If you are using OpenFOAM v1712 / 5.x or older have a look in the `fluid-openfoam/system/controlDict` file and set the appropriate solver name.
```

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+There is an [open issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. Please run the script `removeObsoleteSolvers.sh` to delete the additional files.
```
MakisH, line 48:
```suggestion
There is an [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. Please run the script `removeObsoleteSolvers.sh` to delete the additional files before importing your results in ParaView.
```

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+You may adjust the end time in the precice-config_*.xml, or interupt the execution earlier if you want.
```
MakisH, line 50:
```suggestion
You may adjust the end time in the `precice-config_*.xml`, or interupt the execution earlier if you want.
```

turek-hron-fsi3/README.md
MakisH: Let's also move this to the post-processing part.

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+In the first few timesteps, many coupling iterations are required for convergence. Don't loose hope, things get better quickly.
```
MakisH, line 52:
```suggestion
In the first few timesteps, many coupling iterations are required for convergence. Don't lose hope, things get better quickly. 
```

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+You may adjust the end time in the precice-config_*.xml, or interupt the execution earlier if you want.
+
+In the first few timesteps, many coupling iterations are required for convergence. Don't loose hope, things get better quickly. 
+
+
+## Postprocessing
+   
+You can visualize the results of the coupled simulation using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `Fluid.foam` file. Solid results are in VTK format. If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`). 
+Moreover, we defined a watchpoint at the flap tip and can observe the perpendicular displacement with the script `plotDisplacement.sh`.
```
MakisH, line 58: Let's split this in three paragraphs, as things get complicated:

```suggestion
You can visualize the results of the coupled simulation using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `Fluid.foam` file. Solid results are in VTK format.

If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`). 

Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plotDisplacement.sh`.  The resulting graph shows the vertical (y) displacement of the tip of the flap.
```

(I hope that `y` is correct here)

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,70 @@
...
+
+In the first few timesteps, many coupling iterations are required for convergence. Don't loose hope, things get better quickly. 
+
+
+## Postprocessing
+   
+You can visualize the results of the coupled simulation using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `Fluid.foam` file. Solid results are in VTK format. If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`). 
+Moreover, we defined a watchpoint at the flap tip and can observe the perpendicular displacement with the script `plotDisplacement.sh`. 
+
+TODO: update picture according to actual used mesh.
```
MakisH, line 60: TODO

MakisH (changes requested):
This adds the binary file `nonlinear_elasticity`. We need it in the `.gitingore`.

I also added a few more comments.

Overall, I like it, it only needs a bit of polishing.

turek-hron-fsi3/README.md
DavidSCN: I recently noticed that the foamToVTK only helps for `.org` versions. For `.com` the problem remains.

turek-hron-fsi3/README.md
DavidSCN: 2010 :sweat_smile: 
Anyway this might be interesting. Our whole FSI validation is based on the unofficial git mirror. I hope there are not that big differences for GroovyBC

turek-hron-fsi3/README.md
DavidSCN: I can probably provide some more recent.

DavidSCN:
> Please try to run the case and please add a displacement plot of the actual case (with this exact mesh). You can directly push to my branch.

I guess you mean the most refined mesh on the Fluid side? Or rather the exact given mesh?

turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
DavidSCN: The keyword `scale` is indeed not included in the most recent .org versions. I removed `convertToMeters` because .com raises an information in the beginning 'this keyword is deemed to be .. month old", not sure about the exact number but it was very long and it seems outdated for me.

turek-hron-fsi3/fluid-openfoam/system/decomposeParDict
DavidSCN: I would not choose a more moderate number here since the most refined mesh requires quite some resources. This way, we make the computational cost more clear to the user.

turek-hron-fsi3/README.md
MakisH: More information: https://twitter.com/swakPyFoam/status/1338566289913163777
Let's use both links and clarify that we used the Git repository, but it is unofficial.

turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
MakisH: Can we actually remove it completely? The default should be 1 anyway.

turek-hron-fsi3/precice-config.xml
```diff
@@ -19,52 +19,51 @@
...
       <time-window-size value="1e-3" />
       <max-time value="15" />
       <participants first="Fluid" second="Solid" />
       <exchange data="Stress" mesh="Fluid-Mesh-Centers" from="Fluid" to="Solid" />
-      <exchange data="Displacement" mesh="Solid_mesh" from="Solid" to="Fluid" initialize="0" />
+      <exchange data="Displacement" mesh="Solid-Mesh" from="Solid" to="Fluid" />
       <max-iterations value="100" />
       <relative-convergence-measure limit="1e-4" data="Stress" mesh="Fluid-Mesh-Centers" />
-      <relative-convergence-measure limit="1e-4" data="Displacement" mesh="Solid_mesh" />
-      <extrapolation-order value="2" />
```
DavidSCN, line 65: I just noticed that the number of coupling iterations is a little bit too high according to my expectation. I can of xourse provide the `iterations` file after the run has finished. Are you sure you would like to remove the extrapolation. I once had a comparison and the extrapolation helped to lower the iterations.

turek-hron-fsi3/fluid-openfoam/system/decomposeParDict
MakisH: ok, then @uekerman can you maybe add a note in the README.md that the parallel case runs on 25 processes by default, due to the high mesh size?

turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
DavidSCN: For .com it works without any keyword, just tested it.

DavidSCN:
Another deal.II specific point: since we start the solver now from sub-directories, the log, result and parameter files are all in one directory. We should probably use a specific directory for the solution files.

uekerman:
> > Please try to run the case and please add a displacement plot of the actual case (with this exact mesh). You can directly push to my branch.
> 
> I guess you mean the most refined mesh on the Fluid side? Or rather the exact given mesh?

The results we show here should coincide with the mesh that we provide. 
Which mesh do we currently provide? Compared to the results in the OpenFOAM adapter paper. The mesh should be fine enough to give good results (compute time is not an issue for this case).

DavidSCN:
> > > Please try to run the case and please add a displacement plot of the actual case (with this exact mesh). You can directly push to my branch.
> > 
> > 
> > I guess you mean the most refined mesh on the Fluid side? Or rather the exact given mesh?
> 
> The results we show here should coincide with the mesh that we provide.
> Which mesh do we currently provide? Compared to the results in the OpenFOAM adapter paper. The mesh should be fine enough to give good results (compute time is not an issue for this case).

We provide all meshes (refinement level one to three). The default mesh (20k cells) is the most coarse one, the most refined (45k cells) is called blockMesh_double_refined. If we just show a plot of the y displacement from 0 up to 15 sec, differences will probably only be noticeable in terms of the absolute time.

turek-hron-fsi3/README.md
uekerman: That would be good. If we show physical results they should coincide 100% with the case. Alternative would be to only show a setup image. Please directly push to the source branch.

turek-hron-fsi3/README.md
uekerman: That was just an Easter egg :smile:

uekerman:
> We provide all meshes (refinement level one to three). 

Ah ok. I'll add a note in the README.

> The default mesh (20k cells) is the most coarse one, the most refined (45k cells) is called blockMesh_double_refined. If we just show a plot of the y displacement from 0 up to 15 sec, differences will probably only be noticeable in terms of the absolute time.

Yes, let's keep the 20k as default and let's then add all picture generated from that one.

DavidSCN:
Perfect, 
maybe also add a note for the mapping for the most refined case, where it is probably better to replace the global rbf by sth like:
```xml
        <mapping:rbf-compact-tps-c2 direction="read" from="Fluid-Mesh-Centers" to="Solid-Mesh" support-radius="0.011" constraint="consistent" />
```

turek-hron-fsi3/README.md
uekerman: I added both links. We should add that we used the Git repo in the paper, not here. Anyway, I guess not too much can go wrong with a parabolic inflow profile.

turek-hron-fsi3/README.md
uekerman: done

turek-hron-fsi3/README.md
uekerman: I also removed the `_*`.

turek-hron-fsi3/README.md
uekerman: Which problem are you referring to exactly? The non-synchronized files? Could you make a text suggestion? (or directly push to the source branch).

turek-hron-fsi3/precice-config.xml
uekerman: OK, let's keep it for the FSI3 (but let's not use it for the perpendicular flap). I am just always worried that users copy the file. Extrapolation definitely has robustness issues.

turek-hron-fsi3/README.md
DavidSCN: Yes, the time sync. Since OpenFOAM shows the exact motion of the structure, all in all it is not such a big problem. I have not yet a proper solution for this problem, it might be helpful if you could upload a synchronized nutils result file in https://github.com/precice/dealii-adapter/issues/33

turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
MakisH: Do you have any .org version installed at the moment? If it also works there without a keyword, let's remove it.

turek-hron-fsi3/README.md
uekerman: OK, but I would exclude the time synch problem from this PR. I will add files in the dealii issue. TODO here is done.

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,85 @@
...
+
+Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plotDisplacement.sh`.  The resulting graph shows the vertical (y) displacement of the tip of the flap.
+
+
+![FSI3 watchpoint](images/tutorials-turek-hron-tip-plot.png)
+
+
+## Mesh refinement
+
+In `fluid-openfoam/system/`, we provide three different fluid meshes:
```
MakisH, line 67: How is the user supposed to select a different mesh?

turek-hron-fsi3/README.md
```diff
@@ -1,10 +1,85 @@
...
+
+If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`).
+
+There is an [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. Please run the script `removeObsoleteSolvers.sh` to delete the additional files before importing your results in ParaView.
+
+Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plotDisplacement.sh`.  The resulting graph shows the vertical (y) displacement of the tip of the flap.
+
+
+![FSI3 watchpoint](images/tutorials-turek-hron-tip-plot.png)
+
```
MakisH, line 63:
```suggestion

Before running the simulation again, you may want to cleanup any result files using the script `Allclean`.
```

MakisH (approved):
Some minor things to improve. Other than that, looks very good!

turek-hron-fsi3/fluid-openfoam/system/blockMeshDict_refined
DavidSCN: Checked it now with OF7 and OF8: both versions work with `scale` and without any keyword.

uekerman: I removed it

PR #143, opened 2021-01-04
Reformat FSI/flap_perp/SU2-CalculiX tutorial to new 2D-3D coupling strategy

IshaanDesai:
As discussed in: https://github.com/precice/tutorials/issues/112 the pseudo-2D cases need to use preCICE in 2D. This PR restructures the tutorial `FSI/flap_perp/SU2-CalculiX` in such a way. 
The CalculiX-Adapter needs to be modified in a way to handle the 2D-3D coupling. This modification is being done here: https://github.com/precice/calculix-adapter/pull/49


FSI/flap_perp/SU2-CalculiX/precice-config.xml
```diff
@@ -3,10 +3,10 @@
 <precice-configuration>
 
     <log>
-        <sink filter="%Severity% > debug and %Rank% = 0" format="---[precice] %ColorizedSeverity% %Message%" enabled="true"/>
+        <sink filter="%Severity% > trace and %Rank% = 0" format="---[precice] %ColorizedSeverity% %Message%" enabled="true"/>
```
uekerman, line 6: Any particular reason for this? We would no longer get info output from preCICE.

IshaanDesai: This change is temporary as the this tutorial case is being used to debug the CCX Adapter design for handling 2D-3D coupling. This will be reverted once the debugging is done :+1:

IshaanDesai:
Initial results of SU2-CalculiX perpendicular flap tutorial show stable tip displacement:
tip-displ-SU2-CCX

This plot however widely differs from the SU2-FEniCS perpendicular flap case result:
tip-displ-SU2-FE

Following points are already checked:
* preCICE configuration for both setups is identical
* SU2 configuration for both setups is identical
* Geometric dimensions for both setups are identical

DavidSCN:
I just computed the same setup using the linear deal.II model and the result was exactly the same as in case of FEniCs. The results of Calculix look like the structure is (relative) too stiff or the forces are too small. I don't know how exactly the coupling data is processed in case of Calculix, but a common issue for the quasi 2D cases, which still use 3D meshes, is a unit length geometry for your out-of-plane dimension. If I can identify it correctly, the Calculix mesh has a thickness of 0.3 and not 1. This might lead to differences when processing the data, especially in case the coupling data is integrated over your domain. If it doesn't resolve the differences, let's have look during the next week.

IshaanDesai:
> I just computed the same setup using the linear deal.II model and the result was exactly the same as in case of FEniCs.

This is really great, thanks for checking with deal.II! :smile: deal.II giving the same result is also a good check that the FEniCS side has no problems.

> If I can identify it correctly, the Calculix mesh has a thickness of 0.3 and not 1. This might lead to differences when processing the data, especially in case the coupling data is integrated over your domain. If it doesn't resolve the differences, let's have look during the next week.

This was the case earlier yes, but here: https://github.com/precice/tutorials/pull/143/commits/ea9665a7577151c6775f92dcf13001adfa5be39f I changed the thickness from 0.3 to 1.o to ensure that there is no need to scale forces while mapping between the 2D and 3D meshes. 
I will keep looking for other issues in the CalculiX implementation in the meantime :+1:

IshaanDesai:
As the mesh is rotated from the older YZ-plane orientation to XY-plane orientation, the DoF to be constrained in the CalculiX configuration is Z. The results are now very similar to the SU2-FEniCS case:
tip-displ-SU2-CCX-new

IshaanDesai:
No longer relevant because this reformatting has been handled as part of broad restructuring of the perpendicular flap tutorial cases here: https://github.com/precice/tutorials/pull/146

Issue #137, opened 2020-12-21
Flatten solver hierarchy for partitioned-heat-conduction


uekerman:
As part of the [tutorials restructuring](https://github.com/orgs/precice/projects/5) we want to flatten the second (solver) hierarchy. For the `partitioned-heat-conduction` case this is non-trivial.

We currently have: https://github.com/precice/tutorials/tree/restructure/partitioned-heat-conduction
```
- fenics-fenics 
```
We want instead:
```
- images/
- dirichlet-fenics
- neumann-fenics
- dirichlet-nutils
- neumann-nutils
- precice-config.xml
- clean.sh
- README.md
```

The Nutils case: [coupled.zip](https://github.com/precice/tutorials/files/5724131/coupled.zip) (a deprecated version of the Python bindings is still used)

FEniCS and Nutils need to use the same parameters. Dirichlet/Neumann needs to be fixed to be the left or right part etc.

For further conventions please copy from the `turek-hron-fsi3` structure: https://github.com/precice/tutorials/tree/restructure/turek-hron-fsi3

BenjaminRueth:
One important thing to keep in mind: Currently `partitioned-heat-conduction` uses a [vector-valued flux](https://github.com/precice/tutorials/blob/restructure/partitioned-heat-conduction/fenics-fenics/precice-config.xml#L12) and `flow-over-heated-plate` has a [scalar-valued flux](https://github.com/precice/tutorials/blob/1f814630d06a3ada5a0a11689e6a81fd360cd20c/flow-over-heated-plate/buoyantPimpleFoam-fenics/precice-config.xml#L12) (only x component). In both cases the same method is used by the solver to compute the flux (see [here](https://github.com/precice/tutorials/blob/1f814630d06a3ada5a0a11689e6a81fd360cd20c/flow-over-heated-plate/buoyantPimpleFoam-fenics/Solid/heat.py#L63-L77) and [here](https://github.com/precice/tutorials/blob/1f814630d06a3ada5a0a11689e6a81fd360cd20c/partitioned-heat-conduction/fenics-fenics/heat.py#L40-L53)).

Why do we use a vector valued flux? Because this allows us to also support arbitrary geometries (and get the exact solution) for `partitioned-heat-conduction`. See https://github.com/precice/tutorials/pull/46/commits/23e450bb49b35ad28a4d53a19493c9090e4d709d and related PR.

[...]

PR #153, opened 2021-02-05
Add nutils solver for heat conduction problem

### LINKED ISSUE #152: Fix nutils-nutils heat transport case
#148 applies the new structure from #125 to the heat transport case. We also want to be able to run the heat transport case using nutils-nutils or any combination of fenics and nutils.

I will create a PR with a prototype implementation in a moment.

BenjaminRodenberg:
Closes #152 and #137

# Todo

- [x] make sure that nutils-nutils achieves analytical solution
- [x] make sure that nutils-fenics achieves analytical solution


BenjaminRodenberg:
I and @DavidSCN had a discussion about this PR and whether it still goes into #125. Any opinions on this? From my current point of view wrapping up this PR will take some more time, but it is right now pretty low on my personal priority list. So hard to estimate when it will be done. I think the most reasonable thing would be to postpone this PR, merge #125 as soon as possible and merge this PR into `develop` as soon as it is ready.

DavidSCN:
Let's stick to the original layout of the tutorials, where this case is not yet included. In my opinion there is at the moment also not any urgent need for this case. So, let's add it later on after #125 has been merged.

BenjaminRodenberg:
With d7b5a55 
* `nutils-nutils` works 
* `FEniCS-nutils`(Dirichlet/Neumann) does not work
* `nutils-FEniCS` (Dirichlet/Neumann) does not work

uekerman:
It could very well be that we get a problem with the data mapping as Nutils uses GPs and thus there is no interpolation point in the corner at the coupling interface. For a plot over line at the top boundary, a kink is visible:
Screenshot from 2021-03-26 08-42-54
Here, FEniCS = right participant = Neumann, Nutils = left participant = Dirichlet

The kink is not visible (or less prominent) if you plot the center line.

Another reason could be that the fluxes are computed and set differently?

Anyhow, I am not sure whether we really need to recover the second order solution here. We could also use an infinite-order manufactured solution here (like the one previously used with Nutils). The "does not work" is pretty strict.

BenjaminRodenberg:
@uekerman thanks for checking! At the top boundary the values are fine. As far as I understand that's where you did the measurement.

The main problem that currently worries me is that the flux does not seem to be continuous across the interface. You can see this by looking at the contour lines:

1) Group dataset
2) Plot contour lines

Screenshot from 2021-03-26 14-16-28

I'm using the setup from above (FEniCS = right participant = Neumann, Nutils = left participant = Dirichlet) and show the result at the last timestep. Reference solution is white, computed solution black.

Currently, I'm not really sure where the kink comes from. Especially, since the Nutils-Nutils case does not show this kink.

BenjaminRodenberg:
If I introduce a scaling of the fluxes on the nutils-side by the meshwidth, I get reasonable results (8cf3c7a). I'm not sure why this scaling is required, but it works.

Since I don't want to give up the high accuracy, which is checked automatically in the FEniCS-FEniCS case, I introduced an additional option `--error-tol` on the FEniCS side. With this the following runs are working successfully (Dirichlet-Neumann):

* FEniCS-FEniCS with default `--error-tol`:
    ```
    ~/tutorials/partitioned-heat-conduction/fenics$ python3 heat.py -n
    ~/tutorials/partitioned-heat-conduction/fenics$ python3 heat.py -d
    ```
* FEniCS-Nutils
    ```
    ~/tutorials/partitioned-heat-conduction/fenics$ python3 heat.py -d -e 10e-4
    ~/tutorials/partitioned-heat-conduction/nutils$ ./run_Neumann.sh 
    ```
* Nutils-FEniCS
    ```
    ~/tutorials/partitioned-heat-conduction/nutils$ ./run_Dirichlet.sh
    ~/tutorials/partitioned-heat-conduction/fenics$ python3 heat.py -n -e 10e-3
    ```
* Nutils-Nutils (no error checks, but looks reasonable)
    ```
    ~/tutorials/partitioned-heat-conduction/nutils$ ./run_Dirichlet.sh
    ~/tutorials/partitioned-heat-conduction/nutils$ ./run_Neumann.sh
    ```

@uekerman I think this PR is now ready for review. Can you especially take a look at the rather hacky solution I found for the output of Nutils? We can also use a less strict `--error-tol` (e.g. `10**-3`) as default, but in the tests, we should stick to the strict `--error-tol` (`10**-12`) and therefore we would have to update a few things.

partitioned-heat-conduction/nutils/Dirichlet.vtk
```diff
@@ -0,0 +1 @@
+Don't delete this file. Just needed to make output nicer.
```
uekerman, line 1: Why do we need these files? A bit confusing in my opinion.

uekerman (commented):
Nice catch with the flux :+1:
I still need to have a closer look. Don't we have the same inconsistency then also for the `flow-over-heated-plate` case?

Why is the error tol different for FEniCS-Nutils and FEniCS-FEniCS? Is this expected? 

* Please adjust cleaning scripts to https://github.com/precice/tutorials/pull/169
* Run scripts are a bit inconsistent. There are some for Nutils, but not for FEniCS. Could we do one `run.sh` per solver with an argument `-d/-n` (or similar)?

BenjaminRodenberg:
> Why is the error tol different for FEniCS-Nutils and FEniCS-FEniCS? Is this expected?

I assume that this is due to the GaussPoint / no GaussPoint situation. I guess the mapping might introduce an error here (even though I'm not 100% sure).

uekerman:
> I assume that this is due to the GaussPoint / no GaussPoint situation. I guess the mapping might introduce an error here (even though I'm not 100% sure).

Let's add this assumption to the README as well.

partitioned-heat-conduction/nutils/Dirichlet.vtk
BenjaminRodenberg: The problem is that nutils creates a file `Dirichlet.vtk` for the first/zeroth time step, `Dirichlet-1.vtk` for the next, and so on. Paraview does not understand this pattern and therefore I introduced a `Dirichlet.vtk` to force Nutils to start with `Dirichlet-1.vtk`. I don't know whether I can tell Nutils how to format its output. Right now this "solution" is the only one that gives me filenames that are consistent.

The way how Nutils creates output is also the reason why I introduced the `run_Dirichlet.sh` and `run_Neumann.sh` scripts: If you do not clean up before the run, Nutils just continues with the output where it stopped in the previous run. Meaning, if `Dirichlet-10.vtk` already exists, Nutils will next create a `Dirichlet-11.vtk` for the first timestep. FEniCS just overwrites output from past simulation here.

BenjaminRodenberg:
> > I assume that this is due to the GaussPoint / no GaussPoint situation. I guess the mapping might introduce an error here (even though I'm not 100% sure).
> 
> Let's add this assumption to the README as well.

Done. 
1af25f9

partitioned-heat-conduction/nutils/Dirichlet.vtk
uekerman: We could use the same Nutils solution as in `flow-over-heated-plate` to prevent this. No need for the workaround then.

uekerman: done in https://github.com/precice/tutorials/pull/153/commits/ed40785ddaf13df67d997436c67e6f7c3a19b324

uekerman:
@BenjaminRodenberg https://github.com/precice/tutorials/pull/153/commits/c66626cc4f4ff8b9937c10752a83fdc0452785bf OK?
The pure fact that we have non-matching meshes and thus a mapping error should not be problematic as the mapping is still 2nd order accurate. I assume it's a similar extrapolation "feature" we already observed in the past.

uekerman:
> If I introduce a scaling of the fluxes on the nutils-side by the meshwidth, I get reasonable results (8cf3c7a). I'm not sure why this scaling is required, but it works.

I have the feeling that we need the scaling somewhere else (on the FEniCS side?).

If I print the very first flux computation **with** the scaling, so

```python
writedata = couplingsample.eval( 'flux' @ ns, fluxdofs=fluxdofs(fluxvalues)) * (nelems-1)
```
I get:

nelems = 10

```
[ 31.34941834   2.38542747  -2.18264018  14.30115775  18.43439137
  13.24279768  11.97285338  13.6948947   14.45439284  14.80733904
  13.78946635  10.65564022  13.21745729  23.35029781  14.30794769
 -20.52905276   2.38360811  99.81916223]
```

nelems = 20

```
[ 94.74738819 -22.50456939 -40.70734131  26.81371853  43.44651194
  21.36708876  16.98722325  27.10063131  30.10497682  28.19959332
  27.7646735   28.48183395  28.82687565  29.05238666  29.13855253
  29.14845106  29.20402719  29.34596548  29.39754181  29.39651266
  29.34568972  29.20786803  29.14852495  29.12421832  29.05236686
  28.88037167  28.48183926  27.56502366  28.1995919   30.85008016
  27.10063169  14.20645971  21.36708866  53.82446274  26.81371856
 -79.438381   -22.5045694  239.29359613]
```

nelems = 40

```
[ 322.13954335 -188.29571396 -260.8850861    51.23231845  128.62206706
   27.93738771    4.62766913   41.62901303   51.59620955   41.82575295
   41.03953953   48.66202113   51.65983756   52.22754276   53.0278544
   54.64664127   55.42319024   55.92597465   56.55539976   57.77223673
   58.37564912   58.80736542   59.00820854   59.12520586   58.88474711
   58.11080428   58.20853735   59.24949191   59.76237405   60.12263955
   60.31680291   60.48700155   60.2498941    59.43190587   59.2828346
   59.69355257   60.05639304   60.63697369   60.84948096   60.84948096
   60.63697369   60.05639306   59.69355257   59.28283455   59.43190587
   60.24989431   60.48700155   60.31680213   60.12263955   59.76237696
   59.24949191   58.20852646   58.11080428   58.88478775   59.12520586
   59.00805688   58.80736542   58.37621515   57.77223673   56.55328729
   55.92597464   55.43107408   54.64664126   52.99843154   52.22754276
   51.76964516   48.66202113   40.62973196   41.82575295   53.1256322
   41.62901303   -1.08021392   27.9373877   149.92417661   51.23231845
 -340.38564125 -188.29571397  618.83965438]
```

So, seems like the flux **now** scales with the element size. But it should not, right?

Just in case we need it later: the current hack does not work for a different basis than linear finite elements.

uekerman:
> I have the feeling that we need the scaling somewhere else (on the FEniCS side?).

False alarm. My mistake. I solved it in Nutils in https://github.com/precice/tutorials/pull/153/commits/5783ae56fbb10bb3ca4a5e8e9808ca248cc645b8

uekerman:
@BenjaminRodenberg I also formatted the code back to two spaces. If we want sth different, let's format all Python codes together after this PR.

partitioned-heat-conduction/nutils/run.sh
```diff
@@ -0,0 +1,13 @@
+while getopts ":dn" opt; do
+  case ${opt} in
+    d ) rm -rf Dirichlet-*.vtk
```
BenjaminRodenberg, line 3: I would like to remove the `rm` statements to make `nutils/run.sh` consistent with `fenics/run.sh`, but I think we cannot do this. Nutils does not override existing files, but appends an integer, if there is already a file of the same name (Example: `File` exists, then we get `File-1`, `File-2`...). If I don't remove existing files before running the code again, I get the following files:

```
/tutorials/partitioned-heat-conduction/nutils$ ls
clean.sh         Neumann-0-2.vtk   Neumann-10-2.vtk  Neumann-1-2.vtk  Neumann-2-2.vtk  Neumann-3-2.vtk  Neumann-4-2.vtk  Neumann-5-2.vtk  Neumann-6-2.vtk  Neumann-7-2.vtk  Neumann-8-2.vtk  Neumann-9-2.vtk                  precice-Neumann-events.json         run.sh
heat.py          Neumann-0.vtk     Neumann-10.vtk    Neumann-1.vtk    Neumann-2.vtk    Neumann-3.vtk    Neumann-4.vtk    Neumann-5.vtk    Neumann-6.vtk    Neumann-7.vtk    Neumann-8.vtk    Neumann-9.vtk                    precice-Neumann-events-summary.log
Neumann-0-1.vtk  Neumann-10-1.vtk  Neumann-1-1.vtk   Neumann-2-1.vtk  Neumann-3-1.vtk  Neumann-4-1.vtk  Neumann-5-1.vtk  Neumann-6-1.vtk  Neumann-7-1.vtk  Neumann-8-1.vtk  Neumann-9-1.vtk  precice-Neumann-convergence.log  precice-Neumann-iterations.log
```
This totally confuses paraview. 

There is nothing we can do here, but I already opened an issue in the nutils repo: https://github.com/evalf/nutils/issues/624

BenjaminRodenberg (commented):
To me everything looks like it is ready to merge or am I missing something?

partitioned-heat-conduction/nutils/run.sh
uekerman: Let's not make this a show stopper here. The workaround works for the moment.

PR #145, opened 2021-01-08
Restructure quickstart

DavidSCN:
Open TODOs:
- [x] update README
- [x] update image (essentially an older TODO)

PR #158, opened 2021-03-11
Update README.md

ajaust:
Bumped preCICE version to most current release.


DavidSCN:
Thanks for pointing this out @ajaust. Howeverm, the quickstart has not yet been restructured (in the restructure branch) and your patch is already included in #145.

ajaust:
Ah, that explains why I got this form from a weird branch. I just clicked the "Change me on GitHub" button from the preCICE website. ;)

PR #160, opened 2021-03-17
Release tutorials

uekerman:
This mainly concerns the restructuring: https://github.com/orgs/precice/projects/5

## TODO:
* [x] Update README (done in #162)
* [x] Merge and final test of quickstart (as a very last step) https://github.com/precice/tutorials/pull/145
* [x] Finish https://github.com/orgs/precice/projects/5
* [x] Check website pages for consistency
* [x] Check all `precice-config.xml`s for consistency

### Test all cases manually and compare results where possible
* [x] OpenFOAM (@DavidSCN)
* [x] FEniCS (@IshaanDesai )
* [x] CalculiX (@IshaanDesai )
* [x] deal.II (@DavidSCN )
* [x] Nutils (@DavidSCN )
* [x] SU2 (@IshaanDesai )
### Checklist for the manual tests:
- I can run my case using the `run.sh`
- I can clean the case using `clean.sh` (maybe wait for @MakisH ) 
- the result looks reasonable
- the participant is documented in the `README` file on the website

## While merging this PR:
* [ ] Redirect website to master and pull submodule (after the release)
* [x] Release all adapters (if breaking changes)
* [x] Update the landing page of tutorials, cf. https://github.com/precice/tutorials/pull/162#issuecomment-802014154

PR #162, opened 2021-03-18
Remove 'not yet ported warning'

DavidSCN:
to be merged at the very end when we merge #160. The website introduction site needs to be adjusted as well.


uekerman:
> The website introduction site needs to be adjusted as well.

Should we do this here?

DavidSCN:
We can't, the site is located here https://github.com/precice/precice.github.io/blob/master/pages/tutorials/tutorials.md

MakisH:
We need to first merge develop and apply the same change for partitioned-pipe and anything else that we may have added since.

DavidSCN:
Done. The `elastic-tube-1d` has still this warning, but there is no valuable README/tutorial at all.

PR #163, opened 2021-03-18
Add checks for permalinks and image names

fsimonis:
This workflow adds a basic check for tutorials:
* permalinks of pages have to start with `tutorials-NAME` (this allows `tutorials-epic.html`, but also `tutorials-epic-introduction.html`)
* image names have to strart with `tutorials-NAME-`

This currently ignores the quickstart.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,12 @@
+name: Checks
+on: [push, pull_request]
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Check image names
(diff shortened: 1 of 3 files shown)


MakisH (approved):
I understand that this adds a GitHub action that checks:
- every file under each `(tutorial)/images` for starting with `tutorials-(tutorial)-`
- every `.md` file for the same (I am not sure I understand why we need two checks, though. Is it because of the trailing `-`?)

Since this is only a check, let's add it and see how it evolves. I think it will be useful.
We should then also consider it in our contributing guidelines #164.

fsimonis:
@MakisH looking for `images` makes the extraction of relevant cases in bash simpler.
Merging the checks into a single file is probably a good future idea.

PR #165, opened 2021-03-18
Merge checks and add support for quickstart

fsimonis:
This PR:
* merges both checks for images and permalinks into a single script
* adds support for the quickstart


MakisH (approved):
With a quick look, it looks good (I guess you tested it already).

PR #166, opened 2021-03-18
Update deal.II scripts

DavidSCN:
Related to precice/dealii-adapter#50 and precice/dealii-adapter#51.

@uekerman Now you have an example for the new version. I hope this helps in order to review the changes. In the dealii-adapter, please review first precice/dealii-adapter#51 and then precice/dealii-adapter#50.


perpendicular-flap/solid-dealii/run.sh
```diff
@@ -7,21 +7,5 @@ cd ${0%/*} || exit 1    		    # Run from this directory
 # in another terminal.
 # These scripts present how the two participants would be started manually.
 
-# 1 for true, 0 for false
-nonlinear=0
-if [ "$1" = "-nonlinear" ]; then
-            nonlinear=1
-fi
+./elasticity parameters.prm
```
uekerman, line 10: currently `dealii-precice`, needs probably an update in the adapter

perpendicular-flap/solid-dealii/parameters.prm
```diff
@@ -14,29 +14,28 @@ subsection Time
...
   set rho             = 3000
 
-  # body forces x,y,z
+  # Body forces x,y,z
   set body forces     = 0.0,0.0,0.0
 end
 
-subsection Linear solver
+subsection Solver
+  # Structural model to be used: linear or neo-Hookean
```
uekerman, line 36:
```suggestion
  # Structural model to be used: linear or neo-Hooke
```
At least that's what the adapter wants right now.

perpendicular-flap/solid-dealii/parameters.prm
```diff
@@ -14,29 +14,28 @@ subsection Time
...
+  set Model                     = linear
+
   # Linear solver iterations (multiples of the system matrix size)
   set Max iteration multiplier  = 1
```
uekerman, line 40: This option is hard to understand. Needs more documentation I guess. This is only needed if CG is chosen below, right? We could add such restrictions in the comment: "ignored for Direct" or similar.

perpendicular-flap/solid-dealii/parameters.prm
```diff
@@ -45,6 +44,15 @@ subsection Linear solver
 
   # Linear solver: CG or Direct
   set Solver type               = Direct
+
+  # Number of Newton-Raphson iterations allowed
+  set Max iterations Newton-Raphson = 10
```
uekerman, line 49: ignored if "Model = linear" or similar.

perpendicular-flap/solid-dealii/parameters.prm
```diff
@@ -45,6 +44,15 @@ subsection Linear solver
```
uekerman, line 44: What does "scaled by residual norm" mean exactly? Is this a relative residual?
Why relative here and absolute for the non-linear iteration?

perpendicular-flap/solid-dealii/parameters.prm
```diff
@@ -45,6 +44,15 @@ subsection Linear solver
...
+
+  # Number of Newton-Raphson iterations allowed
+  set Max iterations Newton-Raphson = 10
+
+  # Displacement error tolerance
+  set Tolerance displacement        = 1.0e-6
+
+  # Force residual tolerance
+  set Tolerance force               = 1.0e-9
 end
```
uekerman, line 56:
```suggestion
  # Absolute displacement error tolerance for non-linear iteration (ignored if Model == linear)
  set Tolerance displacement        = 1.0e-6

  # Absolute force residual tolerance for non-linear iteration (ignored if Model == linear)
  set Tolerance force               = 1.0e-9
end
```

uekerman (changes requested):
This review includes https://github.com/precice/dealii-adapter/pull/50 and https://github.com/precice/dealii-adapter/pull/51

I did not check the code details, but the user side is nice. Much easier to handle now :+1:
I was able to run linear and non-linear.
Only a few things we could document better in the parameters file.
(why is there a `parameters.prm` in the adapter repo?)

And the README of the adapter needs to be updated.

perpendicular-flap/solid-dealii/parameters.prm
DavidSCN: It's an absolute residual, the non-linear solver below uses relative residuals.

perpendicular-flap/solid-dealii/parameters.prm
DavidSCN: These residuals are relative, I added a comment.

DavidSCN:
> why is there a parameters.prm in the adapter repo?

It's needed to run a self created case.

>  And the README of the adapter needs to be updated.

Done.

uekerman:
> > why is there a parameters.prm in the adapter repo?
> 
> It's needed to run a self created case.

You could also point to the tutorials, right? In the end, it's just some code duplication, so another place to update when things change. Or what do you mean by self created case? You will always need another participant and a `precice-config.xml`, right?

multiple-perpendicular-flaps/solid-right-dealii/parameters.prm
```diff
@@ -9,54 +9,50 @@ subsection Time
...
+  set Model                     = linear
 
   # Linear solver: CG or Direct
   set Solver type               = Direct
-end
 
-subsection Nonlinear solver
-  # Number of Newton-Raphson iterations allowed
+  # Max CG solver iterations (multiples of the system matrix size)
+  # In 2-d, this value is best set at 2. In 3-d, a value of 1 work fine.
```
uekerman, line 43:
```suggestion
  # In 2D, this value is best set at 2. In 3D, a value of 1 works fine.
```

multiple-perpendicular-flaps/solid-right-dealii/parameters.prm
```diff
@@ -9,54 +9,50 @@ subsection Time
...
+  set Max iteration multiplier  = 1
+
+  # Absolute CG solver residual (multiplied by residual norm, ignored if Model == linear)
```
uekerman, line 46:
```suggestion
  # Absolute CG solver residual (multiplied by residual norm, ignored if Solver type == Direct)
```
right?

DavidSCN: No, the linear system is solved more accurate when approaching the solution of the nonlinear system. I hard coded it to 1e-10 for the linear model. Does it make sense for you?

DavidSCN:
> > > why is there a parameters.prm in the adapter repo?
> > 
> > 
> > It's needed to run a self created case.
> 
> You could also point to the tutorials, right? In the end, it's just some code duplication, so another place to update when things change. Or what do you mean by self created case? You will always need another participant and a `precice-config.xml`, right?

_hm_ The way I think of it is more a parameter file reference, similar to our xml reference. The cases in the tutorials are special configurations for a certain example. In fact, I could remove half of the parameters, where the nonlinear solver is not supposed to be used.

uekerman (approved):

PR #170, opened 2021-03-25
Move removeObsoleteFolders.sh to a function

MakisH:
We currently have many duplicate `removeObsoleteFolders.sh` scripts (related to https://github.com/precice/openfoam-adapter/issues/26). This replaces all these scripts with a central shell function under `tools/`.

It also fixes a few potential bugs (thanks to shellcheck).

@DavidSCN FYI (we can use the same in the Quickstart). We will also need to adapt the heat-exchanger case.


DavidSCN (approved):
Nice, thanks. I didn't like the file in that many places already when creating the tutorials already, much better now :rocket: 
Just asked shellcheck as well, thanks for the recommendation!

PR #174, opened 2021-03-25
Rename and check gnuplot scripts

MakisH:
- Renamed `plotDisplacement.sh` to `plot-displacement.sh` to stay consistent with the rest of the new names.
- Reduced the `/bin/bash` requirement to `/bin/sh` (more portable, no bash features anyway).
- Fixed one quotes-related issue in `perpendicular-flap/plotDisplacement.sh` (thanks, shellcheck!)


DavidSCN (approved):
Thanks

PR #175, opened 2021-03-26
Update deal.II parameter files

DavidSCN:
Related to precice/dealii-adapter#54

..and centralize run scripts.


tools/run-dealii.sh
```diff
@@ -0,0 +1,41 @@
...
+# If it is in the global path
+if [ -n "$(command -v "${SOLVER}")" ]; then
+   "${SOLVER}" parameters.prm
+fi
+
+# If it has been copied to the local directory
+if test -f "elasticity"; then
+    ./"${SOLVER}" parameters.prm
+    else
+    echo "Unable to find the executable ${SOLVER}. Either specify a prefix (-p=/path/to/elasticity) or make it discoverable at runtime (e.g. export PATH)"
```
MakisH, line 39: I like the idea of passing the prefix as an option, but wouldn't it be more natural to directly point at the file itself? Then the option could be `-e --exec`.
This could also let us use the `-p` for the parameter file (imagine comparing two different simulations).

This could then also merge the two `if` statements, and lead to further simplifications I guess.

MakisH (commented):
Thanks a lot, you are one step ahead of me! I wanted to take care of the run scripts today. Here are some comments to consider.

tools/run-dealii.sh
DavidSCN: Maybe @BenjaminRodenberg can let a comment here. [The suggestion](https://github.com/precice/dealii-adapter/pull/50#issuecomment-802107615) was to use a prefix instead of an executable.

I don't get the comparison idea. In general, the paramter file is parsed by deal.II (and I don't want to change it again). Could you give an example?

BenjaminRodenberg: I think the `-e --exec` suggestion sounds good.

DavidSCN: Ok, have a look at the result. If you have any other simplifications in mind, please go ahead.

MakisH: What I had in mind was something like:
```
EXE="./elasticity"
if (provided by -e)
   update EXE

run EXE without checking what it is
```

But this also already looks good enough.

DavidSCN: Is this an approval? Or do you want me to change anything?

MakisH: If you don't want to think more of it right now, then it is an approval. :-)

MakisH (approved):

tools/run-dealii.sh
DavidSCN: In the end, we just need to execute the solver here. There is probably a more fancy way, but let's keep it as it is.

BenjaminRodenberg:
I just tried the new `-e` option and got the following error:

```
~/tutorials/perpendicular-flap/solid-dealii$ ./run.sh -e=/home/benjamin/dealii-adapter/linear_elasticity/linear_elasticity
-----------------------------------------------------------------------------
--     . adapter revision f9c2c65 on branch develop
--     . deal.II 9.2.0 (revision unknown)
-----------------------------------------------------------------------------



----------------------------------------------------
Exception on processing: 

--------------------------------------------------------
An error occurred in line <538> of file </build/deal.ii-vFp8uU/deal.ii-9.2.0/source/base/parameter_handler.cc> in function
    virtual void dealii::ParameterHandler::parse_input(const string&, const string&, bool, bool)
The violated condition was: 
    is
Additional information: 
    The file "linear_elasticity.prm" was not found in the path for files of class ParameterHandler.
--------------------------------------------------------

Aborting!
----------------------------------------------------
```

This looks to me like the `linear_elasticity.prm` file is not found correctly. I tried to solve the problem by copy-pasting the file to the directory from which I'm running the case, but ran into the next problem.

I finally went back to an older revision (a166efadfe7dbd3231e14897dcbeecebb90ea97e) ran the simulation by adding the folder to the path.

DavidSCN:
@BenjaminRodenberg please update the dealii-adapter develop branch. The executable has been renamed.

PR #180, opened 2021-04-10
Make cleaning scripts less strict

### LINKED ISSUE #179: Make cleaning tools available for the command line
We unified our cleaning strategies in #169 and all this works really well if we use the scripts/symlinks located in the individual directories. However, it would be very nice to have these tools also available on the command line so that I don't need to create any symlinks when working with custom cases. 
This is not really be meant to be used by our users but more as an optional thing for developer and everyone who would like to use it this way. In practice, it would already be possible, but the initial ` set -e -u` triggers an exit in case something goes wrong. Opinions on this @MakisH ?

MakisH:
In #169, I created some function-only files with tools to clean tutorial cases and to remove obsolete OpenFOAM directories. Trying to follow "safety first", I set `-e -u` in every script, including the function ones. This made it very unpractical to source globally.

I now moved these options inside each function.

This means that one can now source these scripts in their `~/.bashrc`:
```
source /path/to/tools/cleaning-tools.sh
source /path/to/tools/openfoam-remove-empty-dirs.sh
```
and use them from inside the directory which they want to clean as:
```
clean_tutorial
clean_openfoam
...
openfoam_remove_empty_dirs
```
Tab-completion also works.

I am not sure where to document this.

@DavidSCN does this close #179 for you? I tried once more making the scripts runnable from any path, but I think there is no clean way to do it in simple POSIX-compliant shells.

P.S.: This also removes two `TODO` comments that should actually be addressed only if an issue arises.

### DIFF
--- a/tools/openfoam-remove-empty-dirs.sh
+++ b/tools/openfoam-remove-empty-dirs.sh
@@ -1,6 +1,7 @@
 #! /bin/sh
 
 openfoam_remove_empty_dirs() {
+	(
 	set -e -u
 	echo "Looking for any time directories without results (e.g. stray functionObjectProperties files, see openfoam-adapter issue #26 on GitHub)..."
 
@@ -20,4 +21,5 @@ openfoam_remove_empty_dirs() {
 			cd ..
 		done
 	fi
+	)
 }
--- a/tools/openfoam-remove-empty-dirs.sh
+++ b/tools/openfoam-remove-empty-dirs.sh
@@ -1,7 +1,7 @@
 #! /bin/sh
-set -e -u
 
 openfoam_remove_empty_dirs() {
+	set -e -u
 	echo "Looking for any time directories without results (e.g. stray functionObjectProperties files, see openfoam-adapter issue #26 on GitHub)..."
 
 	for f in [0-9]* [0-9]*.[0-9]*; do
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-set -e -u
 
 error() {
     echo "Error: $1" >&2; exit 1;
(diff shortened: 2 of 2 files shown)


tools/openfoam-remove-empty-dirs.sh
```diff
@@ -1,7 +1,7 @@
 #! /bin/sh
-set -e -u
 
 openfoam_remove_empty_dirs() {
+	set -e -u
```
DavidSCN, line 4: Similar to the other functions, we need brackets here so that the `set` statement is in a local scope.

DavidSCN (approved):
Thanks! Apart from one comment (see below) everything works fine and it closes #179 . Sourcing it makes it also runnable from any arbitrary path without any argument.
Where to document it is a good question. The best place is probably somewhere in the tutorial docs. Maybe we can just add a documentation when we are sure it works and find a good place during the next week.

tools/openfoam-remove-empty-dirs.sh
MakisH: Good point, there is no reason not to start this in a subshell. I fixed it.

MakisH:
> Sourcing it makes it also runnable from any arbitrary path without any argument.

This should only work for `openfoam_remove_empty_dirs`, the others should always need an argument (the path to clean). But what the scripts clean is still the directory where they are executed from (i.e. it can be used in an unintended way).


> Where to document it is a good question. The best place is probably somewhere in the tutorial docs. Maybe we can just add a documentation when we are sure it works and find a good place during the next week.

I also believe we need to see the full picture first. It can probably go in the [contributing guidelines for tutorials](https://www.precice.org/community-contribute-to-precice.html#contributing-tutorials) (or in a `CONTRIBUTING.md` here).

PR #181, opened 2021-04-10
Restructure runscripts and related resources

### LINKED ISSUE #171: heat-exchanger OpenFOAM is missing run.sh
https://github.com/precice/tutorials/tree/develop/heat-exchanger

Previous `Allrun`: https://github.com/precice/tutorials/blob/master/CHT/heat_exchanger/buoyantSimpleFoam-CalculiX/Allrun

It also specifies CalculiX v2.12 in the README, which we should remove.

MakisH:
Similarly to #169 and #166, this collects all the run scripts into common files, to reduce duplication.

- [x] OpenFOAM: separated between a "preparation" and an "execution" phase. The execution is same for every case, the preparation can differ. I did not find any reason to use a function, so I just call the general script from inside each case-specific script.
   - I also took the opportunity to clean up some old crimes in the heat exchanger, so this also closes #171. 
   - Since this case is special because we download some additional files, I added these extra files in smaller `.gitignore` files, so this also rouches aspects of #87.
- [x] deal.II: no changes.
- [x] CalculiX: I only removed unnecesary/inconsistent comments and variables. Some cases with large meshes already set (default) values for OpenMP, I assume that these actually work and can be useful.
- [x] SU2: We only have one case, I left it hard-coded and simplified it.
- [x] FEniCS: In some cases, I added a `run.sh` or a `clean.sh` that was missing. I guess that some of the cases are not yet in the fully restructured form, but these are elements we need anyway. I have not checked if the runscripts run. I also softened a bit an existing `.gitignore` file for a FEniCS case.
- [x] Nutils: Similarly to FEniCS. I also extended the `clean_nutils` to remove all `./*.vtk` files. Not tested.
- [x] code_aster: Only formatting and consistency changes.

All scripts have been validated with shellcheck.

Before merging:
- [x] Check documentation for renamed scripts: plot displacements, download meshes, run.sh / runFluid / Allrun, ...
- [x] Check documentation for `0.orig` (moved to `0` in most cases) . What happens when we use the cleaning scripts of the foundation version?

After merging:
- Update Quickstart (@DavidSCN would be a good review/cross-check that you apply the same changes there and I review the separate PR)


flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
```diff
@@ -1,40 +1,8 @@
...
-if [ "$1" = "-parallel" ]; then
-    parallel=1
-fi
-
-echo "Preparing and running the Fluid participant..."
-
-rm -rfv 0/
-cp -r 0.orig/ 0/
 blockMesh
-checkMesh
```
DavidSCN, line 21: Removing this is very much appreciated, since it checks all result files as well.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
```diff
@@ -1,40 +1,8 @@
...
+rm -rfv 0/ && cp -r 0.orig/ 0/
```
DavidSCN, line 5: It might be better to use symbolic links here. Sometimes it happens that I (and probably others as well) edit the files in the `0` directory and the get deleted when starting the simulation. A symbolic links resolves the issue.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
```diff
@@ -1,40 +1,8 @@
...
-else
-    $solver
-fi
-
-# Workaround for issue #26
-. ../../tools/openfoam-remove-empty-dirs.sh && openfoam_remove_empty_dirs
-
-echo ""
-echo "### NOTE ### Make sure to use the correct solver for your OpenFOAM version! (pimpleFoam for OpenFOAM v1806, OpenFOAM 6, or newer, vs pimpleDyMFoam for older) You may change this in your Fluid/system/controlDict file, if needed."
+../../tools/run-openfoam.sh "$@"
```
DavidSCN, line 8: One general comment (I already observed it in the cleaning tools): is there a specific reason to suppress a newline at the end of each file? Some editors (e.g. vim) automatically insert a newline at the of a file (which can probably be disabled). But this leads to git-visible changes anytime you open and save the file.

DavidSCN (commented):
Very nice! I guess your checklist keeps track of already checked run scripts and we can check tomorrow if everything works.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
MakisH: I mainly removed it because it does not make sense for a tutorial, it only makes sense when developing a case. An OpenFOAM user that will modify the mesh will probably already know about this and will want to execute it only once.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
MakisH: I thought of completely dropping the `0.orig`, but I noticed that `cleanCase` of OpenFOAM removes it (which actually should only be done by `cleanCase0`. I keep this open to have another look.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
MakisH: Good point, which we could only fix by formatting. I have not configured my editor to trim new lines at the end of the file, I guess it is just a coincidence.

I reformatted all files of this PR with VSCode for now (extension `foxundermoon.shell-format`), we could later add some project-specific configuration, but we have such a variety of file formats here that I don't know if this will work as expected.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
DavidSCN: In my OpenFOAM environment, `0` is not deleted and `0.orig` is not picked up automatically, in case `0` is not present.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
DavidSCN: Thanks! I guess a shell formatting might be a bit over-engineering. The newline at the end is all vim (and other editors) do, so this works now very well.

MakisH: It makes sense, I guess it was just because of my workflow: I select the complete content of the source and copy, I select the complete content of the target and paste and then edit. If the source was wrong, the rest would be wrong as well. I usually turn off all the editor-specific "format on save" actions to not conflict with what is already in the project/file, but for new files it can backfire.

I also formatted the clean scripts, since anyway this is about scripts.

flow-over-heated-plate-nearest-projection/solid-openfoam/run.sh
MakisH: This is strange. With OpenFOAM v2012, if I `mv 0.orig 0` and then run `./clean.sh` I get:
```
tutorials/perpendicular-flap/fluid-openfoam [runscripts]$ ./clean.sh 
--- Cleaning up OpenFOAM case in /home/makish/github/tutorials/perpendicular-flap/fluid-openfoam
Cleaning case /home/makish/github/tutorials/perpendicular-flap/fluid-openfoam
---- Cleaning up preCICE logs in /home/makish/github/tutorials/perpendicular-flap/fluid-openfoam
removed './0/U'
removed './0/pointDisplacement'
removed './0/p'
removed './0/phi'
removed directory './0/'
~/github/tutorials/perpendicular-flap/fluid-openfoam [runscripts]$ ls
clean.sh  constant  run.sh  system
```

What do you get?

MakisH: Sorry, wrong, I had it in `clean.sh`. Nothing to check.

DavidSCN: Still, I guess the behavior might be different for the foundation versions. I use the ESI version as well.

MakisH: @DavidSCN thank you for raising this again. I dropped the `0.orig` from all cases apart from the `turek-hron-fsi3` (which I guess modifies `0/U` with `groovyBC`, but I don't have `groovyBC` installed on this system to check).

DavidSCN: I have an installation and just checked it: actually no, the `0/U` file stays exactly the same. Let's remove the `orig` file here as well.

MakisH: Thanks! Moved it there as well. Everything starts getting extremely simpler (what were we even doing all this time?).

MakisH:
@DavidSCN ready for review! If you can easily run something with Nutils/FEniCS, please do. Other than than, any comments are welcome. Especially: did I forget anything? Is there anything in the documentation to fix?

tools/cleaning-tools.sh
```diff
@@ -40,7 +41,7 @@ clean_calculix() {
         set -e -u
         cd "$1"
         echo "--- Cleaning up CalculiX case in $(pwd)"
-        rm -fv ./*.cvg ./*.dat ./*.frd ./*.sta ./*.12d spooles.out
+        rm -fv ./*.cvg ./*.dat ./*.frd ./*.sta ./*.12d spooles.out dummy
```
MakisH, line 44: This `dummy` is generated at least in the heat exchanger CalculiX case, I have no idea where it comes from.

tools/cleaning-tools.sh
```diff
@@ -71,7 +72,7 @@ clean_fenics() {
         set -e -u
         cd "$1"
         echo "--- Cleaning up FEniCS case in $(pwd)"
-        rm -fv ./*.pvd spooles.out FSI-S/*
+        rm -fv ./*.pvd spooles.out out FSI-S/*
```
MakisH, line 75: This `out` I saw in an `Allclean`, I don't know if it is up to date.

tools/cleaning-tools.sh
```diff
@@ -81,6 +82,7 @@ clean_nutils() {
         set -e -u
         cd "$1"
         echo "--- Cleaning up Nutils case in $(pwd)"
+        rm -fv ./*.vtk
```
MakisH, line 85: Now we delete all the VTK files inside Nutils cases. I guess not much can go wrong.

DavidSCN (commented):
A two more comments:
- the OpenFOAM run ends for me with the following message `../../tools/run-openfoam.sh: 17: .: openfoam-remove-empty-dirs.sh: not found
`
- I created the deal.II run script as symbolic link similar to our cleaning scripts. Do you want to keep it in case of OpenFOAM as it currently is?

I will adjust the quickstart in #145 and we can check the remaining parts during the coding days :)

MakisH:
>     * the OpenFOAM run ends for me with the following message `../../tools/run-openfoam.sh: 17: .: openfoam-remove-empty-dirs.sh: not found `

Good catch, I fixed it by moving the cleaning part into `run.sh` as a separate step.
 
>     * I created the deal.II run script as symbolic link similar to our cleaning scripts. Do you want to keep it in case of OpenFOAM as it currently is?

In the way I see it, we cannot do this for OpenFOAM: some cases have different preparation steps and for all of them I create a named `.foam` file (opening two .foam files in ParaView should keep it clear what is what).

> I will adjust the quickstart in #145 and we can check the remaining parts during the coding days :)

Then I don't wait for the Quickstart :).

tools/cleaning-tools.sh
MakisH: It is up to date and it is a directory: https://www.precice.org/tutorials-partitioned-heat-conduction.html#visualization
This would not work, I moved to a recursive `rm` step.

MakisH:
This PR is ready from my side. I also looked at the diff and it is as intended. Are there further comments or things to check?

DavidSCN (approved):

PR #183, opened 2021-04-12
Extend checkpointing in nutils cases to subcycling

uekerman:
@BenjaminRodenberg looks good?
I tested both cases already. The vtk logging is then screwed up, but that's not soo critical here.


BenjaminRodenberg (approved):
Looks reasonable to me :+1:

PR #196, opened 2021-04-13
Add reference results for perpendicular flap

uekerman:
I am not sure yet whether we want to merge this in the end or not. Opinions?

reference

I guess all as expected.
We still have the issue with CalculiX that only first order elements are supported.
And SU2 is compressible vs. incompressible for OpenFOAM and Nutils.


IshaanDesai:
My opinion is that there are enough reasons to defend the non-matching results for every case and they are outweighed by the advantage of showing that we have results with the same trend for a wide range of solvers.

MakisH:
Is there an option to add some kind of comments on the plot? It would be good to document which solution we accept as closest to the truth (if any) and the important differences between each solver. Essentially what you wrote in the PR description.

BenjaminRodenberg:
Before merging:

* Remove the data (we don't have a systematic approach to storing data yet)
* Add instructions how to create and (if necessary) rename watchpoint files from runs
* Add plot to test case description in the visualization section of the `README.md`

uekerman:
I added some documentation in the README.
I will still delete the reference results (so the subfolder `watchpoints`).
OK like this?

perpendicular-flap/README.md
```diff
@@ -51,7 +51,18 @@ in order to use OpenFOAM and FEniCS for this test case.
...
+As we defined a watchpoint on the 'Solid' participant at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh.` You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-fenics`. The resulting graph shows the x displacement of the flap tip. You can modify the script to plot the force instead.
 
 
 ![Flap watchpoint](images/tutorials-perpendicular-flap-displacement-watchpoint.png)
+
+There is moreover a script `plot-all-displacements.sh` to plot and compare all possible variants. This scripts expects all watchpoint logs to be available in a subfolder `watchpoints` in the format `openfoam-dealii.log` or similar. If you want to use this script, you need to copy the files over accordingly.
```
BenjaminRodenberg, line 59:
```suggestion
There is moreover a script `plot-all-displacements.sh` to plot and compare all possible variants. This script expects all watchpoint logs to be available in a subfolder `watchpoints` in the format `openfoam-dealii.log` or similar. If you want to use this script, you need to copy the files over accordingly.
```

perpendicular-flap/README.md
```diff
@@ -51,7 +51,18 @@ in order to use OpenFOAM and FEniCS for this test case.
...
+
+You should get results similar to this one:
+
+![All flap watchpoints](images/tutorials-perpendicular-flap-displacement-all-watchpoints.png)
+
+Reasons for the differences:
+
+* The CalculiX adapter only supports linear finite elements (deal.II uses 4th order, FEniCS 2nd order).
+* SU2 models a compressible fluid, OpenFOAM and SU2 an incompressible one.
```
BenjaminRodenberg, line 68:
```suggestion
* SU2 models a compressible fluid, OpenFOAM and Nutils an incompressible one.  
```

BenjaminRodenberg (approved):
Looks good :+1:

MakisH (approved):
Looks good, I have not tested it.

PR #197, opened 2021-04-13
Rework elastic-tube-1d cpp solvers

fsimonis:
This PR:
* Modernizes and clarifies the solid-cpp and fluid-cpp solvers
* Removes memory leaks in both solvers
* Disables the parallel version of the fluid-cpp solver


MakisH (approved):
Thanks for the amazing work! I briefly looked at the code, I did not see anything suspicious, but the diff is of course very large.

While trying to build, I got some warnings for unused variables:
```
[ 25%] Building CXX object CMakeFiles/FluidSolver.dir/src/FluidSolver.cpp.o
/home/makish/github/tutorials/elastic-tube-1d/fluid-cpp/src/FluidSolver.cpp: In function ‘int main(int, char**)’:
/home/makish/github/tutorials/elastic-tube-1d/fluid-cpp/src/FluidSolver.cpp:32:16: warning: unused variable ‘tau’ [-Wunused-variable]
   32 |   const double tau        = atof(argv[3]);
      |                ^~~
/home/makish/github/tutorials/elastic-tube-1d/fluid-cpp/src/FluidSolver.cpp:54:16: warning: unused variable ‘E’ [-Wunused-variable]
   54 |   const double E = 10000; // elasticity module
      |                ^
```

Other than that, it runs, the results look good (very similar with Python). I also tried all four combinations, and I get similar results.

MakisH (changes requested):
Actually, together with the unused variables, I noticed that the help mesages and the run scripts are out of date. @fsimonis please comment if you are working on this at the moment, otherwise I can fix it.

fsimonis:
@MakisH could you reimplement the command line arguments?
They are also completely lacking in the python solvers, but I guess they should be easy to add there.

MakisH:
I am currently going the other way around and removing them from the C++ version, as we have not really tested that this works with more values. We could, in the future, add them again.

MakisH (approved):
I removed the additional command-line arguments from the C++ version, as we have not really tested if arbitrary values work and I don't want to open another box at this point. It is anyway not important at all.

I also removed the Elasticity module from both the C++ and Python solvers, it was not needed there.

Ready to merge! :tada:

PR #199, opened 2021-04-13
Postprocess and visualize CHT flow over plate results

BenjaminRodenberg:
Similar to #196, just for a different case. In the end we want to plot the temperature across the coupling interface.

# What kind of plot?

image

In the end we want a plot similar to the one above taken from [Fig3c from this paper](https://arxiv.org/pdf/2103.11191.pdf), but automatically generated from the preCICE output data. Note that this will restrict us to the coupling interface, i.e. the interval x = [0,1] and not the whole bottom of the channel x = [-0.5,3].

# What we get with this PR:

tutorials-flow-over-heated-plate-results-comparison


BenjaminRodenberg:
@oguzziya knows a different approach to this problem. Maybe worth switching to it?

oguzziya:
If the script is that long, it might not worth the hassle. However, if we gonna have different kinds of VTK files, such as unstructured, structured, etc. It might be a good approach. [PyVista](https://docs.pyvista.org/) is pretty handy to manipulate the data. For example, the following snippet is a code that I use in general to obtain a contour plot.

```python
def contour_plot(self, file_path, field_name, save_name=None, cmap="bone"):
     mesh = pv.read(file_path) 
    midplane = mesh.ctp().slice(normal=self.axis[self.directions["tangential"]], generate_triangles=True)
   
    midplane.active_scalars_name = field_name

    x = midplane.points
    tri = midplane.faces.reshape((-1,4))[:, 1:]
    data = midplane.active_scalars

    data_max = round(float(np.max(data)) / 10) * 10
    data_min = int(float(np.min(data)) / 10) * 10

# Rest is matplotlib
    fig = plt.figure(figsize=(20, 5))
    ax = fig.add_subplot(1,1,1)
    cont = ax.tricontourf(x[:, self.directions["axial"]], x[:, self.directions["radial"]], tri, data, levels=np.linspace(data_min, data_max, 39), cmap=cmap)
    cbar = fig.colorbar(cont, ax=ax)
    cbar.set_ticks(np.linspace(data_min, data_max, 5, endpoint=True))
    ax.set_yticklabels([])
    ax.set_yticks([])
    ax.set_label(field_name)
    cbar.set_label(field_name, fontsize=20)
    if save_name is not None:
      fig.savefig(save_name + ".png", dpi=300)
```

BenjaminRodenberg:
@oguzziya Never heard about `PyVista`. This looks so much more reasonable than plain `vtk` code (which is a nightmare). Thanks for the hint!

flow-over-heated-plate/README.md
```diff
@@ -57,6 +57,29 @@ You may use additional filters, such as the Calculator and the Plot Over Line, t
...
+### Compare results
+
+First generate the output for each run by adding export to `precice-config.xml`. You can just apply the following patch:
+
+```
+diff --git a/flow-over-heated-plate/precice-config.xml b/flow-over-heated-plate/precice-config.xml
+index 9c13374..c18a49e 100644
+--- a/flow-over-heated-plate/precice-config.xml
++++ b/flow-over-heated-plate/precice-config.xml
+@@ -34,6 +34,7 @@
```
uekerman, line 69: Isn't this a bit over-engineered? Will also only work if people start from the same commit, right?

flow-over-heated-plate/README.md
```diff
@@ -57,6 +57,29 @@ You may use additional filters, such as the Calculator and the Plot Over Line, t
...
+     </participant>
+ 
+     <participant name="Solid">
++      <export:vtk directory="preCICE-output" />
+       <use-mesh name="Fluid-Mesh" from="Fluid" />
+       <use-mesh name="Solid-Mesh" provide="yes" />
+       <mapping:nearest-neighbor
+```
+
+After that running a case from this tutorial will export data into `solid-*/preCICE-output`. To visualize and compare these results run `python3 plot-final-interface-temperature.py`. This will plot `theta = (T-300)/(310-300)` (with `T` being the temperature) across the coupling interface, i.e. where the solid and the fluid meet and exchange heat. The x-axis shows the x coordinate and the y-axis `theta` at the interface. If you want to exclude certain cases, simply comment out the corresponding lines in the script. For reference see below:
```
uekerman, line 79: Could you give a name what theta is? Some relative non-dimensional temperature?

uekerman (approved):
I like it :ok_hand: Works like a charm.
Small comments:
* Please add the dependency `vtk` in the README (and how to get it normally from PyPI). matplotlib as well?
* `example.vtk` can be deleted, right?

and see below.
Please merge after fixing these.

flow-over-heated-plate/plot-final-interface-temperature.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return data_dict
+
+
+cases = []
+cases.append('fenics')
+cases.append('openfoam')
+cases.append('nutils')
+
+case_labels = {'fenics': 'fenics-openfoam', 'openfoam': 'openfoam-openfoam', 'nutils': 'nutils-openfoam', }
```
uekerman, line 48:
```suggestion
case_labels = {'fenics': 'OpenFOAM-FEniCS', 'openfoam': 'OpenFOAM-OpenFOAM', 'nutils': 'OpenFOAM-Nutils', }
```
I am picky. But this would be more consistent with the perp flap.

flow-over-heated-plate/README.md
BenjaminRodenberg: Probably yes :laughing: 

I will just provide the complete block of code.

flow-over-heated-plate/README.md
BenjaminRodenberg: In the Vynnycky paper its called "dimensionless temperature". I will use this name.

PR #201, opened 2021-04-13
Fixes for solid-codeaster

### LINKED ISSUE #155: Port and test Code_Aster case
In #151 we did not yet completely port and test the "flow over heated plate steady state" with OpenFOAM and Code_Aster.
Does Code_Aster actually support 2D simulations? We also still need to update the mesh names.

MakisH:
Work together with @IshaanDesai. I guess that these changes are enough to make the code_aster adapter work with the new structure of the tutorials.

- [x] Add an empty `REPE_OUT` directory for the results.
- [x] Add a warning that the `exchange-directory` needs to be absolute.
- [x] Update `solid.astk` with new defaults.
- [x] Switch interface dimensions back to 3D.

Closes #155. Also related to https://github.com/precice/code_aster-adapter/pull/22.


MakisH:
@IshaanDesai Ready for review from my side, everything looks as intended.

IshaanDesai (approved):
Looks good :+1: I ran this case and checked the output in GMSH, it works.

PR #203, opened 2021-04-14
Add a check for the image sizes

DavidSCN:
Related to https://github.com/precice/precice.github.io/issues/48

The check works, but we have currently some files above the limit of 250 kb. So we either need to live with the failing check, decrease the corresponding image size further or increase the limit.


MakisH:
Where can I see the list of files remaining? Yesterday I resized most of them, I only kept the VM screenshot because I am going to update it very soon.

DavidSCN:
I reduced the verbosity of passing images in order to make it more obvious.
You can either click on `Details` in the GH CI or run `./tools/check-size.sh` in the tutorial repository.

MakisH:
I resized two PNG images that were left. Could you please exclude `.gif` files or set a different limit there? Something around 1.5MB-2MB should be enough.

DavidSCN:
Increased it to 2000 kb. The GH CI supports now also coloring.

MakisH:
The partitioned heat conduction GIF I already tried to reduce in size but I could not. I guess it makes sense to increase the threshold then (e.g. 3MB). We have almost no GIFs anyway.

DavidSCN:
Then let me increase it to 2.2. so that the test passes. If this becomes problematic in the future we can still increase the limit.

MakisH (approved):

Issue #188, opened 2019-02-05
elastic-tube-1d: Missing documentation elements for visualization


BenjaminRodenberg:
The documentation and code for visualization of the [PythonTube](https://github.com/precice/elastictube1d/tree/master/PythonTube) really need some polishing. Currently, it is really hard to understand how plots are created etc.

This is just a quick, unsorted collection of material:

* We can create VTK files with velocity and tube diameter. See [output.py](https://github.com/precice/tutorials/blob/master/elastic-tube-1d/fluid-python/output.py)
* ~~I do not know a nice way for using these VTK files in paraview.~~ The VTK files are used in plotting [here](https://github.com/precice/tutorials/blob/master/elastic-tube-1d/plot-vtk.py)
* ~~The script [fluid.py](https://github.com/precice/elastictube1d/blob/master/Postproc/fluid.py) allows us to create [this plot](https://raw.githubusercontent.com/wiki/precice/precice/images/elastictube_diameter.png). I do not think that this plot is really useful, but at least there exists some documentation on it [here](https://github.com/precice/precice/wiki/Running-the-1D-elastic-tube-example#postprocessing--visualization).~~ A newer plot is now available and provides much more relevant information regarding results.
* We have a nice [gif](https://raw.githubusercontent.com/wiki/precice/precice/images/tube.gif) of the elastictube. However, it is not documented, how this gif was actually created. With some insider knowledge, it is possible to find the code [here](https://github.com/precice/elastictube1d/blob/convergenceStudies/PythonTube/output.py#L172).

What would be nice: 

- [ ] ~~Merge the gif/video output from [`convergenceStudies`](https://github.com/precice/elastictube1d/tree/convergenceStudies) into `master`~~
- [x] Document: How do I create gif/video output? ([replace this paragraph](https://github.com/precice/precice/wiki/Running-the-1D-elastic-tube-example#postprocessing--visualization))
- [x] Find out how to make use of the VTK output or drop it.

IshaanDesai:
Closed in https://github.com/precice/tutorials/pull/209

PR #209, opened 2021-04-27
Improve elastic-tube-1d code and documentation

IshaanDesai:
Related to https://github.com/precice/tutorials/issues/188 but in the context of the new tutorials structure. The `elastic-tube-1d` was ported to the tutorials from a separate repository and this PR is a check if all the post-processing steps work as expected.

Closes https://github.com/precice/tutorials/issues/188


elastic-tube-1d/README.md
```diff
@@ -88,25 +88,19 @@ cd solid-python
 ./run.sh
 ```
 
-Parameters such as `N` can be modified directly at the `FluidSolver.py` and at the `SolidSolver.py`. The parameters must be consistent between the different solvers and participants.
-
-**Optional:** Visualization and video output of the fluid participant can be triggered via the options `--enable-plot` and `--write-video` of `FluidSolver.py`. To generate .vtk files during execution, you need to add the flag `--write-vtk`.
```
MakisH, line 93: Does this mean that these options have been removed from `FluidSolver.py`?

elastic-tube-1d/README.md
```diff
@@ -25,6 +25,8 @@ The following parameters have been chosen:
 - Fluid density: $$ \rho = 1 $$
 - Young modulus: E = 10000
 
+Additionally the solvers use the parameters `N = 100`, `tau = 0.01`, `kappa = 100` by default and can be modified in the solver.
+
```
MakisH, line 29: Somehow I cannot add a suggestion here, but minor tweak:

```
Additionally, the solvers use the parameters `N = 100`, `tau = 0.01`, `kappa = 100` by default. These values can be modified directly in each solver.
```

elastic-tube-1d/plot-diameter.sh
```diff
@@ -1,15 +1,17 @@
 #!/bin/sh
 set -e -u
 
+# File check solution from: https://stackoverflow.com/questions/91368/checking-from-shell-script-if-a-directory-contains-files
+
 # Plot diameter from fluid-cpp
-if [ -d "./fluid-cpp/output/" ]; then
+if [ -n "$(ls -A ./fluid-cpp/output/*.vtk 2>/dev/null)" ]; then
```
MakisH, line 7: I have not tried to run this, but it looks reasonable and `shellcheck` passes. Do you want me to check it? I understand that the problem was that the folder may exist, but the files not, in which case... `plot-vtk.py` crashed?

IshaanDesai: I tried this many times so you dont really need to check it explicitly. The problem was that the original command check if the folder `./fluid-cpp/output` exists or not, but it always exists, and hence the check always passes and then an error is thrown when no `*.vtk` files are found. To avoid this error I tried to check if the folder is empty or not, but that also does not work because the folder has a file `./fluid-cpp/output/.keepme` at all times. So now the check is explicitly for `.vtk` files and if none are found then the post processing is not done.

elastic-tube-1d/README.md
IshaanDesai: `.vtk` files are written even if the flag `--write-vtk` is not passed. Also, adding the flags `--enable-plot` or `--write-video` have no real change on the output obtained at the end or during the simulation. With this knowledge the code can be further cleaned up.

MakisH: But then it would be better to keep these options documented, right? Otherwise, they will quickly be forgotten. Of course, we could move them to a different location in the README.md.

IshaanDesai: The flag `--write-vtk` is passed inside the `run.sh` script so it is a bit confusing to know if the flag is actually passed or not. Now the flag `--write-vtk` has been removed because in any case one needs the `.vtk` files to see some output and other tutorials do not provide the optional choice of writing or not writing VTK files. Is this reasonable? 
`--enable-plot` option is necessary to see the video and will be reintroduced.

MakisH: So, in other words:
- `--write-vtk` is deprecated/broken/not even there
- we should always write VTK files
- the documentation then is not relevant anymore.

If yes, then I totally happy! :)

IshaanDesai: Yes thats exactly the state! Now the `fluid-python`always writes `.vtk` files, just as `fluid-cpp`

IshaanDesai:
Attempting to generate a video in `FluidSolver.py` by passing the flag `--write-video` leads to the following error:
```
MovieWriter stderr:
convert-im6.q16: delegate failed `'ffmpeg' -nostdin -v -1 -i '%M%%d.jpg' '%u.%m' 2> '%u'' @ error/delegate.c/InvokeDelegate/1928.

  File "./FluidSolver.py", line 161, in <module>
    writer.finish()
  File "/home/desaiin/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 354, in finish
    self.cleanup()
  File "/home/desaiin/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 390, in cleanup
    raise subprocess.CalledProcessError(
subprocess.CalledProcessError: Command '['convert', '-size', '640x480', '-depth', '8', '-delay', '6.666666666666667', '-loop', '0', 'rgba:-', 'writer_test.mp4']' returned non-zero exit status 1.
(0) 10:32:47 [impl::SolverInterfaceImpl]:147 in ~SolverInterfaceImpl: Implicitly finalizing in destructor
```
@MakisH could you please reproduce this to ensure it is not just something with my system? If this persists then I suggest temporarily suspending the video writing functionality and resolving this in another PR. This PR already has a lot of clean-up

MakisH:
```shell
python3 FluidSolver.py --enable-plot --write-video
```

works for me, I see the live animation. We should document this option.

I am using Python 3.8.6 and the Python binding v2.2.0.2.

IshaanDesai:
>  I see the live animation

I also see the live animation which is the succession of plots generated by `matplotlib`. But do you get a valid `.mp4` file in the directory at the end of the simulation? I see this error at the very end.

MakisH:
I get a valid video file, here it is:

https://user-images.githubusercontent.com/4943683/116394059-b9883d00-a822-11eb-84f2-8d2fe49522db.mp4

I don't see any errors even after completing (fluid-python + solid-python).

IshaanDesai:
> I get a valid video file

Okay great, in that case I add the `--write-video` flag information to the README and this PR is ready to be merged.

MakisH:
Maybe already try fixing the failing "Lint docs" pipeline to experience these new checks, it should be very easy. :-)

elastic-tube-1d/fluid-cpp/src/FluidSolver.cpp
```diff
@@ -72,11 +72,9 @@ int main(int argc, char **argv)
 
   interface.setMeshVertices(meshID, chunkLength, grid.data(), vertexIDs.data());
 
-  std::cout << "Initialize preCICE..." << std::endl;
-  interface.initialize();
-
   double t  = 0.0;
-  double dt = 0.01;
```
MakisH, line 79: Actually, shouldn't the solver be setting a time step size and then taking the minimum after advance? Here we don't see any problem as all time step sizes are the same, but it looks like this solver is always taking the time step size from the other participant. Is this intended here?

MakisH: Looking at the other solvers, it looks like we always take the coupling time step size, so this looks reasonable here. Ah, all the assumptions and simplifications of the elastic-tube-1d!

IshaanDesai: Yes, actually the following workflow should be there for the time step:
```
dt = 0.01
precice_dt = interface.initialize()

while (..)
  dt = min(dt, precice_dt)
```
But I think the simplification we have right now is okay.

IshaanDesai:
> Maybe already try fixing the failing "Lint docs" pipeline to experience these new checks, it should be very easy. :-)

Interestingly doing `act -j check_md` passes on my local machine but somehow fails in the remote deployment! Have you experienced this before?

MakisH:
> Interestingly doing `act -j check_md` passes on my local machine but somehow fails in the remote deployment! Have you experienced this before?

I have not yet had the chance to try out `act`, so no clue here. But I use the `davidanson.vscode-markdownlint` extension in VSCode and it shows the same errors (since all exceptions are already in the files or in the project markdownlint config.

IshaanDesai:
> I have not yet had the chance to try out `act`, so no clue here. But I use the `davidanson.vscode-markdownlint` extension in VSCode and it shows the same errors (since all exceptions are already in the files or in the project markdownlint config.

Thanks! `davidanson.vscode-markdownlint` extension caught the problem even though `act -j check_md` did not catch it.

MakisH (approved):

PR #213, opened 2021-05-05
Add Nutils to list of available solvers

DavidSCN:


MakisH:
The change is perfectly fine, my only question now is: what should we do with such fixes? Add them to `master` (and merge back to `develop`), but only release when deemed worthy?

This way, at least the website will be up to date and we don't have to release the whole develop or cherry-pick.

MakisH (approved):

DavidSCN:
Sounds good to me.

PR #214, opened 2021-05-11
Add OpenFOAM partitioned heat equation

### LINKED ISSUE #204: Missing partitioned heat conduction tutorial with OpenFOAM
In https://github.com/precice/openfoam-adapter/tree/HH there is a draft of a case that I created to investigate a "kink" we could see when plotting a line across interfaces. Related issue: https://github.com/precice/openfoam-adapter/issues/93

It would be helpful to integrate this case in the tutorials already, so that the community can also help investigating this.

I can eventually clean up and update the respective branch and transfer the case here.

MakisH:
This is an attempt to port the [2x-laplacianFoam case from the OpenFOAM adapter repository](https://github.com/precice/openfoam-adapter/compare/HH) and shape it to simulate the same problem as the FEniCS and Nutils cases that are already here.

**Change with side effects:** In the `precice-config.xml`, the data `Flux` needs to be renamed to `Heat-Flux`, as the OpenFOAM adapter uses the data name (prefix) to determine the data type. Naming it `Flux` in the first case sounds like a historical residue and does not agree with our `flow-over-heated-plate` case. This change means that the data needs to be renamed in the (case-specific) FEniCS and Nutils examples.

This is only an incomplete attempt, as I cannot manage to set the heat source and sink to be a single point/cell in OpenFOAM. I have tried different (less or more wrong) approaches:

- Define an L-shape `faceSet` and create a patch from it using `createPatch`. This is the setup I include here. It runs, but not as expected: the additional patch does not get any temperature values and ParaView complains. A problem may be that there is a conflict of boundary conditions for the same faces.
- Define a cellSet and set its internal field with `setFields`. Somehow this also gave weird results.
- Define a smaller box around the corner in `blockMeshDict`. I guess the problem there was that the blocks were overlapping.

A trivial workaround is to set the complete bottom & left walls of the `Dirichlet` participant as a heat sink and the complete top & right walls of the `Neumann` participant as a heat source (I think I even swapped the names accidentally). But this is then a different setup and the isolines have completely different shape.

@DavidSCN if you want, feel free to start from this and debug it.

TODO:

- [ ] Make the case work with point source & sink.
- [ ] Check the case parameters for consistency. I have set `T_sink = 1`, `T_source = 8`, `k = 3` (which should be `alpha`). I have no clue what to do with `beta`.
- [ ] Make the same changes also to the Neumann participant.
- [ ] Cleanup any unused files: `topoSetDict`, `createPatchDict`, unused settings in `blockMeshDict`/`T`, etc.
- [ ] Update FEniCS / Nutils for `Flux`/`Heat-Flux`
- [ ] Add entry in the `README.md`

The original motivation for this case was to study https://github.com/precice/openfoam-adapter/issues/93. The issue is observable also in the current state of the case.

This also closes #204.

### DIFF
--- a/partitioned-heat-conduction/fenics/heat.py
+++ b/partitioned-heat-conduction/fenics/heat.py
@@ -170,7 +170,7 @@ f.t = t + dt(0)
 
 if problem is ProblemType.DIRICHLET:
     flux = Function(V_g)
-    flux.rename("Flux", "")
+    flux.rename("Heat-Flux", "")
 
 while precice.is_coupling_ongoing():
 
--- a/partitioned-heat-conduction/fenics/precice-adapter-config-D.json
+++ b/partitioned-heat-conduction/fenics/precice-adapter-config-D.json
@@ -3,7 +3,7 @@
   "config_file_name": "../precice-config.xml",
   "interface": {
       "coupling_mesh_name": "Dirichlet-Mesh",
-      "write_data_name": "Flux",
+      "write_data_name": "Heat-Flux",
       "read_data_name": "Temperature"
     }
 }
--- a/partitioned-heat-conduction/fenics/precice-adapter-config-N.json
+++ b/partitioned-heat-conduction/fenics/precice-adapter-config-N.json
@@ -4,6 +4,6 @@
   "interface": {
       "coupling_mesh_name": "Neumann-Mesh",
       "write_data_name": "Temperature",
-      "read_data_name": "Flux"
+      "read_data_name": "Heat-Flux"
     }
 }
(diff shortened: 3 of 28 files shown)


DavidSCN:
After some investigations:
I don't think we can use `laplacianFoam` for this test case, since `laplacianFoam` solves the heat equation with a zero RHS whereas the described test case has a non-zero RHS. We could adjust the existing test case by applying the constraint `RHS=0` using alpha and beta (set beta=2+2*alpa), but I'm not sure if this would be intended.

MakisH:
Closing as duplicate of the now merged #223.

PR #215, opened 2021-05-31
Clarify perpendicular-flap deal.II documentation 

DavidSCN:
..still not sure if pointing here to `develop` is a good idea.

Now @MakisH, better?


MakisH (approved):
Yes, much clearer, maybe even too clear now! :sweat_smile: Thanks!

PR #216, opened 2021-06-02
Throw error in fenics and nutils run scripts if no argument was provided 

DavidSCN:
Nothing happens if no argument was provided, which is a bit unfortunate for people like me who are too lazy to read our documentation. This would be an easy fix.

### DIFF
--- a/partitioned-heat-conduction/fenics/run.sh
+++ b/partitioned-heat-conduction/fenics/run.sh
@@ -1,6 +1,14 @@
 #!/bin/sh
 set -e -u
 
+usage() { echo "Usage: cmd [-d] [-n]" 1>&2; exit 1; }
+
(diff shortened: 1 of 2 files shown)


DavidSCN:
@MakisH I guess this change could go to master as well?!

MakisH:
> @MakisH I guess this change could go to master as well?!

Is it a bugfix or a feature? If it is a bugfix, in principle it goes to master. If it is a feature/"nice to have", it goes to develop. The question is essentially how urgent and breaking it is.

I guess it can go directly to master.

DavidSCN:
You can decide yourself. It's just a usability improvement; not really a bug, not really a feature.

DavidSCN:
There is another commit I don't want to mix here. Merging to develop.

PR #220, opened 2021-06-20
Add a workflow to run tests nightly

MakisH:
(under construction)

### DIFF
--- a/.github/workflows/install-solvers-dependencies.sh
+++ b/.github/workflows/install-solvers-dependencies.sh
@@ -1,0 +1,8 @@
+#!/bin/bash -e
+
+# Install OpenFOAM v2012
+wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | sudo bash
+sudo apt-get install openfoam2012-dev
+
+# Install preCICE dependencies
+sudo apt-get install -y cmake libeigen3-dev libxml2-dev libboost-all-dev petsc-dev python3-dev python3-numpy
--- a/.github/workflows/run-all-develop.yml
+++ b/.github/workflows/run-all-develop.yml
@@ -1,0 +1,58 @@
+name: Run with all components from develop
+on:
+  schedule:
+    - cron: '0 4 * * *' # Every night at 4:00
+jobs:
+  install-dependencies:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Setup caches
+      uses: actions/cache@v2
+      id: cache-id
+      with:
+        path: ${{ runner.temp }}/cache-directory
+        key: ${{ runner.os }}-cache-dependencies-v1.0
+    - name: Install solvers (considering cache)
+      uses: airvzxf/cache-anything-new-action@v1.0.1
+      with:
+        script: 'install-solvers.sh'
+        is_cached: ${{ steps.cache-id.outputs.cache-hit }}
+        cache: ${{ runner.temp }}/cache-directory
+        snapshot: '/'
+        exclude: '/boot /data /dev /mnt /proc /run /sys'
+  build-all-develop:
+    needs: dependencies
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Build preCICE from develop
+      run: |
+        git clone --depth=1 --branch develop https://github.com/precice/precice.git
+        cd precice
+        git pull
+        mkdir -p build && cd build/
+        cmake -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -Wno-dev ..
+        make -j 2
+        make package
+        sudo apt-get install -y ./libprecice2_*.deb
+    - name: Build the preCICE Python bindings from develop
+      run: |
+        git clone --depth=1 --branch develop https://github.com/precice/python-bindings.git
+        cd python-bindings
+        python3 setup.py install --user
+    - name: Build the OpenFOAM adapter from develop for OpenFOAM v2012
+      run: |
+        git clone --depth=1 --branch develop https://github.com/precice/openfoam-adapter.git
+        cd openfoam-adapter
+        /usr/bin/openfoam2012 ./Allwmake
+    - name: Check out repository
+      uses: actions/checkout@v2
+      with:
+        ref: develop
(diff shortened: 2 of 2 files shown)


MakisH:
I did some progress on this in the past few hours. Current state:
- :heavy_check_mark: I build a preCICE + OpenFOAM + OpenFOAM adapter image based on Ubuntu 20.04 in `precice/openfoam-adapter` (https://github.com/precice/openfoam-adapter/pull/184) and publish it to the GitHub container registry.
- :heavy_check_mark: I start a workflow that runs two OpenFOAM solvers in the same tutorials container and archives the logs and case artifacts.
- :heavy_check_mark: The workflow starts and checks out the repository as `precice` ([not as `root`, that apparently the `checkout` action requires](https://stackoverflow.com/a/68761652/2254346))
- :heavy_check_mark: The tutorial starts and OpenFOAM finds the adapter (to achieve this, I had to manually set `USER` and `HOME`, I am not sure what problem is hidden behind this)
- :heavy_check_mark: Initialization works, the tutorial runs for a while
- :x: The tutorial seems to stop unexpectedly after a few iterations, but there seems to be no information in the logs about it and the job reports "success".

Next step after making the tutorial complete normally: setup `docker-compose` to start multiple containers and work together.

MakisH:
I added a similar step to run elastic-tube-1d, it also stops at an arbitrary time.  In the same job, OpenFOAM went further down, so there seems to be some inherent randomness involved.

MakisH:
:heavy_check_mark: Now the OpenFOAM-OpenFOAM tutorial runs correctly. I fixed the issue by adding a step to wait for the second participant (background) process to finish.

MakisH:
just trying to formulate a few thoughts regarding the system tests UID/GID issue that we discussed in the coding days:

- @fsimonis is right, mounting a volume on the container and writing files in it changes the owner to the container user. Running the systemtests locally like this can lead to a permission issue when running on a system where your user is not the first on the system (e.g., on a shared workstation).
- We need a [volume](https://docs.docker.com/storage/volumes/) to be able to write a commonly-visible `precice-run`. Tempfs is not visible across containers (and would probably complicate the artifact uploading), while binds are meant for read-only situations (managed by the host) and don't affect the UID/GID either.
- There are a few reasons I need to set a user and that the same user is used across containers:
   - OpenFOAM heavily uses the username and other user-specific details when building or running applications.
      - I could still set a username every time I run something, but would need some hacking.
   - OpenFOAM build the OpenFOAM adapter to a user directory (this is our default).
      - I could still move the built adapter to the system path.
   - We normally install every PIP package with `--user`.
      - We could still install them in the system, but this would be complicated.
- One workaround is to just return ownership of the files at the very end. While this covers most of the cases, it is not exception-safe. If the job gets interrupted, then this step will not be executed.
- Another workaround is to not create a user inside the container, but always mount the `/etc/passwd /etc/group` (as read-only), essentially [making Docker run as the current user](https://stackoverflow.com/a/66112924/2254346). However, this still means that whatever user-specific preparation we do in upstream containers will not be available downstream.

As a very simplistic failsafe mechanism at the moment, I could add a check in the `run-tests.sh` script that checks if you are trying to run the tests as user with different `UID`/`GID`. But maybe it is really the right time to look at the complete system and solve the issue globally.

MakisH:
An update on the files permission issue (I think I solved it):

Unfortunately, the "keep the same user name, but [change the user id](https://www.cyberciti.biz/faq/howto-change-rename-user-name-id/) just when running the case." idea could not really work, as you cannot change your own `UID:GID` while logged in.

The solution (or workaround) was in the end much simpler. I threw away all the code that was setting a specific user. We don't need a specific user for this use case. The images are now set up with the default `root` user and all the files are in `/`. This seems to work both with OpenFOAM and with the Python bindings (remember: `pip install --user`).

I now just followed [this simple approach](https://stackoverflow.com/a/71027261/2254346) to run as the current user:

```yaml
  my-service:
    image: ubuntu:latest
    user: ${MY_UID}:${MY_GID}
    volumes:
      - /etc/passwd:/etc/passwd:ro
      - /etc/group:/etc/group:ro
```

```bash
MY_UID="$(id -u)" MY_GID="$(id -g)" docker-compose up
```

Important confusion sources to avoid when debugging such cases:
- Cleanup the files completely between runs. I was for a while trying to understand why OpenFOAM could still not write on some files.
- Keep it simple. Premature optimization (and polishing) is the root of all evil.

valentin-seitz:
> ```yaml
>   my-service:
>     image: ubuntu:latest
>     user: ${MY_UID}:${MY_GID}
>     volumes:
>       - /etc/passwd:/etc/passwd:ro
>       - /etc/group:/etc/group:ro
> ```
For me this did not work as the current [openfoam-adapter container](https://github.com/precice/openfoam-adapter/blob/6b4e0cbc15bb0036a141d687d4e99aabafe0ca63/.github/workflows/Dockerfile) installs the `libpreciceAdapterFunctionObject.so` in `/root/OpenFOAM/user-v2112/platforms/linux64GccDPInt32Opt/lib` 

I needed to set `$MY_UID=0` to make it work, as openfoam otherwise cannot load the shared object.

MakisH:
> For me this did not work as the current [openfoam-adapter container](https://github.com/precice/openfoam-adapter/blob/6b4e0cbc15bb0036a141d687d4e99aabafe0ca63/.github/workflows/Dockerfile) installs the `libpreciceAdapterFunctionObject.so` in `/root/OpenFOAM/user-v2112/platforms/linux64GccDPInt32Opt/lib`
> 
> I needed to set `$MY_UID=0` to make it work, as openfoam otherwise cannot load the shared object.

I tried for a while setting up a special user, same for every repository, hoping that this would resolve this issue. However, this only made everything more complicated and did not resolve anything.

As long as the current solution works on GitHub, the problem we have is running locally as well. It would be very interesting to understand why this worked on my machine but not yours. I suspect that we are doing some step differently, which would need ~~more~~ documentation. Could you please describe what exactly you did?

In the end, maybe a solution specifically for OpenFOAM would be to create the missing user directory and move the built `libpreciceAdapterFunctionObject.so` there. This is very cheap and could be done at runtime, by the local running script.

MakisH:
Closing in favor of https://github.com/precice/tutorials/pull/347

PR #223, opened 2021-06-25
Add partitioned-heat OpenFOAM participant with solver

DavidSCN:
Since #214 is a far away from its purpose and includes various changes I don't understand let me add a clean PR. The tutorial requires groovyBC in order to handle the time dependent  inhomogenous Dirichlet boundary condition. 
Open to discuss: this PR comprises the solver as well.


MakisH:
Are the solver files part of OpenFOAM? If yes, why include them here? (guessing from the copyright notices)

DavidSCN:
No the are not, just started from OpenFOAM own files,

uekerman:
@DavidSCN Could you also please quickly comment here which results you get? (I did not run the case yet). Do you get the correct results when e.g. coupling to FEniCS?

uekerman:
> Open to discuss: this PR comprises the solver as well.

We do the same for FEniCS and Nutils. If they are not part of OpenFOAM, the right way to got is indeed to add them here in my opinion.

MakisH:
> We do the same for FEniCS and Nutils. If they are not part of OpenFOAM, the right way to got is indeed to add them here in my opinion.

Definitely, but currently the copyright notices look as if these solvers are part of OpenFOAM v2012, that's why I got confused.

uekerman:
> Definitely, but currently the copyright notices look as if these solvers are part of OpenFOAM v2012, that's why I got confused.

I see. How should / could we change the copyright notice? Just remove?

MakisH:
> > Definitely, but currently the copyright notices look as if these solvers are part of OpenFOAM v2012, that's why I got confused.
> 
> I see. How should / could we change the copyright notice? Just remove?

@DavidSCN knows best here. If the code is mostly from OpenFOAM with some small changes, then keep the headers as they are and write some "extended by" (David, preCICE, whatever policy we want to have). If the code is mostly new, then I don't see any need for detailed copyright notices, maybe only an attribution such as "based on previous work on X by Y".

DavidSCN:
I wrote some custom text in order to highlight from where we started..

DavidSCN:
This PR is now ready to go. As documented in the issue, it works as expected. I only need to update the documentation if we want to merge it. Any volunteer how wants to try out the actual tutorial (groovyBC required)?

uekerman:
> I only need to update the documentation if we want to merge it.

To which documentation are you referring?

Let's wait for the changelog PR from @IshaanDesai such that we can also document this change.

DavidSCN:
> To which documentation are you referring?

https://precice.org/tutorials-partitioned-heat-conduction.html

uekerman:
I need some documentation to build the OpenFOAM solvers :grin: 
Could you please update the documentation already?
Also run scripts are missing.

A new inconsistency issue is now the folder structure :confused: 
* fenics  
* nutils  
* openfoam-dirichlet  
* openfoam-neumann

DavidSCN:
> A new inconsistency issue is now the folder structure confused
>
>   fenics
>    nutils
>    openfoam-dirichlet
>    openfoam-neumann

Hm, not sure which part is inconsistent. Merging the OpenFOAM cases is, however, not possible as long as we explicitly want to run the OpenFOAM-OpenFOAM coupling.

MakisH:
I think that `fenics` and `nutils` are the special case here because they needed to store the solver together with the case.

We have a kind-of-similar situation in the [elastic-tube-1d](https://github.com/precice/tutorials/tree/develop/elastic-tube-1d). Previously, the Python solvers were sharing some files, which we merged into the solvers to not have this issue (e.g. hard-code parameters, instead of using configuration files, move the tools to the root directory). If we have the duplication there, we could also have it here, if it helps the structure.

An alternative we already use in the repository to avoid duplications is symbolic links.

Another alternative would be to have the solvers in a `tools/` directory in the case and then the `run.sh` script of each case would e.g. build and copy the solver binary before executing it. There are many alternatives in this direction, but this is one option.

uekerman:
Yes, we could duplicate `fenics` and `nutils` into `fenics-dirichlet`, ...

... @BenjaminRodenberg might not like this idea? (I think we had this discussion already) 

Symbolic links or `tools/` feel a bit over-engineered to me here.

The other alternative is to keep it as it is and just be careful to properly document the "how to run" in the README.

MakisH:
> Yes, we could duplicate `fenics` and `nutils` into `fenics-dirichlet`, ...
> 
> ... @BenjaminRodenberg might not like this idea? (I think we had this discussion already)

Yes, we had such a discussion already and I understand that this duplication avoidance was the original motivation.

My main counterargument was (and remains here) that special cases can lead again to a difficult to maintain situation and can make automation harder. But I completely understand that this is not a trivial case that fits nicely into our current structure.

> Symbolic links or `tools/` feel a bit over-engineered to me here.

For me too. However, I think the structure breaks because normally we assume that the solver already exists and the tutorial is only a set of config files. So, if we want to stick to the structure, we should try to separate the solver from the configs. One way is to have the solvers in a directory of a standardized name inside each tutorial case.

> The other alternative is to keep it as it is and just be careful to properly document the "how to run" in the README.

I am afraid this will set a precedent for more special cases in the future, so I would prefer to adjust the system now.

BenjaminRodenberg:
I quickly skimmed over this PR. I would suggest to merge the structure as it is `fenics, nutils, openfoam` (if possible) or `fenics, nutils, openfoam-dirichlet, openfoam-neumann` (if necessary). The discussion about whether we should restructure the `nutils` and `fenics` case does not really belong into this PR and should be moved to an issue in my opinion. If we do the restructuring in this PR, if will become gigantic. Therefore, let's try to move it into an issue / another PR as soon as possible.

> > Yes, we could duplicate `fenics` and `nutils` into `fenics-dirichlet`, ...
> > ... @BenjaminRodenberg might not like this idea? (I think we had this discussion already)
> 
> Yes, we had such a discussion already and I understand that this duplication avoidance was the original motivation.
> 
> My main counterargument was (and remains here) that special cases can lead again to a difficult to maintain situation and can make automation harder. But I completely understand that this is not a trivial case that fits nicely into our current structure.

And yes. I don't like this idea :sweat_smile: We could stick to our structure by introducing some links (e.g. `heat-dirichlet.py` and `heat-neumann.py` are just links pointing to `heat.py`). But I think this is now an idea that @MakisH does not like (and probably @uekerman also?).

BenjaminRodenberg:
> And yes. I don't like this idea :sweat_smile:

We could still draft the "duplication" idea. I can imagine that it also has some possible benefits. Right now `heat.py` has no duplication, but it is sometimes also a bit hard to read and convoluted due to the many `if`-statements. We can also create modules, where we put code that is common to both solvers and which would require a lot of duplication (e.g. https://github.com/precice/tutorials/blob/master/partitioned-heat-conduction/fenics/problem_setup.py).

DavidSCN:
Ok, then I would continue with the following structure: 
```bash
.
├── clean-tutorial.sh -> ../tools/clean-tutorial-base.sh
├── fenics
├── images
├── nutils
├── openfoam-dirichlet
├── openfoam-neumann
├── openfoam-solver
├── precice-config.xml
└── README.md
```

uekerman:
> Therefore, let's try to move it [the discussion] into an issue / another PR as soon as possible.

Please don't forget

BenjaminRodenberg:
> > Therefore, let's try to move it [the discussion] into an issue / another PR as soon as possible.
> 
> Please don't forget

Done https://github.com/precice/tutorials/issues/228

DavidSCN:
Ok, ready for review. I guess every potential reviewer is now already notified.

partitioned-heat-conduction/openfoam-solver/Make/files
```diff
@@ -0,0 +1,3 @@
+heatTransfer.C
+
+EXE = ./heatTransfer
```
MakisH, line 3: Does it make sense to build the solver into the `FOAM_USER_APPBIN`? Then we could avoid the copying.

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
```diff
@@ -0,0 +1,39 @@
...
+    writeData
+    (
+      Temperature
+    );
+  };
+};
+
+CHT
+{
+   k   [ 1  1 -3 -1 0 0 0 ] -1;
```
MakisH, line 37: Why the `-1` here?

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+funkySetFields -field T -expression '1+pow(pos().x,2)+(3*pow(pos().y,2))+1.3*time()' -time '0'
```
MakisH, line 3: Why not include this in the `run.sh`?

partitioned-heat-conduction/openfoam-neumann/constant/transportProperties
```diff
@@ -0,0 +1,11 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "constant";
+    object      transportProperties;
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
```
MakisH, line 9:
```suggestion
```

partitioned-heat-conduction/openfoam-neumann/0/T
```diff
@@ -0,0 +1,10041 @@
...
+}
+
+dimensions      [0 0 0 1 0 0 0];
+
+
+internalField   nonuniform List<scalar> 
+10000
+(
+2.0101
+2.0303
```
MakisH, line 17: I guess you wanted to have the original file before the `funkySetFields` here. In this case, it makes sense to use the `0.orig -> 0` preparation step.

partitioned-heat-conduction/openfoam-dirichlet/system/preciceDict
```diff
@@ -0,0 +1,39 @@
...
+    writeData
+    (
+      Heat-Flux
+    );
+  };
+};
+
+CHT
+{
+   k   [ 1  1 -3 -1 0 0 0 ] -1;
```
MakisH, line 37: Same here: why `-1`? A comment would help.

partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
```diff
@@ -0,0 +1,50 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      controlDict;
+}
+
+// In order to prevent misuse
```
MakisH, line 10: I understand this comment, but I am not sure a user will. Why not use a custom name and the regular `run.sh`? If we build the solver in `FOAM_USER_APPBIN`, this should work.

We would still need a comment to tell the user that this is a custom solver and where to find it.

partitioned-heat-conduction/openfoam-dirichlet/0/T
```diff
@@ -0,0 +1,10043 @@
...
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+dimensions      [0 0 0 1 0 0 0];
+
+
+internalField   nonuniform List<scalar> 
+10000
+(
+1.0001
```
MakisH, line 17: Same here as in Neumann: Is it intentional to have the initialized field here? This makes remeshing challenging.

partitioned-heat-conduction/README.md
```diff
@@ -15,50 +15,42 @@ We solve a partitioned heat equation. For information on the non-partitioned cas
...
...which needs to be compiled before running the tutorial. The solver can be compiled by executing `wmake` in the solver directory `./openfoam-solver/`. Afterwards, the custom solver `heatTransfer` can be started from the respective OpenFOAM case directory by copying it into the OpenFOAM solver directory or copying it directly into the case directory. The second challenge is given by the time- and space-dependent Dirichlet boundary conditions required for domain boundaries belonging not to the interface. For this purpose, a valid installation of `groovyBC` (part of `swak4Foam`) is required. Having the solver and an evaluator for the boundaries, the tutorial is ready to run. However, if you want to modify the tutorial the space-dependent initial condition of this tutorial state the third challenge. We let `funkySetFields`, which is installed along with OpenFOAM by default, evaluate the initial condition. You can simply execute the script `setInitialField.sh` in order to evaluate and store the required initial condition in the `0` directory. Note that `funkySetFields` overrides unintentionally the boundary conditions as well, so that you might need to restore them after you executed it.
```
MakisH, line 30: I think that this is a very long paragraph with multiple different points. Better move it to an independent section and refer there.

MakisH (commented):
A quick first review. Please trigger me again afterwards to try it.

partitioned-heat-conduction/openfoam-solver/Make/files
DavidSCN: I personally don't like these directories that much, but I changed it now.

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
DavidSCN: Compatibility. `fenics` and `nutils` have different sign conventions here.

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
DavidSCN: It is essentially described in the `README`: the problem with `funkySetFields` is that it overrides the boundary conditions (in the wrong way). Thus, this function is only meant for people who want to modify the mesh.

partitioned-heat-conduction/openfoam-neumann/0/T
DavidSCN: Not really. As explained above, the `funkySetFields` is not part of the default tutorial. These are the original files.

partitioned-heat-conduction/openfoam-dirichlet/0/T
DavidSCN: Indeed, but an explanation is given in the `README`. The default cases should be as easy as possible, so that I would go with this solution.

partitioned-heat-conduction/README.md
DavidSCN: I'm not sure: This section was originally dedicated to case specific details and all these things are only related to this specific tutorial cases and openfoam. Why moving it to another section and where should the section go? For me, this makes things more complicated.

partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
DavidSCN: I added now a comment and the solver is build into the `FOAM_USE_APPBIN`, works.

partitioned-heat-conduction/openfoam-dirichlet/system/preciceDict
DavidSCN: I added a comment.

DavidSCN:
> A quick first review. Please trigger me again afterwards to try it.

Done and ready to try.

partitioned-heat-conduction/openfoam-solver/Make/files
MakisH: But why not? Please don't incorporate any suggestions you are not convinced of. :)

partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
```diff
@@ -0,0 +1,52 @@
...
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      controlDict;
+}
+
+// Make sure you build the solver located
+// in ./openfoam-solver before running
+// the case in the
```
MakisH, line 12: "in the"?

partitioned-heat-conduction/openfoam-neumann/system/controlDict
```diff
@@ -0,0 +1,52 @@
...
+    location    "system";
+    object      controlDict;
+}
+
+// Make sure you build the solver located
+// in ./openfoam-solver before running
+// the case in the
+application     heatTransfer;
+
+libs ( "libgroovyBC.so" ) ;
```
MakisH: It is currently not documented that we need groovyBC. Maybe add a comment in the `README.md`?

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
MakisH: Are we then restricted to which solver combinations we can run? Can the user still select between `openfoam-openfoam`, `openfoam-nutils`, `fenics-openfoam`? I understand that now only the `openfoam-<other>` and `<other>-openfoam` are valid cases now. Maybe we could find a more fundamental solution to this.

If I am looking at the right places, this is not documented.

partitioned-heat-conduction/openfoam-neumann/system/controlDict
DavidSCN: It is already documented in the `README`. Where else should I document it?

MakisH: I am wrong, it is documented.

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
MakisH: What do you mean by "in the wrong way"? Should we document this?

DavidSCN:
> 

Are we then restricted to which solver combinations we can run? Can the user still select between openfoam-openfoam, openfoam-nutils, fenics-openfoam? I understand that now only the openfoam-<other> and <other>-openfoam are valid cases now. Maybe we could find a more fundamental solution to this.

If I am looking at the right places, this is not documented.


All combinations are valid as the sign cancels out for `openfoam-openfoam`

partitioned-heat-conduction/openfoam-neumann/0/T
MakisH: But these files are generated automatically. Could we make this automatic preparation part of the tutorial?

I am just afraid of mesh-specific files since #6. Right now, I could not easily make a coarser or finer mesh.

partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
DavidSCN: Damn, my vi skills went wrong. Will fix it.

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
MakisH: (let's keep the discussion in the `openfoam-dirichlet` file, keeping open as a reminder)

partitioned-heat-conduction/openfoam-neumann/0/T
MakisH: (let's keep the discussion in the `openfoam-dirichlet` file, keeping open as a reminder)

partitioned-heat-conduction/openfoam-dirichlet/0/T
MakisH: (let's keep the discussion in the `openfoam-dirichlet` file, keeping open as a reminder)

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
DavidSCN: Ah GitHub duplicated your comment, see the answer below.

MakisH (commented):
A couple more things

partitioned-heat-conduction/openfoam-solver/Make/files
DavidSCN: The directory is somewhat arbitrary in your home directory

partitioned-heat-conduction/openfoam-dirichlet/0/T
DavidSCN: This is the `openfoam-dirichlet` file?! Feel free to discuss

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
DavidSCN: > What do you mean by "in the wrong way"? Should we document this?

It deletes the `groovyBC` and adds `zeroGradient` conditions. I already added a small comment regarding this in the `README`.

partitioned-heat-conduction/README.md
MakisH: My observation is that this part is difficult to parse quickly, it needs to explain far more content than the FEniCS/Nutils, and I cannot add line-specific comments and suggestions for the review.

What I have in mind is something like:

```markdown
* OpenFOAM. Install [OpenFOAM](<our docs>) and the [OpenFOAM adapter](<our docs>). This tutorial uses a custom solver, which you can find in <location> and build using `cd <location> && wmake`. 

### Notes on the OpenFOAM case

(more details)
```

Alternatives:
- Use nested bullets for more details (avoid the subsection).
- Move the running parts to the "running the simulation" section.

I just see a large difference between the OpenFOAM case documentation and the rest and I am trying to make it specific. Can you also see this difference?

partitioned-heat-conduction/openfoam-neumann/0/T
DavidSCN: Hm I agree that it is not trivial (though it is quiet easy if you know how to). I could certainly hack something together with some efforts, but I'm not sure if this pays off and is the actual goal of this particular tutorial case. For us, it is mainly for testing purpose, right?

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
MakisH: > Ah GitHub duplicated your comment, see the answer below.

I am not sure if we found a GitHub bug or if I commented in the wrong places, but something confusing is going on here.

partitioned-heat-conduction/openfoam-neumann/constant/transportProperties
MakisH: Oh, we should also remove it from there, then. :-/

partitioned-heat-conduction/openfoam-neumann/0/T
MakisH: Well, we should at least document the "how to", otherwise this test case would one day become abandoned.

partitioned-heat-conduction/openfoam-solver/Make/files
MakisH: But isn't it the expected way for the user to build OpenFOAM applications available across their installation? We also build the adapter in the `FOAM_USER_LIBBIN`, although there it really makes sense, as we are using it in more places. Here, it is arguable, but it avoids the duplication/copying.

partitioned-heat-conduction/README.md
DavidSCN: > My observation is that this part is difficult to parse quickly, it needs to explain far more content than the FEniCS/Nutils, 

Opening up a dedicated section will not change it, right?

> and I cannot add line-specific comments and suggestions for the review.

That's not a problem of the documentation on t he website in the first place.

> I just see a large difference between the OpenFOAM case documentation and the rest and I am trying to make it specific. Can you also see this difference?

There are differences, but my assumption is that the people only read the parts they are interested in. I will try to break it up into smaller subsections.

partitioned-heat-conduction/openfoam-neumann/system/preciceDict
DavidSCN: I think you did not comment here but rather sent it with a new review. Too much confusion for GitHub. Anyway, let's consider the discussion.

uekerman: > Compatibility. fenics and nutils have different sign conventions here.

Sounds like sth we need to resolve. All combinations should be possible without changing any of the participants.

DavidSCN: As documented below, all combinations work with the current setup.

partitioned-heat-conduction/README.md
DavidSCN: I changed it now. Feel free to review and re-open again if not sufficient.

partitioned-heat-conduction/openfoam-neumann/0/T
DavidSCN: I scripted now everything and it should work (including the mesh modifications),

partitioned-heat-conduction/openfoam-neumann/setInitialField.sh
DavidSCN: Done

DavidSCN:
I addressed now everything and it should work very easy now, even the mesh modifications. Feel free to review again.

partitioned-heat-conduction/README.md
```diff
@@ -27,7 +27,17 @@ You can either couple a solver with itself or different solvers with each other.
...
+
+### Notes on the OpenFOAM case
+
+Running this tutorial with OpenFOAM is a bit of a challenge and requires some special considerations:
+
+* First of all, OpenFOAM does not provide a Laplace solver with a non-zero right-hand side. Therefore, we provide a modified Laplace solver together with the tutorial, which needs to be compiled before running the tutorial. The solver can be compiled by executing `wmake` in the solver directory `./openfoam-solver/`. The generated executable will be stored in the `FOAM_USER_APPBIN` by default. Afterwards, the custom solver `heatTransfer` can be started from the respective OpenFOAM case directory, as usual.
+
+* The second challenge is given by the time- and space-dependent Dirichlet boundary conditions required for domain boundaries belonging not to the interface. For this purpose, a valid installation of `groovyBC` (part of `swak4Foam`) is required.
```
MakisH, line 38:
```suggestion
* The second challenge is given by the time- and space-dependent Dirichlet boundary conditions required for domain boundaries not belonging to the interface. For this purpose, a valid installation of `groovyBC` (part of `swak4Foam`) is required.
```

partitioned-heat-conduction/README.md
```diff
@@ -27,7 +27,17 @@ You can either couple a solver with itself or different solvers with each other.
...
+
+* The third challenge is given by the space-dependent initial conditions. We let `funkySetFields`, which is installed along with OpenFOAM by default, evaluate the initial condition. You can simply execute the script `./run.sh` script, as usual, which calls the `setInitialField.sh` in order to evaluate and store the required initial condition in the `0` directory. Note that the `./run.sh` script deletes the `0` time directory and copies it again from `0.orig`. If you start modifying the initial or boundary conditions, make sure you modify the files located in the `0.orig` directory in combination with the default `./run.sh` scripts.
```
MakisH, line 40: Trying to make it a bit more concise (and fixing 1-2 typos):
```suggestion
* The third challenge is given by the space-dependent initial conditions. We use `funkySetFields` (installed with OpenFOAM) to evaluate the initial condition. You can directly execute the `./run.sh` script, which calls the `setInitialField.sh` in order to evaluate the required initial condition and store it in the `0` directory. Note that `run.sh`  deletes the `0` time directory and copies it again from `0.orig`. If you start modifying the initial or boundary conditions, make sure you modify the files located in the `0.orig` directory in combination with the default `run.sh` scripts.
```

partitioned-heat-conduction/openfoam-neumann/0.orig/T
```diff
@@ -0,0 +1,47 @@
+/*--------------------------------*- C++ -*----------------------------------*\
+| =========                 |                                                 |
+| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
+|  \\    /   O peration     | Version:  2012                                  |
+|   \\  /    A nd           | Website:  www.openfoam.com                      |
+|    \\/     M anipulation  |                                                 |
+\*---------------------------------------------------------------------------*/
```
MakisH, line 7: Do we need these headers? I understand that this file is pretty much own work.

(could also delete the comment rulers below, as discussed previously)

MakisH (commented):
I have not tried to run it yet, but a few more comments on the updates.

partitioned-heat-conduction/openfoam-neumann/0.orig/T
DavidSCN: No let me remove them (generated by `funkySetFields`, I will just keep the rulers for consistency for now.

DavidSCN:
@MakisH Do you want to try running this tutorial before merging? I only need to add a change log before merging.

partitioned-heat-conduction/openfoam-neumann/clean.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e -u
+
+. ../../tools/cleaning-tools.sh
+
+clean_openfoam .
```
MakisH, line 6: This tutorial is special in the sense that it first copies the `0.orig` to `0`. Let's clean this `0/` here (and in the openfoam-dirichlet):

```suggestion
clean_openfoam .
rm -rfv 0
```

Similarly, you can add a `clean.sh` in `openfoam-solver` that executes `wclean`. Don't forget to mark it as executable.

I noticed this as I am usually running `git status` after running and after cleaning and check for residue files.

MakisH (approved):
I also ran it now and it works as expected. If you want, consider these two additional points and feel free to merge.

partitioned-heat-conduction/openfoam-neumann/clean.sh
DavidSCN: Done, but the `wclean` will not remove the solver executable (FYI)

DavidSCN:
Done, thanks @MakisH

partitioned-heat-conduction/openfoam-neumann/clean.sh
MakisH: At least it removes the intermediate files from the `openfoam-solver/Make/<platform>` directory.

PR #224, opened 2021-06-29
Bump versions in Quickstart

MakisH:
This bumps the following in our Quickstart page:

- preCICE v2.2.0 --> v2.2.1 This is still hard-coded, as it is not part of the website repository. It anyway also links to related resources, but it is arguable. Our policy is still unclear here, but if I follow the same principle as we use for hyperlinks, the tutorials repository should be readable without the website. @uekerman what do you think?
- OpenFOAM v2012 --> v2106.


uekerman (approved):
Hard-coding here is tolerable in my opinion. If we forget it at some point, everything is still correct and will work (it's just a bit too old), and also users will rather easily understand how to adapt the code to get the newest preCICE.

PR #227, opened 2021-07-14
Port elastic-tube-3d

IshaanDesai:
This PR ports the files of the `elastic-tube-3d` case to the new tutorials format.


DavidSCN (approved):
Looks good. The failing test must be unrelated as no link was touched. However, we should fix so that everything passes. Maybe the test run itself was erroneous. Let me trigger it again.

IshaanDesai:
I will merge this once the change log is in `develop`

Issue #226, opened 2021-07-14
Required Problem Parameters for flow-over-heated-plate


HishamSaeed:
Hi i am trying to couple openFOAM with Elmer FEM solver, i am using the flow over heated plate problem as a test problem. For this tutorial, i cannot find the density of the fluid as well as the solid which are required by Elmer solver, otherwise the solver stops and does not continue. I cannot calculate the fluid density from Prandtl number because the density cancels out in equation 
```$$Pr = \nu / \alpha $$ , $$Pr = (\mu / \rho) / (\cp * k / \rho) $$``` .
For more details about Prandtl number equation: https://en.wikipedia.org/wiki/Prandtl_number

Also in the solid i cannot calculate it because `alpha(thermal diffusivity)` is provided as well as `k(Heat Conductivity`) but the `cp(Heat Capacity)` is not provided which also is required by Elmer. Is there a way to access these data or calculate it.

Thanks in advance
Hisham

DavidSCN:
According to OpenFOAM, the fluid density has initially a uniform value of `1 kg/m^3` (and of course varies during the course of the simulation). AFAIU you are running the solid participant, right? Does a density value of `1 kg/m^3` work for you?

BenjaminRodenberg:
> According to OpenFOAM, the fluid density has initially a uniform value of `1 kg/m^3` (and of course varies during the course of the simulation).

Then this information should also go into our documentation. I will open a PR on this.

BenjaminRodenberg:
> Does a density value of `1 kg/m^3` work for you?

From my perspective the density should not be required for solving the heat equation in the solid, if you are only interested in heat flux and temperature. In this case [thermal diffusivity](https://en.wikipedia.org/wiki/Thermal_diffusivity) and [thermal conductivity](https://en.wikipedia.org/wiki/Thermal_conductivity) should be sufficient. So our current specification for the solid is complete from this perspective.

However, as soon as you are interested in the actual energy you either need the [specific heat capacity](https://en.wikipedia.org/wiki/Specific_heat_capacity) or the density. Additionally, we do not know what different solvers require. Even if the [FEniCS-based heat solver](https://github.com/precice/tutorials/blob/1672e89181e66346127ba40cd15cb513d3fafb39/flow-over-heated-plate/solid-fenics/solid.py) works without density, other solvers may compute the energy (even if we don't need it) and therefore the density will be required.

[...]

PR #233, opened 2021-08-04
flow-over-heated-plate: Add density and pressure to problem parameters

BenjaminRodenberg:
Related to #226. Currently the density is missing. This is no problem for OpenFOAM, where density defaults to 1kg/m^3 (see https://github.com/precice/tutorials/issues/226#issuecomment-888344050), but for other solvers (e.g. ELMER), where the density must be specified by the user.

# Open questions

* Do we also need density for the solid solver? Let's try to solver this in #226.

### DIFF
--- a/flow-over-heated-plate/README.md
+++ b/flow-over-heated-plate/README.md
@@ -15,7 +15,7 @@ This scenario consists of one fluid and one solid participant and it is inspired
 
 The test case is two-dimensional and a serial-implicit coupling with Aitken underrelaxation is used for the coupling.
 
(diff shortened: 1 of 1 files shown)


DavidSCN:
I was wondering why OpenFOAM does not need the parameter and ELMER does. Before merging here, it would be interesting to see whether a coupling with ELMER using the density parameter delivers the same results as other solver combinations do.

BenjaminRodenberg:
From my current understanding (see https://github.com/precice/tutorials/issues/226#issuecomment-892549613), I would propose to provide a density for solid and fluid. For the solid the density is technically a "don't care", but 

* this is something we can and should verify (not sure if the Elmer adapter is already far enough for this kind of verification) and
* I would still define a density (even if it is arbitrary), because this will avoid discussions in the future and make it easier for everybody who does not want to dig deep into the technical details.

For this PR this means:

* Leave it as it is
* Do some experiments with densities for more security

DavidSCN:
> From my current understanding (see [#226 (comment)](https://github.com/precice/tutorials/issues/226#issuecomment-892549613)), I would propose to provide a density for solid and fluid. For the solid the density is technically a "don't care", but
> 

Agree

> 
> For this PR this means:
> 
>     * Leave it as it is
> 
>     * Do some experiments with densities for more security

I would vote for the second option because once we merge here nobody cares any more and we might painfully realize an issue later on.

BenjaminRodenberg:
I just added the ambient pressure as an additional parameter. From my perspective this closes the system and allows us to derive the molar mass (if the solver requires it). See https://github.com/precice/tutorials/issues/226#issuecomment-897376368 for the derivation.

I think this PR is ready to merge. We could still perform some experiments, but since the derivation is quite simple I don't think that this is necessary anymore. @DavidSCN could you review and merge/comment?

DavidSCN (approved):

PR #234, opened 2021-08-11
flow-over-heated-plate: Remove unnecessary parameter

BenjaminRodenberg:
See https://github.com/precice/tutorials/issues/226#issuecomment-896933844: From my current perspective, we do not need `nMoles`.

### DIFF
--- a/changelog-entries/234.md
+++ b/changelog-entries/234.md
@@ -1,0 +1,1 @@
+* Remove unnecessary parameter `nMoles` from flow-over-heated-plate cases with OpenFOAM #234
--- a/flow-over-heated-plate-nearest-projection/fluid-openfoam/constant/thermophysicalProperties
+++ b/flow-over-heated-plate-nearest-projection/fluid-openfoam/constant/thermophysicalProperties
@@ -23,7 +23,6 @@ mixture
 {
     specie
     {
-        nMoles          1;
         molWeight       24.0999;
     }
     thermodynamics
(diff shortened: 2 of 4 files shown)


BenjaminRodenberg:
I also removed `nMoles` in the cases `flow-over-heated-plate-steady-state` and `flow-over-heated-plate-nearest-projection`, but did not check any results.

MakisH:
Have you also tried this with different OpenFOAM versions? I can imagine that this was there for technical reasons, even if not used. But most probably this originates from the very first prototypes, which were probably also based on something else (as is common in OpenFOAM).

BenjaminRodenberg:
> Have you also tried this with different OpenFOAM versions?

Of course, not :smirk:

uekerman:
@BenjaminRodenberg Don't forget the changelog entry.

MakisH (approved):
Indeed. From the [OpenFOAM 8 docs](https://cfd.direct/openfoam/user-guide/v8-thermophysical/):

> nMoles: number of moles of component. This entry is only used for combustion modelling based on regress variable with a homogeneous mixture of reactants; otherwise it is set to 1.

MakisH (approved):
This looks correct. From the OpenFOAM.org user guide:

> nMoles: number of moles of component. This entry is only used for combustion modelling based on regress variable with a homogeneous mixture of reactants; otherwise it is set to 1.

PR #238, opened 2021-09-22
Remove 'www' from preCICE website links

DavidSCN:
```bash
perl -pi -e 's/https:\/\/www.precice.org/https:\/\/precice.org/g;' `find *.md */*md -type f`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # preCICE tutorials
 
-This repository contains ready-to-run tutorial cases for the coupling library [preCICE](http://www.precice.org/).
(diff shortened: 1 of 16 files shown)


MakisH (approved):
Looks good with a quick look at the diff. :+1:

The action we are using does not seem to have any option to check SSL certificates, but maybe this is then really beyond scope.

Edit: Maybe also check for links in comments in scripts and configuration files.

DavidSCN:
There was indeed a link in a run script i fixed it. I will also update the website import afterwards.

PR #244, opened 2021-11-29
ASTE Turbine Tutorial

kursatyurt:
ASTE was lack of a tutorial in tutorials.

This PR suggests a tutorial that shows the usage of ASTE  tools: mesh partitioner, mesh joiner, calculator, beside ASTE itself.

The tutorial uses the turbine meshes from the preCICE repository. This way it provides a variety of meshes for different trials.


aste-turbine/README.md
```diff
@@ -0,0 +1,32 @@
...
+permalink: tutorials-aste-turbine.html
+keywords: ASTE, Testing, Turbine
+summary: This tutorial is an example case for ASTE, showcasing basic features and usage of ASTE.
+---
+
+{% include note.html content="Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/aste-turbine). Read how in the [tutorials introduction](https://precice.org/tutorials.html)." %}
+
+## Setup
+
+This case is a simple usage scenario for. Some features offered by this case:
```
MakisH, line 12: oops: for what? :sweat_smile: 

I guess you meant to link to ASTE and some documentation here.

Also, maybe write "features demonstrated" instead of "features offered".

aste-turbine/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Running the ASTE
+
+Run the `run.sh` script. It do the followings:
+
+* Downloads the meshes from preCICE repository.
+* Calculates function `x + y` on a finer grid.
+* Partitions the coarse and fine mesh into 2 domains.
+* Map the data from fine grid to coarse grid.
+* Join the result meshes into a final mesh.
+* Using calculator calculates difference between mapped and original function on coarse grid.
```
MakisH, line 30: Maybe add a few more details on each step?

MakisH, line 30: Notice how I change the beginning of each bullet point to follow the same style:
```suggestion
* Maps the data from fine grid to coarse grid.
* Joins the result meshes into a final mesh.
* Calculates difference between mapped and original function on coarse grid.
```

aste-turbine/precice.xml
```diff
@@ -0,0 +1,55 @@
...
+
+  <log enabled="true">
+    <sink type="stream" output="stdout" filter="%Severity% > debug" enabled="true" />
+    <sink type="file" output="filtered.log"  filter= "(%Severity% > debug) or (%Severity% >= trace and %Module% contains SolverInterfaceImpl)" enabled="true" />
+    <sink type="file" output="debug.log" filter="" enabled="true"/>
+  </log>
+  
+  <solver-interface dimensions="3">
+    
+    <!-- Data fields that are exchanged between the solvers -->
```
MakisH, line 13: In the rest of the tutorials, we decided to not include such explanation comments in the XML files. Please remove them also from here (I know these are old anyway).

The different mapping methods that we are expected to try can remain.

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
```
MakisH, line 4: Even more compact:

```suggestion
#!/usr/bin/env bash
set -e -x
```

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# This script assumes the ASTE binaries and python scripts are in $PATH
+
+# Download the meshes
+test -f meshes.tar.gz  || wget https://gitlab.lrz.de/precice/precice2-ref-paper-setup/-/raw/main/meshes/meshes.tar.gz
+
+# Extract the meshes
+tar -xvf meshes.tar.gz
```
MakisH, line 12: You can even remove the verbosity here, it makes output a bit more difficult to parse:
```suggestion
tar -xf meshes.tar.gz
```

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+# This script assumes the ASTE binaries and python scripts are in $PATH
+
+# Download the meshes
+test -f meshes.tar.gz  || wget https://gitlab.lrz.de/precice/precice2-ref-paper-setup/-/raw/main/meshes/meshes.tar.gz
```
MakisH, line 9: Nice that you are testing if the file exists before downloading it again! Does it make sense to also check for the target directory and skip the extraction step as well?

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Extract the meshes
+tar -xvf meshes.tar.gz
+
+# Calculate on fine mesh
+./vtk_calculator.py 0.009.vtk x+y -t "x + y"
```
MakisH, line 15: Maybe comment on these seemingly arbitrary numbers?
Maybe rename the meshes to something like `meshA.vtk` and `meshB.vtk` to make the script more understandable?

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
...
+mpirun -n 2 preciceMap -v -p A --mesh fine_mesh/fine_mesh --data "x + y" &
+mpirun -n 2 preciceMap -v -p B --mesh coarse_mesh/coarse_mesh --output mapped/mapped --data "x + y"
+
+# Join the output files together to result.vtk,
+# Recovery cannot be used since GlobalID's are not exist in mapped mesh
+./join_mesh.py mapped/mapped -o result.vtk
+
+# Measure the difference between the original function and the mapped values
+# Save into data array called difference
+./vtk_calculator.py result.vtk x+y -t difference -it "x + y" --diff
```
MakisH, line 34: GitHub complains for a missing line at the end, let's add it:

```suggestion
./vtk_calculator.py result.vtk x+y -t difference -it "x + y" --diff

```

MakisH (changes requested):
Thanks for the contribution, this is something we definitely need and many future students will thank you for!

Here are a few small comments, mostly on the documentation and consistency side of things.

aste-turbine/README.md
```diff
@@ -0,0 +1,32 @@
+# ---
+title: ASTE (Artificial Solver Testing Environment) Turbine Tutorial
+permalink: tutorials-aste-turbine.html
```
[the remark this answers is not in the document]
kursatyurt: Thank you for notice, here : https://github.com/precice/precice.github.io/pull/104

aste-turbine/README.md
kursatyurt: Small detail makes huge difference :+1:

aste-turbine/run.sh
kursatyurt: Now it extracts to the main directory checking each file might be cumbersome but we can change it to extracting to a directory and check it

aste-turbine/run.sh
kursatyurt: The turbine meshes consist of many different resolutions, even the run script just only uses two of them.

If we rename them as `fine` `coarse` or `meshA` `meshB` the resolutions might get confusing. 

Might be a better solution renaming all them as `turbine_l1` `turbine_l2` ... `turbine_l#` for different levels of resolution?

aste-turbine/precice.xml
kursatyurt: The comments are removed just different mapping techniques are kept as reference

aste-turbine/README.md
kursatyurt: I added the scripts which are used for each step.

aste-turbine/README.md
```diff
@@ -0,0 +1,32 @@
...
+Run the `run.sh` script. It performs the following steps:
+
+* Downloads the meshes from preCICE repository.
+* Using `vtk_calculator.py` script, calculates function `x + y` on a finer grid.
+* Using `partition_mesh.py` script, partitions the coarse and fine mesh into 2 domains.
+* Using `preciceMap` executable, maps the data from the fine grid to the coarse grid.
+* Using `join_mesh.py`script, joins the result meshes into a final mesh.
+* Using ``vtk_calculator.py` script, calculates difference between mapped and original function on coarse grid.
+
+Important note:  The `run.sh` script assumes that the ASTE binaries and python scripts are in `$PATH`.
```
DavidSCN, line 32: Does it work for you when you use (from `PATH` in a different directory) `partition_mesh.py --algorithm topology` which relies on `libmetis`. Also, it would be good to have an install target in the `CMakeLists.txt` of `ASTE` so that user can easily install it without prepending system paths.

kursatyurt: You are right it is not working with `libmetis`, since in python script look `metisAPI.so` in `cwd`. 

Now it is fixed in https://github.com/precice/aste/pull/53. 

I opened an issue about installation
https://github.com/precice/aste/issues/54

aste-turbine/README.md
```diff
@@ -0,0 +1,32 @@
...
+* Usage of ASTE Calculator for calculating a function on given mesh.
+* Usage of ASTE.
+
+
+## Running the ASTE
+
+Run the `run.sh` script. It performs the following steps:
+
+* Downloads the meshes from preCICE repository.
+* Using `vtk_calculator.py` script, calculates function `x + y` on a finer grid.
```
uekerman, line 26: Let's try to use a more sophisticated test function here. There are some standard functions which are often used to test RBF interpolation, e.g the [eggholder function](https://www.sfu.ca/~ssurjano/egg.html).

uekerman:
What is our strategy to explain users how to reproduce their setup with ASTE? Do we want to do this in this tutorial as well, in another tutorial, or simply "only" in the documentation of ASTE?

DavidSCN:
> What is our strategy to explain users how to reproduce their setup with ASTE? Do we want to do this in this tutorial as well, in another tutorial, or simply "only" in the documentation of ASTE? 

Not completely sure at the moment, but I guess it fits best into the replay mode. We will have another tutorial for this anyway.

kursatyurt:
> What is our strategy to explain to users how to reproduce their setup with ASTE? Do we want to do this in this tutorial as well, in another tutorial, or simply "only" in the documentation of ASTE?

I think this tutorial just shows simply how users evaluate mapping of preCICE on their meshes, like the section on the reference paper. There would be another tutorial for reproducing their setup in ASTE (vector replay mode?).

aste-turbine/README.md
kursatyurt: Thank you, I will definitely consider using this.

uekerman:
Just to avoid confusion, "replay mode" is yet another use case. I see the following three use case of ASTE:

* To evaluate/study mapping methods on "artificial" geometries/meshes. This is what the tutorial here does.
* Reproducing a specific mapping setup of a coupled case. For example: case crashes since mapping fails. User exports both meshes and values and imports in ASTE. Tries to reproduce in ASTE. Then, we could debug for them or they could test different basis functions etc.
* Replay mode. Replace a participant in a coupled setup with ASTE. One then always gets a uni-directional coupling. This is useful for debugging, for developing new adapters, but also for efficiency reasons (explicit instead of implicit coupling etc).

DavidSCN:
@uekerman considering the use-cases: would it make sense to create an `aste` directory on the top level and the three use-cases as subdirectories?

@kursatyurt You can ignore the `check-scripts` CI (please run shellcheck locally). However, the markdown file still has some issues.

aste-turbine/run.sh
MakisH: You could also define the fine and coarse meshes with variables in the script. This way, the intention becomes clearer, and the numbers are easily configurable.

But I don't yet fully understand the setup, I need to run it once.

aste-turbine/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This case is a simple usage scenario for ASTE. Some features demonstrated by this tutorial:
+
+* Mapping from a fine grid (`0.009.vtk`) to a coarse grid (`0.01.vtk`).
+* Usage of ASTE partitioner.
+* Usage of ASTE mesh joiner.
+* Usage of ASTE Calculator for calculating a function on given mesh.
+* Usage of ASTE.
+
+## Running the ASTE
```
MakisH, line 20:
```suggestion
## Running ASTE
```

aste-turbine/README.md
```diff
@@ -0,0 +1,31 @@
...
+{% include note.html content="Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/aste-turbine). Read how in the [tutorials introduction](https://precice.org/tutorials.html)." %}
+
+## Setup
+
+This case is a simple usage scenario for ASTE. Some features demonstrated by this tutorial:
+
+* Mapping from a fine grid (`0.009.vtk`) to a coarse grid (`0.01.vtk`).
+* Usage of ASTE partitioner.
+* Usage of ASTE mesh joiner.
+* Usage of ASTE Calculator for calculating a function on given mesh.
```
MakisH, line 17:
```suggestion
* Usage of ASTE calculator for calculating a function on a given mesh.
```

aste-turbine/README.md
```diff
@@ -0,0 +1,31 @@
...
+Run the `run.sh` script. It performs the following steps:
+
+* Downloads the meshes from preCICE repository.
+* Using `vtk_calculator.py` script, calculates function `x^2+exp(y^3)+cos(z)` on a finer grid.
+* Using `partition_mesh.py` script, partitions the coarse and fine mesh into 2 domains.
+* Using `preciceMap` executable, maps the data from the fine grid to the coarse grid.
+* Using `join_mesh.py`script, joins the result meshes into a final mesh.
+* Using ``vtk_calculator.py` script, calculates difference between mapped and original function on coarse grid.
+
+Important note:  The `run.sh` script assumes that the ASTE binaries and python scripts are in `$PATH` or ASTE is installed.
```
MakisH, line 31: Maybe give a quick check for this and link to the new ASTE documentation page. This is currently not mentioned anywhere here.

aste-turbine/run.sh
```diff
@@ -0,0 +1,34 @@
...
+# This script assumes the ASTE binaries and python scripts are in $PATH or ASTE installed on your system
+
+# Download the meshes
+test -f meshes.tar.gz  || wget https://gitlab.lrz.de/precice/precice2-ref-paper-setup/-/raw/main/meshes/meshes.tar.gz
+
+# Extract the meshes
+tar -xf meshes.tar.gz
+
+# Calculate on fine mesh
+vtk_calculator.py -m 0.009.vtk -f "x^2+exp(y^3)+cos(z)" -d "x^2+exp(y^3)+cos(z)"
```
MakisH, line 13: Are these the same options documented in `tooling-aste`? From that page:

```
| -l   | Logging level (default="INFO")                                                      |
| -d   | Calculate the difference to present data.                                           |
```

tools/clean-tutorial-base.sh
```diff
@@ -4,4 +4,9 @@ set -e -u
 # shellcheck disable=SC1091
 . ../tools/cleaning-tools.sh
 
+rm -fv result.vtk 
+rm -fvr fine_mesh coarse_mesh mapped
 clean_tutorial .
+clean_precice_logs .
+rm -fv *.log
```
MakisH, line 11: Would it be reasonable to restrict this to more specific filenames / patterns of logfiles?

MakisH (commented):
Thanks again for improving everything! Here are some details.

aste-turbine/precice.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0"?>
```
uekerman, line 1: We should call the file `precice-config.xml` to be consistent with other tutorials.

aste-turbine/precice.xml
```diff
@@ -0,0 +1,53 @@
...
+
+<precice-configuration>
+
+  <log enabled="true">
+    <sink type="stream" output="stdout" filter="%Severity% > debug" enabled="true" />
+    <sink type="file" output="filtered.log"  filter= "(%Severity% > debug) or (%Severity% >= trace and %Module% contains SolverInterfaceImpl)" enabled="true" />
+    <sink type="file" output="debug.log" filter="" enabled="true"/>
+  </log>
```
uekerman, line 9:
```suggestion
  <log>
    <sink
      filter="%Severity% > debug and %Rank% = 0"
      format="---[precice] %ColorizedSeverity% %Message%"
      enabled="true" />
  </log>
```
To be consistent with other preCICE tutorials.

aste-turbine/precice.xml
```diff
@@ -0,0 +1,53 @@
...
+    
+    <data:scalar name="Data" />
+
+    <mesh name="MeshA" flip-normals="no">
+      <use-data name="Data" />
+    </mesh>
+
+    <mesh name="MeshB" flip-normals="no">
+      <use-data name="Data" />
+    </mesh>
```
uekerman, line 21:
```suggestion
    <mesh name="A-Mesh" >
      <use-data name="Data" />
    </mesh>

    <mesh name="B-Mesh">
      <use-data name="Data" />
    </mesh>
```
Let's not specify default arguments.
Changed the mesh names to follow the [naming conventions](https://precice.org/community-contribute-to-precice.html#contributing-tutorials).

aste-turbine/precice.xml
```diff
@@ -0,0 +1,53 @@
...
+    
+    <m2n:sockets from="A" to="B" network="lo" exchange-directory="." />
```
uekerman, line 23:
```suggestion
    <m2n:sockets from="A" to="B" exchange-directory="." />
```

aste-turbine/precice.xml
```diff
@@ -0,0 +1,53 @@
...
+      <use-mesh name="MeshB" provide="yes" />
+      <read-data name="Data" mesh="MeshB" />
+
+      <mapping:nearest-neighbor constraint="consistent" direction="read" from="MeshA" to="MeshB" />
+      <!-- <mapping:nearest-projection constraint="consistent" direction="read" from="MeshA" to="MeshB" /> -->
+      <!-- <mapping:rbf-gaussian shape-parameter="4" constraint="consistent" direction="read" from="MeshA" to="MeshB" x-dead="false" y-dead="false" z-dead="false" /> -->
+      <!-- <mapping:rbf-gaussian shape-parameter="5600" solver-rtol="1e-9" constraint="consistent" direction="read" from="MeshA" to="MeshB" x-dead="true" y-dead="false" z-dead="false" polynomial="separate" preallocation="tree"/> -->
+      <!-- <mapping:rbf-volume-splines solver-rtol="1e-9" constraint="consistent" direction="read" from="MeshA" to="MeshB" x-dead="false" y-dead="false" z-dead="false" /> -->
+      
+      <!-- <export:vtk every-n-time-windows="1" directory="vtkB/" normals="0"/> -->
```
uekerman, line 41:
```suggestion
      <!-- <mapping:nearest-projection constraint="consistent" direction="read" from="MeshA" to="MeshB" /> -->
      <!-- <mapping:rbf-gaussian shape-parameter="4" constraint="consistent" direction="read" from="MeshA" to="MeshB" /> -->
      <!-- <mapping:rbf-gaussian shape-parameter="5600" solver-rtol="1e-9" constraint="consistent" direction="read" from="MeshA" to="MeshB" x-dead="true" polynomial="separate" preallocation="tree"/> -->
      <!-- <mapping:rbf-volume-splines solver-rtol="1e-9" constraint="consistent" direction="read" from="MeshA" to="MeshB" /> -->
      
      <!-- <export:vtk directory="vtkB/" /> -->
```
Again removing default options. Why `x-dead="true"` here?

uekerman (commented):
Some brief suggestions on the preCICE config file.

uekerman:
> @uekerman considering the use-cases: would it make sense to create an `aste` directory on the top level and the three use-cases as subdirectories?

I am not yet sure about this, but I guess this is also a future problem. Let's first finish this one.

For the "test your mapping" use case, I am not sure whether we need a testcase (tutorial) at all. Maybe a step-by-step description in the docs would be better for this.

For the replay mode, I could imagine a:

```
- perpendicular-flap-one-way
  - fluid-aste
  - solid-dealii
  - solid ...
``` 
Or we just tell the user to copy the solid folders.

aste-turbine/precice.xml
kursatyurt: I have no idea about this part

davidscn:
I will tweak the documentation in a follow-up.

PR #245, opened 2021-12-08
Delete irrelevant couplingProperties file

MakisH:
This file seems to be a relic of the (very long) past. This has nothing to do with preCICE at the moment and it can lead to confusion.

Do we even need a changelog entry for this?

@KyleDavisSA please confirm that this is indeed irrelevant.


KyleDavisSA:
Test case runs just the same. This is an artifact from the old FOAM-FSI I think.

KyleDavisSA (approved):
Change does not affect simulation and the file is not required.

PR #247, opened 2021-12-31
changed visualize.py to python 3

### LINKED ISSUE #235: partitioned-elastic-beam: Visualization script is still in Python 2
https://github.com/precice/tutorials/blob/6d90ae3eb179113beb400e472c8832e29a48db9b/partitioned-elastic-beam/visualize.py

With a quick look, the `print` needs parentheses. After fixing this, Python complains about `xrange`. Maybe there are a few more.

mohamad-altaweel:
In Python 3, there is no xrange, but the range function behaves like xrange in Python 2. If you want to write code that will run on both Python 2 and Python 3, you should use range().

https://www.geeksforgeeks.org/range-vs-xrange-python/

Resolves #235 ,


IshaanDesai (approved):
Thanks for correcting this :+1:

Issue #176, opened 2021-03-27
perpendicular-flap: Results for CalculiX differ from FEniCS and Deal.II


BenjaminRodenberg:
# Problem & some Background

We observed a mismatch between results of the perpendicual flap case when preparing [this publication](https://arxiv.org/abs/2103.11191). In the end we used the Deal.II - SU2 and FEniCS - SU2 cases in the publication. They show good agreement. CalculiX - SU2 disagrees with these cases. Our current explanation for this behavior is that the CalculiX case uses `C3D8` elements, which should not be used according to the CalculiX documentation (see ["...the locking phenomena observed in the C3D8 element..."](https://petegustafson.com/CalculiX/ccx_2.15/doc/ccx/node28.html)). 

# Solution 

From my current point we should therefore use the "right" elements for CalculiX (no idea which elements are the right ones). Note that this will also require modifications of the calculix-adapter, since up to my knowledge currently only `C3D8` and `C3D8R` elements work (please correct me, if I'm wrong).

# Additional Material

For debugging this case we developed a `fluid-fake` solver that I would like to share in this issue (see [fluid-fake.zip](https://github.com/precice/tutorials/files/6215545/fluid-fake.zip)). This allows to replace the fluid solver with a simple constant force for quicker debugging. I'm plotting the tip displacement over time. Already this simple setup can be used to show the mismatch of CalculiX and the other two solvers:

![plot](https://user-images.githubusercontent.com/5740604/112718898-35c2f400-8ef6-11eb-828a-ccef1b536b3b.png)

## Code Versions

* `tutorials`: `a166efadfe7dbd3231e14897dcbeecebb90ea97e`
* `fenicsprecice`: `v1.0.1`
* `precice`: `v2.2.0`
* `pyprecice`: `v2.2.0.1`
* `calculix-adapter`: `4635aa87439d154269d7f6141e8684a733f3e68f`
* `dealii-adapter`: `f9c2c65eead05ecea1c2f7a1c7ff48fe5942f930`
* `dealii`: `v9.2.0`
* `calculix`: did not check this, but I think it should be `2.16`

IshaanDesai:
Some important points to note for this results mismatch:
* The CalculiX version used in this testing is `2.16` 
[...]

PR #250, opened 2022-01-24
Change C3D8 elements to C3D8I elements in perpendicular-flap solid-calculix

AndresPedemonteFIUBA:
This modifies the CalculiX case for the perpendicular-flap tutorial to use incompatible mode eight-node brick elements (C3D8I) instead of basic linear fully integrated eight-node brick elements (C3D8). 

C3D8 elements are prone to shear-locking, resulting in the structure appearing to be much stiffer than it actually is. Quoting the [CCX manual:](http://www.dhondt.de/ccx_2.16.pdf)
> Other elements frequently exhibit unsatisfactory behavior in certain instances,
e.g. the C3D8 element in bending states.

> the element tends to be to too stiff in bending 

C3D8I elements are much better for this case, where the flap bends:
>The incompatible mode eight-node brick element is an improved version of the C3D8-element. In particular, shear locking is removed and volumetric locking is much reduced. (...) The C3D8I element should be used in all instances in which linear elements are subject to bending.

This "fix" just requires chaning `TYPE=C3D8` to `TYPE=C3D8I` in `all.msh` and adding the `DIRECT` parameter to the `*DYNAMIC` keyword in `flap.inp`. This last addition is necessary to avoid CalculiX's automatic incrementation kicking in and changing the timestep size. 

The change should help bring the results obtained with CalculiX more in line with those obtained with other structural solvers (FEniCS, deal.II) for this case. See #176. 

Andrés


BenjaminRodenberg:
I will take care of reviewing this PR and also check how these changes affect the [results presented in the reference paper (see Fig 21)](https://arxiv.org/pdf/2109.14470v2.pdf).

BenjaminRodenberg (approved):
I applied the suggested changes and reran the examples presented in [the preprint for the preCICE reference paper](https://arxiv.org/abs/2109.14470v2). The results are much better with the suggested `C3D8I` elements and in good agreement with the other available solid solvers FEniCS and Deal.II:
plot
I used the preCICE vm version that is based on https://github.com/precice/tutorials/releases/tag/v202104.1.0, namely https://github.com/precice/vm/commit/66ee0cdddd9b2d131639b99181499a33fa2c892d. 

@AndresPedemonteFIUBA thank you again for this contribution and your help!

BenjaminRodenberg:
Merging this PR closes #176.

BenjaminRodenberg:
@AndresPedemonteFIUBA We would like to use the updated results in our paper and also point to this PR. If you want, we can also mention your github or real name. Can you write me a email to benjamin.rodenberg@in.tum.de, in case you want to be mentioned personally? If not, we can also refer to this PR without mentioning you personally.

Issue #131, opened 2020-12-19
Failure: FSI/flap_perp_2D/OpenFOAM-FEniCS


BenjaminRodenberg:
I just ran into an error, when trying to run `FSI/flap_perp/OpenFOAM-FEniCS` using the current state of develop (b74fdffb8062d6fd49733244be0d918643ec543c). Everything works for 13f101bb77461ea3f31451e3d22e6d72320c9f3e (before merging precice/fenics-adapter#71).

I used the following revision of the tutorial: https://github.com/precice/tutorials/pull/124/commits/ac6a94c6ae69b4dc7d4f6dd089ba57704e74bc35 (latest commit of https://github.com/precice/tutorials/pull/124). Observed the same behavior for https://github.com/precice/tutorials/commit/1c891a7882ddf3a2abdfa50ecafb941f1395eebb (latest commit of https://github.com/precice/tutorials/tree/develop).

IshaanDesai:
@MakisH need your perspective on this error. The conclusion is that nothing is wrong on the FEniCS side. Is this a reasonable assumption?

MakisH:
I see that it does solve a few iterations. How do the intermediate results look like visually? I guess that it either computes wrong forces or it receives wrong displacements.

@IshaanDesai would it also be easy for you to try with OpenFOAM v2006 and the adapter `develop` and document if you get the same error?

IshaanDesai:
> I see that it does solve a few iterations. How do the intermediate results look like visually?

![Screenshot from 2020-12-22 15-28-45](https://user-images.githubusercontent.com/44898158/102898919-81cf4a00-446a-11eb-8ad9-c492a7be619e.png)

This is the fluid domain just before the crash, it looks reasonable.

> @IshaanDesai would it also be easy for you to try with OpenFOAM v2006 and the adapter `develop` and document if you get the same error? Do you get any error with OpenFOAM 7 and any other FSI tutorial?

I checked with OpenFOAM v2006 and the latest `develop` branch of the adapter and sadly the error is the same and it occurs slightly sooner than the error with OpenFOAM 7. The error log is similar:
```
---[precice]  it 2 of 50 | dt# 43 | t 0.42 of 5 | dt 0.01 | max dt 0.01 | ongoing yes | dt complete no | read-iteration-checkpoint | 
---[preciceAdapter] [DEBUG] Reading a checkpoint...
---[preciceAdapter] [DEBUG] Reloaded time value t = 0.420000
[...]

PR #251, opened 2022-01-31
Add hint for time synchronization in ParaView

MakisH:
I added this in the perpendicular-flap case, as we get this issue, for example, with OpenFOAM-deal.II. In that case, applying the `TemporalShiftScale` filter with a `Scale: 0.1` synchronizes the results.

@DavidSCN do you think we should mention this anywhere else?

Edit: Switched to draft to cleanup some git spaghetti first.

### DIFF
--- a/partitioned-heat-conduction/README.md
+++ b/partitioned-heat-conduction/README.md
@@ -29,6 +29,8 @@ You can either couple a solver with itself or different solvers with each other.
 
 * Nutils. Install [Nutils](http://www.nutils.org/en/latest/).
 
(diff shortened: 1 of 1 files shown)


MakisH:
I thought I already integrated this one with some other PR, but apparently, I did not.

This also triggers me to think if we need a "general guidelines" in the tutorials overview (or a separate case), as tips like this apply to all cases, not only to the perpendicular flap.

@uekerman opinion?

MakisH:
Added in #131.

PR #253, opened 2022-02-02
Fix note regarding OpenFOAM and residual control

DavidSCN:
I received a message on discourse pointing out that our hint for the ´residualControl` for OpenFOAM foundation versions is actually not correct. I added a note here and also a link to the corresponding OpenFOAM bug.

### DIFF
--- a/elastic-tube-3d/fluid-openfoam/system/fvSolution
+++ b/elastic-tube-3d/fluid-openfoam/system/fvSolution
@@ -199,7 +199,7 @@ residualControl
 }
 */
 
-// OpenFOAM 6 (.org) or newer probably
+// OpenFOAM 6 (.org) or newer
 // see also https://bugs.openfoam.org/view.php?id=3336
 /*
 outerResidualControl
--- a/elastic-tube-3d/fluid-openfoam/system/fvSolution
+++ b/elastic-tube-3d/fluid-openfoam/system/fvSolution
@@ -199,9 +199,10 @@ residualControl
 }
 */
 
-// OpenFOAM 6 (.org) or newer
+// OpenFOAM 6 (.org) or newer probably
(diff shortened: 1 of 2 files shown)


MakisH (approved):
More specific [reference](https://github.com/OpenFOAM/OpenFOAM-dev/commit/4c8122783aedaa7dadf0486163a98350e625db32):

> ... Note that existing PIMPLE "residualControl" entries will need to be
renamed "outerCorrectorResidualControl".

The commit has tags pointing to OpenFOAM6 and newer, so we don't need the "probably"

elastic-tube-3d/fluid-openfoam/system/fvSolution
```diff
@@ -199,9 +199,10 @@ residualControl
 }
 */
 
-// OpenFOAM 6 (.org) or newer
+// OpenFOAM 6 (.org) or newer probably
```
MakisH, line 202:
```suggestion
// OpenFOAM 6 (.org) or newer
```

turek-hron-fsi3/fluid-openfoam/system/fvSolution
```diff
@@ -87,9 +87,10 @@ PIMPLE
 	}
     }
 
-    // OpenFOAM 6 (.org) or newer
+    // OpenFOAM 6 (.org) or newer probably
```
MakisH, line 90:
```suggestion
    // OpenFOAM 6 (.org) or newer
```

DavidSCN:
Interestingly, I cannot access the link, can you? The website states permission issues.

MakisH:
> Interestingly, I cannot access the link, can you? The website states permission issues.

I accessed both the issue tracker and the link to GitHub I posted just now.

PR #256, opened 2022-02-03
Making output directories for all FEniCS participants consistent

IshaanDesai:
The cleaning scripts now clean all FEniCS case outputs perfectly.


MakisH (approved):
I have not tried it, but the diff looks reasonable. Do you want me to check anything in particular?

IshaanDesai:
> Do you want me to check anything in particular?

Not in particular, just wanted another pair of eyes to look at the diff. Thanks!

PR #258, opened 2022-02-06
Cleanup OpenFOAM configuration files

MakisH:
This does a minor cleanup over all OpenFOAM cases. It is definitely a "high effort / low benefit" task, but these tiny aspects always looked strange and sometimes also triggered confusion.

Summary:
- Remove optional or completely foreign fields from the headers. A `FoamFile` only needs`version`, `format`, `class`, and `object` (mandatory but not used). The `location` is optional and it was, in many cases, misleading. Anything else is probably a relic from the original tool/source.
- Bring a bit of consistency in the indentation.
- Remove misleading and version-specific OpenFOAM copyright notices. These are not OpenFOAM files and it is quite a stretch to keep the headers here.
- Remove the `RAS` options from `turbulentProperties` (since we always set `laminar`).
- Remove some duplicate settings (setting the same value as the default) from `fvSchemes` in the heat-exchanger (probably relic from SimScale).
- Remove the unused `radiationProperties` from the heat-exchanger.
- Update the note regarding pimpleDyMFoam to make it really a special case that most people should not worry about.
- Some whitespace and formatting here and there.

We can probably still cleanup a lot if we look at the semantics and default settings, but we can also do this in a later point. If anyone has any suggestions for usual suspects to check and cleanup, they are very welcome!

I have collected all the changes for each case in one commit, for easier review.

@DavidSCN a second pair of eyes would be useful here.


partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
```diff
@@ -3,12 +3,11 @@ FoamFile
     version     2.0;
     format      ascii;
     class       dictionary;
-    location    "system";
     object      controlDict;
 }
 
 // Make sure you build the solver located
-// in ./openfoam-solver before running
+// in ../../openfoam-solver/ before running
```
DavidSCN, line 10: Is this hint actually needed? We don't have it in all places.

partitioned-heat-conduction/openfoam-neumann/system/controlDict
```diff
@@ -3,12 +3,11 @@ FoamFile
     version     2.0;
     format      ascii;
     class       dictionary;
-    location    "system";
     object      controlDict;
 }
 
 // Make sure you build the solver located
-// in ./openfoam-solver before running
+// in ../../openfoam-solver/ before running
```
DavidSCN, line 10: Is this hint actually needed? We don't have it in all places.

DavidSCN (approved):
Applied changes look good. I did not go through all the files myself.

partitioned-heat-conduction/openfoam-dirichlet/system/controlDict
MakisH: I am not sure what you mean. This is the only tutorial that uses a custom solver and it is located two directories up.

Are you suggesting removing the hint, or something else?

DavidSCN: Ah right, forget about it.

MakisH:
I have ran a few of the cases locally and I had another look at the diff. Merging now, so that any issues will be raised as soon as possible.

PR #261, opened 2022-02-08
Release (version upcoming)

MakisH:
(under construction)


BenjaminRodenberg:
I compared the OpenFOAM-Calculix and OpenFOAM-FEniCS results for the perpendicular flap and they look good. There is no visible regression. I plotted the `REF` results (based on the last distribution that we created for the latest version of the paper) and the `RC` results (using the most recent version of tutorials and adapters going into the distribution):

tutorials-perpendicular-flap-displacement-all-watchpoints

IshaanDesai (approved):
I ran a few tutorial cases with different solver combinations and everything seems to be in order :+1:

PR #267, opened 2022-03-14
Add workflow to update website

fsimonis:
This PR adds a trigger to update the website submodules.

This will be triggered by any change to `.md` files and `images/` in case folders.


MakisH (approved):
Nice! I understand the workflow and it looks clean. Have you already tested it? If it is too complicated to test, then just merge and we fix it later, if needed.

Issue #103, opened 2020-09-02
Missing flow-over-heated-plate with CalculiX


Alphaoo1:
Currently, CHT tutorials for OpenFOAM+FENICS, OpenFOAM+NUTILS and OpenFOAM+Aster are available.
Although, a tutorial case for OpenFOAM+CalculiX CHT is available, it is currently not working for me due to [issue](https://precice.discourse.group/t/openfoam-calculix-heat-exchanger-cht-tutorial-does-not-run/312) mentioned on discourse.

Also, it would be nice to have a simple case as compared to the heat-exchanger problem.

Thank you.

uekerman:
Yes, we should add the "flow over heated plate" also for OpenFOAM + CalculiX.

PR #271, opened 2022-04-10
Added CalculiX case for flow-over-heated-plate

boris-martin:
# Context

To showcase the improvements to 2D-3D coupling with CalculiX (https://github.com/precice/calculix-adapter/pull/94, https://github.com/precice/calculix-adapter/pull/92) and the support of hexaedral elements on face meshes (https://github.com/precice/calculix-adapter/pull/91), an new CHT case can be designed with CalculiX. It also improves the list of illustrated cases with heat transfer, as  the only CalculiX tutorial here with heat transfer is the heat exchanger, which is very specific (Robin BC's with explicit coupling on a steady-state simulation). This shows how to use DIrichlet-Neumann coupling on transient problems.

Closes https://github.com/precice/tutorials/issues/103

# Todolist

- [x] Cleanup of the case file and precice-config
- [x] Add a `run.sh` and a cleanup file
- [ ] Validation of the case. (For now: looks good but not exactly similar to other cases. Detailed comparison to do)
- [ ] format Python script (if we keep it ?)


flow-over-heated-plate/README.md
```diff
@@ -53,6 +53,9 @@ cd solid-fenics
 
 in order to use OpenFOAM and FEniCS for this test case. Feel free to try different combinations, they should all run and give approximately similar results.
 
+
+Note that to use CalculiX as solid solver, the configuration file `precice-config-calculix.xml` must be used instead of `precice-config.xml`. The CalculiX case is already configured to use it, but the fluid solver must be adapted. Modify the `system/preciceDict` file in the OpenFOAM case to use the correct one. This is necessary because CalculiX use different meshes for the temperature (read at the nodes) and for the heat flux, computed on face centers.
```
uekerman, line 57: This is problematic in my opinion. The CalculiX participant should use the same config. We need to bring temperature and heat flux values on the same mesh.

boris-martin: Not sure what's the right approach then. I think we can apply concentrad heat fluxes on nodes instead of faces, but then we'd need a conservative mapping, so here we go again for a new config file. Unless we want to add some dirty hack like "read fluxes on nodes then convert them to CalculiX faces fluxes, with some kind of averaging of a node for a given face"?
That would require a significant upgrade of the adapter but should be doable. Then we'd need to have either fluxes on faces or on nodes for backwards compatibility.

To be sure: do you dislike the fact that we need a different config for a tutorial, or the fact that CalculiX in general has two kinds of meshes? In the first case we could simply have a different tutorial?

Poke @MakisH as we discussed a bit about this.

boris-martin: I'll investigate then. Parts look easy (map from vertices to faces) but other directions should be harder. I'll let you know

MakisH: I also think that extending the CalculiX adapter to handle both cases would probably be the most sustainable solution here. We should still inform the user that there is an additional interpolation happening under the hood, though.

MakisH:
Since people are asking about this case, this PR is currently marked as "draft" and it seems to not be ready to run yet. For me, running (with `ccx_preCICE -i solid -precice-participant Solid`, as a `run.sh` is not yet available), leads to a configuration error:

```
---[precice]  This is preCICE version 2.4.0
---[precice]  Revision info: v2.4.0
---[precice]  Build type: Debug
---[precice]  Configuring preCICE with configuration "../precice-config.xml"
---[precice]  I am participant "Solid"
Using quasi 2D-3D coupling
Set ID Found 
---[precice] ERROR:  The given mesh name "Solid-Mesh-nodes" is unknown to preCICE. Please check the mesh definitions in the configuration.
```

Just check again in a few weeks.

boris-martin:
It's because the wrong precice-config is used. The precice-config-calculix must be used instead. I'll clean this up and move it to another directory with correct configuration. (Also, up-to-date develop version of the adapter is needed)

boris-martin:
Refactored this to be a new tutorial in a new folder. Big part of the diff is the copy of the `fluid-openfoam` from the non-CalculiX version of this. I have yet to write a new README, and there's a small issue where CalculiX crashes at the last time window which i'll have to investigate. Otherwise, ready for review.
If you want to run it, make sure you're using the develop branch of the CalculiX adapter.

flow-over-heated-plate/README.md
boris-martin: `flow-over-heated-plate-two-meshes` Looks good to me if you think that's a better idea.

MakisH:
I ran the case and the results look good, but only after a few iterations (this looks more like a steady-state case). In the beginning, the temperature around the interface is quite off:

https://user-images.githubusercontent.com/4943683/180025687-d0851f5c-ed67-4c7c-8281-c245e31a3873.mp4

Useful information for the README: here I converted the CalculiX results to VTK using the frd converter bundled in the CalculiX adapter. CalculiX produces 1000 result files, which one can then synchronize with OpenFOAM using the "Temporal Shift Scale" filter, using a scale of 0.01. Since the temperature is named differently in each solver, a "group" filter does not work, but setting the same scale to both (from 300K to 310K) helps.

boris-martin:
I have the same results. I'm a bit surprised we don't get a continuous temperature after a few iterations (since coupling is implicit, I thought it'd be enough). I don't know if it's a bug or if it's due to bad choice of mesh/mapping.

flow-over-heated-plate-calculix/README.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+
+{% note %}
+Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/flow-over-heated-plate). Read how in the [tutorials introduction](https://www.precice.org/tutorials.html).
+{% endnote %}
+
+## Setup
+
+The scenario is exactly the same as the one described in the [flow over heated plate tutorial](http://precice.org/tutorials-flow-over-heated-plate.html). However, this tutorial is specialized for the case when heat fluxes and temperatures live on different meshes. This is the case with CalculiX: heat fluxes are defined on face centers, while temperatures are read at nodes. This requires updating the `precice-config.xml` file to take this into account. On the fluid side, a unique mesh can still be used.
```
MakisH, line 15: The motivation sounds clear. Minor consistency issue:

```suggestion
The scenario is exactly the same as the one described in the [flow over heated plate tutorial](http://precice.org/tutorials-flow-over-heated-plate.html). However, this tutorial is specialized for the case when heat fluxes and temperatures live on different meshes. This is the case with CalculiX: heat fluxes are written on face centers, while temperatures are read on nodes. This requires updating the `precice-config.xml` file to take this into account. On the fluid side, a single mesh can still be used.
```

(feel free to change the `read`, instead)

flow-over-heated-plate-calculix/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+By default, the fluid participant reads heat-flux values and the solid participant reads temperature values for the coupled simulation. The following participants are available:
+
+Fluid participant:
+
+* OpenFOAM (buoyantPimpleFoam). For more information, have a look at the [OpenFOAM adapter documentation](https://www.precice.org/adapter-openfoam-overview.html).
+
+Solid participant:
+
+* CalculiX. For more information, have a look at the [CalculiX adapter documentation](http://precice.org/adapter-calculix-overview.html).
```
MakisH, line 27: Shall we document that this needs a minimum version of the adapter? (once available)

flow-over-heated-plate-calculix/README.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Flow over heated plate with two meshes
+permalink: tutorials-flow-over-heated-plate-two-meshes.html
```
MakisH, line 3: Don't forget to add this to the sidebar on the website (separate PR).

flow-over-heated-plate-calculix/solid-calculix/generate_mesh.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
+
```
MakisH, line 2: I understand that this script is specific to this case (and very helpful). Maybe document it in the README?

MakisH:
Thanks for the README! Still open:

- [ ] The strange "double free" error at the end of the CalculiX simulation.
- [ ] Closer look at the results / tuning

flow-over-heated-plate-calculix/solid-calculix/generate_mesh.py
boris-martin: Not sure about the optimal location. Another section ? An additional sentence in the available solvers section ?
But I agree it's worth mentioning.

uekerman:
Let's also add a link from the normal `flow-over-heated-plate` to this one.

boris-martin:
> Let's also add a link from the normal `flow-over-heated-plate` to this one.

There's already one in the Setup description

uekerman:
> There's already one in the Setup description

Is there? This PR only touches `flow-over-heated-plate-calculix` right now, right?

boris-martin:
Oh my bad I misread. There's one in the other direction, but the one you mention is indeed missing.

flow-over-heated-plate-calculix/solid-calculix/generate_mesh.py
MakisH: Actually, where is this called from? Because it is not in the `run.sh`. If it is not called at the moment, let's call it and remove the produced mesh. Otherwise, the script will get forgotten very soon.

boris-martin:
I changed the solid mesh to make it much more refined (~same vertex density as the fluid at the interface). If I do that and modify OpenFOAM to have one output every 0.01s (i.e. one per time window), I get something good-looking (using the time shift you recommended, which I'll have to document). Also, I quite increased the implicit coupling accuracy (reduced tolerance, 1e-9 instead of 1e-5.:

flow-plate-fine

I had weird things using the original output interval (0.20), like times being [0.19, 0.20, 0.39, 0.40, ...]. @MakisH I need your OpenFOAM expertise to understand that part.

In conclusion, I got good results by changing a lot of things, I'll have to isolate the dominant one because I kind of lost track of all these changes. But we're getting there, and the adapters aren't broken :)

MakisH:
> I had weird things using the original output interval (0.20), like times being [0.19, 0.20, 0.39, 0.40, ...]. @MakisH I need your OpenFOAM expertise to understand that part.

This is a known issue we can do little about: https://github.com/precice/openfoam-adapter/issues/26

That's why at the end of every OpenFOAM `run.sh`, we call the `removeObsoleteFolders`.

boris-martin:
I ran again without changing preCICE implicit coupling tolerance and results look satisfying to me. Should be getting ready then. I think the calculix crash at the end was solved in https://github.com/precice/calculix-adapter/pull/102.
Updated the main README too.

flow-over-heated-two-meshes/README.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Flow over heated plate with two meshes
+permalink: tutorials-flow-over-heated-plate-two-meshes.html
+keywords: tutorial, CHT, conjugate-heat transfer, OpenFOAM, CalculiX
+summary: This tutorial describes how to run a conjugate heat transfer coupled simulation using preCICE and CalculiX as solid solver, which requires the usage of two meshes instead of one.
```
MakisH, line 5: Concision:

```suggestion
summary: This tutorial describes how to run a conjugate heat transfer coupled simulation using preCICE and CalculiX as solid solver, which requires two meshes instead of one.
```

flow-over-heated-two-meshes/README.md
```diff
@@ -0,0 +1,53 @@
...
+```bash
+cd solid-calculix
+./run.sh
+```
+
+## Post-processing
+
+On the OpenFOAM side, you can open the `.foam` file with ParaView, or create VTK files with `foamToVTK`. CalculiX outputs `.frd` files which can be opened with `cgx` or converted into VTK files using the converter available in the adapter repository.
+
+CalculiX produces 1000 result files, which one can then synchronize with OpenFOAM using the "Temporal Shift Scale" filter, using a scale of 0.01. This is because CalculiX considers that each time window (of length 0.01 by default) is one time step, and increases the total time by 1 in the VTK files. For efficient visualization, use the same scale for temperature on both outputs; we recommend from 300K to 310K.
```
MakisH, line 49: It is not just a recommendation, but this is what the global minimum and maximum should actually be, if the simulation is correct:

```suggestion
CalculiX produces 1000 result files, which one can then synchronize with OpenFOAM using the "Temporal Shift Scale" filter, using a scale of 0.01. This is because CalculiX considers that each time window (of length 0.01 by default) is one time step, and increases the total time by 1 in the VTK files. For efficient visualization, use the same scale for temperature on both outputs (from 300K to 310K, which should be the minimum and maximum values in this setup).
```

flow-over-heated-two-meshes/README.md
```diff
@@ -0,0 +1,53 @@
+---
```
MakisH, line 1: The directory name is missing a `-plate-`.

flow-over-heated-two-meshes/fluid-openfoam/system/controlDict
```diff
@@ -0,0 +1,46 @@
...
+
+stopAt          endTime;
+
+endTime         1;
+
+deltaT          0.01;
+
+writeControl    runTime;
+
+writeInterval   0.2;
```
MakisH, line 23:
```suggestion
writeInterval   0.01;
```
Let's actually set the OpenFOAM output frequency to be the same as CalculiX. This will make the animation less awkward.

MakisH:
I realize too late that this needs two more fixes:
- Making OpenFOAM write as frequently as CalculiX did not help and actually made the simulation quite slow. I should have actually thought this a bit longer and suggested to make CalculiX write as infrequently as OpenFOAM. But most importantly, we should probably write in every coupling time. Otherwise, we get inconsistencies between coupling times. But this may also hide a deeper problem.
- We should make the CalculiX plate as thick as the OpenFOAM channel. Right now, the one is 0.2m thick, and the other 0.5m, leading to strange visuals in paraview.

PR #273, opened 2022-04-26
Minor update to first two Nutils examples

scdivi:
The current active & stable nutils version is v7.0 (28ae68f3). With this pull request the first two examples are updated to nutils v7.0. At the moment the only change in this update is the way the libraries are imported. In order to have backward compatibility to previous nutils version, i.e., v6.3 (87281a0b), one needs to change only the importing style. The two styles of library import associated to the nutils version is demonstrated below.

nutils v7.0 (28ae68f3):

```python
from nutils import cli, solver

def main()
      lhs = solver(...)

cli.run(main)
```

nutils v6.3 (87281a0b):

```python
import nutils

def main():
      lhs = nutils.solver(...)

nutils.cli.run(main)
```

### DIFF
--- a/flow-over-heated-plate/solid-nutils/solid.py
+++ b/flow-over-heated-plate/solid-nutils/solid.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python3
 
-import nutils
+from nutils import cli, mesh, function, solver, export 
 import treelog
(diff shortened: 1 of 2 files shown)


IshaanDesai (approved):
Looks good :+1:

PR #278, opened 2022-06-16
Chemical reactions with FEniCS

boris-martin:
# Description of the case
Added a new tutorial for partitioned advection-diffusion-reaction of chemical species in a incompressible fluid.
- Fluid participant solves the NS equations and sends its velocity to Chemical participant
- Chemical participants computes diffusion, production and reaction of chemical species (A + B -> C with sources of A, B) as well as advection through the fluid, reading velocity from the Fluid participant.

Coupling is unidirectional, serial explicit.

Due to tight time stepping constraints on the fluid flow and high cost of the non-linear chemical reactions, different time scales are used: the fluid solver is subcycling.

This test case was made for showcasing Linear Cell Interpolation data mapping.

# Todo-list

- [x] Writing README
- [x] add run.sh and clean.sh scripts
- [x]  add some measurements: the chemical case keeps track of total quantities of A, B and C species over time (through FEniCS built-in support of integrals)
- [x]  General cleaning, formatting, etc
- [x]  Investigate better discretization choices (e.g. more accurate schemes, like mid-point rule for velocities, choose between 2nd order and 1st order elements for chemistry)


uekerman:
Nice :+1:
Alternative names could be `channel-chemistry` or `channel-transport-reaction`. It is a channel flow with no-slip walls, right?
This would put the case close to the existing `channel-transport`.

boris-martin:
Yes it's a channel. I imagine `channel-transport-reaction` is the most appropriate since it describes what is coupled. On the other hand, no reference to chemistry at all doesn't seem right.

uekerman:
You could say that "chemistry" is already covered by "reaction"? :grin:

channel-transport-reaction/README.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Channel transport reaction
+permalink: tutorials-channel-transport-reaction.html
+keywords: FEniCS, Volumetric coupling, reaction, transport
+summary: We simulate chemical reactions in a moving fluid. One solver simulates the fluid flow and another handles the reactions.
```
IshaanDesai, line 5:
```suggestion
summary: We simulate an advection-diffusion-reaction scenario in a fluid in motion. One solver simulates the fluid flow and another handles the reactions.
```

channel-transport-reaction/README.md
```diff
@@ -0,0 +1,48 @@
...
+---
+
+{% note %}
+Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/channel-transport-reaction). Read how in the [tutorials introduction](https://www.precice.org/tutorials.html).
+{% endnote %}
+
+## Setup
+
+We solve a system of advection-diffusion-reaction in a fluid flowing past a cylinder. This case is inspired from [1]. A fluid flow inside a channel containing a cylinder. Chemical species A, B and C live in the fluid, with sources of A and B at the entry of the channel. Due to diffusion and fluid mixing, A and B react to form some C.
```
IshaanDesai, line 14:
```suggestion
We solve a system of advection-diffusion-reaction in a fluid flowing past a cylinder. This case is inspired from a FEniCS tutorial [1]. A fluid flows inside a channel containing a cylinder as an obstacle. Chemical species A, B and C live in the fluid, with sources of A and B at the entry of the channel. Due to diffusion and fluid mixing, A and B react to form a certain amount of C.
```

channel-transport-reaction/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Insert pic**
+
+The simulation is split into two participants: a Fluid participant that computes the fluid flow and sends the velocity field to the Chemical participant, computing the reactions, the diffusion and advection of all chemical species. The coupling is unidirectional (`serial-explicit` with only one data transfer).
```
IshaanDesai, line 18:
```suggestion
The simulation is split into two participants: a Fluid participant that computes the fluid flow and sends the velocity field to the Chemical participant which computes the diffusion and advection of all chemical species. The coupling is unidirectional (`serial-explicit` with only one data entity being transferred).
```

channel-transport-reaction/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Available solvers and dependencies
+
+Both participants run on FEniCS. Install [FEniCS](https://fenicsproject.org/download/) and the [FEniCS-adapter](https://github.com/precice/fenics-adapter).
```
IshaanDesai, line 22:
```suggestion
Both participants use the FEniCS library. Install [FEniCS](https://fenicsproject.org/download/) and [fenicsprecice](https://github.com/precice/fenics-adapter).
```

channel-transport-reaction/chemistry-fenics/chemistry-precice.py
```diff
@@ -0,0 +1,116 @@
...
+mesh = generate_mesh(domain, 64)
+normal = FacetNormal(mesh)
+
+# Three dimensional vector for three species
+P = FiniteElement('P', triangle, 1)
+V = FunctionSpace(mesh, MixedElement([P, P, P]))
+W = FunctionSpace(mesh, MixedElement([P, P])) # Velocity field is 2D
+
+# No BCs: only homgenuous Von Neumann. It means nothing leaves the domain
+# Von neumann BCs give rise to surface loads, which are simply zero here.
```
IshaanDesai, line 22:
```suggestion
```

channel-transport-reaction/chemistry-fenics/chemistry-precice.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Initialize preCICE
+precice = fenicsprecice.Adapter(adapter_config_filename="chemistry-config.json")
+precice_dt = precice.initialize(coupling_subdomain=CouplingDomain(), read_function_space=W)
+
+flow_expr = precice.create_coupling_expression()
+
+dt = np.min([default_dt, precice_dt])
+
+# Create used terms (_n suffix means "at previous step")
```
IshaanDesai, line 36:
```suggestion
# Create used terms (_nm1 suffix means "at previous step")
```

The suffix `_n` for values from previous time steps is rather confusing. In other FEniCS tutorials we use `_nm1` which means `n - 1`.

channel-transport-reaction/README.md
boris-martin: Not a big fan of naming the link this way. Using "The X adapter" seems common in the rest of the documentation. See e.g. http://precice.org/tutorials-perpendicular-flap.html

channel-transport-reaction/chemistry-fenics/chemistry-precice.py
boris-martin: Do you want to remove this documentation part completely ?

channel-transport-reaction/chemistry-fenics/chemistry-precice.py
boris-martin: what about something like `_prev`? I've never seen `nm1` and the tutorial I used as basis used `_n`

IshaanDesai: If the original tutorial uses `_n` then it is completely alright, lets keep it this way.

channel-transport-reaction/chemistry-fenics/chemistry-precice.py
IshaanDesai: I would remove it as on first look it does not provide valuable information. If we are to keep it, can you make it a bit more detailed?

boris-martin: I put a more detailed version around line 60, closer to the formulation

channel-transport-reaction/README.md
IshaanDesai: Okay, lets keep this consistent and not make this change. Reading fenicsprecice is indeed a bit weird.

PR #281, opened 2022-07-12
Use waveform iteration with partitioned-heat-conduction/fenics/heat.py

BenjaminRodenberg:
This PR introduces waveform iteration for the partitioned-heat-conduction case. Currently only the fenics-adapter supports this feature via https://github.com/precice/fenics-adapter/pull/153 + https://github.com/precice/python-bindings/pull/147. This also means that the tutorials for other adapters/solvers will break.

I think that the partitioned-heat-conduction case + subcycling is a nice test case for waveform iteration + subcycling. However, we might want to shift this specialized configuration of the case into an independent folder (as long as not all adapters support waveform iteration).

This is a breaking change, because some required features will only be introduced with preCICE v3.0.0.

### DIFF
--- a/partitioned-heat-conduction/fenics/heat.py
+++ b/partitioned-heat-conduction/fenics/heat.py
@@ -62,7 +62,7 @@ parser.add_argument("-e", "--error-tol", help="set error tolerance", type=float,
 
 args = parser.parse_args()
 
-fenics_dt = .1  # time step size
+fenics_dt = .01  # time step size
 # Error is bounded by coupling accuracy. In theory we would obtain the analytical solution.
 error_tol = args.error_tol
 
@@ -178,7 +178,7 @@ while precice.is_coupling_ongoing():
     if precice.is_action_required(precice.action_write_iteration_checkpoint()):
         precice.store_checkpoint(u_n, t, n)
 
-    read_data = precice.read_data()
+    read_data = precice.read_data(dt(0))
 
     # Update the coupling expression with the new read data
     precice.update_coupling_expression(coupling_expression, read_data)
--- a/partitioned-heat-conduction/precice-config.xml
+++ b/partitioned-heat-conduction/precice-config.xml
@@ -7,7 +7,7 @@
       enabled="true" />
   </log>
 
-  <solver-interface dimensions="2">
+  <solver-interface dimensions="2" experimental="true">
     <data:scalar name="Temperature" />
     <data:scalar name="Heat-Flux" />
 
@@ -25,7 +25,7 @@
       <use-mesh name="Dirichlet-Mesh" provide="yes" />
       <use-mesh name="Neumann-Mesh" from="Neumann" />
       <write-data name="Heat-Flux" mesh="Dirichlet-Mesh" />
-      <read-data name="Temperature" mesh="Dirichlet-Mesh" />
+      <read-data name="Temperature" mesh="Dirichlet-Mesh" waveform-order="1"/>
       <mapping:rbf-thin-plate-splines
         direction="read"
         from="Neumann-Mesh"
@@ -38,7 +38,7 @@
       <use-mesh name="Neumann-Mesh" provide="yes" />
       <use-mesh name="Dirichlet-Mesh" from="Dirichlet" />
       <write-data name="Temperature" mesh="Neumann-Mesh" />
-      <read-data name="Heat-Flux" mesh="Neumann-Mesh" />
+      <read-data name="Heat-Flux" mesh="Neumann-Mesh" waveform-order="0"/>
       <mapping:rbf-thin-plate-splines
         direction="read"
         from="Dirichlet-Mesh"


BenjaminRodenberg:
Basically this tutorial is ready to merge now that `precice/develop` is broken. Since it shifts the focus a bit (not simple tutorial, but a bit more advanced with subcycling), we should first agree here whether we actually want this change.

On the other hand: We sooner or later will need tutorials that illustrate how to use subcycling and this one along with https://github.com/precice/tutorials/tree/master/oscillator are good candidates. Any opinion @uekerman ?

uekerman:
Yes, let's use new features (here waveform relaxation) whereever reasonable. I don't see any reason against this.
This will now also require the develop version of the FEniCS adapter, right? I think we should document such knowledge somewhere while updating everything.
@MakisH Where would be a good place?

MakisH:
Having tutorials for such new features is indeed quite important for communicating and teaching the new features. Since the partitioned-heat-conduction was already designed to be a tutorial for studying the methods themselves, it makes a lot of sense to use it here.

The first place to document this change would be in the changelog (missing here). We should then mention it in the release notes / blog post.

Regarding the changes required in adapters, they need to go to the [v3 porting guide](https://precice.org/couple-your-code-porting-v2-3.html). I think it would be great to document the necessary changes there, and/or link to the [waveform feature page](https://precice.org/couple-your-code-waveform.html#undefined) from there.

This tutorial also has OpenFOAM and Nutils participants.

For the OpenFOAM adapter, we have an umbrella issue to migrate to v3, where waveform is already on the list: https://github.com/precice/openfoam-adapter/issues/235

For Nutils, we need an issue here - opened: https://github.com/precice/tutorials/issues/312

partitioned-heat-conduction/precice-config.xml
```diff
@@ -38,7 +38,7 @@
       <use-mesh name="Neumann-Mesh" provide="yes" />
       <use-mesh name="Dirichlet-Mesh" from="Dirichlet" />
       <write-data name="Temperature" mesh="Neumann-Mesh" />
-      <read-data name="Heat-Flux" mesh="Neumann-Mesh" />
+      <read-data name="Heat-Flux" mesh="Neumann-Mesh" waveform-order="0"/>
```
BenjaminRodenberg, line 41: This actually looks wrong to me. Probably it still works here, because both solvers work at the same rate and implicit Euler only samples at the end of the step. So actually no interpolation is required, but the exchange of the substeps still matters.

partitioned-heat-conduction/precice-config.xml
```diff
@@ -7,7 +7,7 @@
       enabled="true" />
   </log>
 
-  <solver-interface dimensions="2">
+  <solver-interface dimensions="2" experimental="true">
```
BenjaminRodenberg, line 10: We don't need this anymore for v3, because waveform relaxation is not experimental

MakisH:
@BenjaminRodenberg the only feature required is to read data at specific times, right? This is already implemented in the OpenFOAM adapter and in the Nutils examples here (in develop at least).

I cannot test the OpenFOAM at the moment, as I don't have `funkySetFields` in this system. Nutils seems to at least run. What should I check?

BenjaminRodenberg:
> I cannot test the OpenFOAM at the moment, as I don't have `funkySetFields` in this system. Nutils seems to at least run. What should I check?

If you can port the OpenFOAM version of the tutorial to a waveform iteration version following this PR, this might help you debugging https://github.com/precice/openfoam-adapter/pull/305. The only thing you have to keep in mind is that quasi Newton is not supported. But you should be able to get the analytical solution -- even with subcycling.

MakisH:
> > I cannot test the OpenFOAM at the moment, as I don't have `funkySetFields` in this system. Nutils seems to at least run. What should I check?
> 
> If you can port the OpenFOAM version of the tutorial to a waveform iteration version following this PR, this might help you debugging [precice/openfoam-adapter#305](https://github.com/precice/openfoam-adapter/pull/305). The only thing you have to keep in mind is that quasi Newton is not supported. But you should be able to get the analytical solution -- even with subcycling.

This PR just works with the OpenFOAM case and the results look fine.

In the OpenFOAM adapter develop, we are already reading at the relative time:

https://github.com/precice/openfoam-adapter/blob/c3968687a9cbaeacc599730681cb266108dd3a71/Interface.C#L552-L557

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -62,7 +62,7 @@ def determine_gradient(V_g, u, flux):
 
 args = parser.parse_args()
 
-fenics_dt = .1  # time step size
+fenics_dt = .01  # time step size
```
MakisH, line 65: I somehow cannot push to your fork (even though you have allowed maintainers to edit). For OpenFOAM, you just need to change the following in `openfoam-dirichlet` and `openfoam-neumann`, in the file `system/controlDict`:

- Set `deltaT` to 0.01
- To observe the intermediate results, also set `writeInterval` to 0.01

MakisH (commented):
Reducing the dt (similarly to FEniCS), I get a non-smooth interface in the intermediate times, with a gap that increases as the relative read time increases. Is this expected with a `waveform-order=1`?

Screenshot from 2023-11-14 00-05-34

partitioned-heat-conduction/fenics/heat.py
MakisH: I guess we are doing something terribly wrong in the OpenFOAM adapter. Let's look at it in the coding days, or together sometime this week.

MakisH: Discussing with @BenjaminRodenberg, the problem seems to be just in visualization:

OpenFOAM only writes result files the first iteration of each timestep. At the end of the coupling time window, we enforce OpenFOAM to write again. Doing it for every time step, however, would be a performance killer.

In FEniCS, we control what results to write when. In the OpenFOAM adapter, this would be quite cumbersome. Hence, we decided to only document that only results at the end of the coupling time window make sense.

Follow-up for OpenFOAM: https://github.com/precice/tutorials/pull/406

MakisH (approved):

PR #284, opened 2022-07-20
Modal dynamic simulation of the perpendicular flap with CalculiX

boris-martin:
Adds a variant of the CalculiX case for the perpendicular flap, using modal dynamic simulations: a few eigenfrequencies are computed, then a simulation is run where all loads are projected into the subspace formed by the used eigenmodes.

Showcases https://github.com/precice/calculix-adapter/pull/99.


boris-martin:
@uekerman I'd need an opinion on the structure: is it appropriate to have 2 CalculiX subfolders for the same tutorial ? It looks cleaner to me than having both cases in the same folder.

uekerman:
How different are both cases? Is it only a different `inp` file?

boris-martin:
> How different are both cases? Is it only a different `inp` file?

It's a different process. Here there are two `inp` files: one for the frequency analysis, and one for the actual simulation (which reads stored frequencies). It's quite tricky to setup with no experience, which is why I wanted to add it.

uekerman:
I ask, because if it is only two different `inp` files, I would not go for different subfolders.
We could add an argument to the `run.sh` to distinguish both. Similar to how we treat Dirichlet and Neumann BCs in the [partitioned heat tutorial](https://precice.org/tutorials-partitioned-heat-conduction.html).

> It's quite tricky to setup with no experience, which is why I wanted to add it.

Completely agree :+1:

boris-martin:
It's two more files instead of one, but a shared `run.sh` would work yes.

uekerman:
> It's two more files instead of one, but a shared run.sh would work yes.

I have a slight tendency towards one folder, but of course depends on how much hassle it is.

@MakisH opinions?

boris-martin:
I wrote a working `run.sh` script with argument. I'll push it once we agree on this if we decide to keep one case.
(Now that it's done I'm in favor of this approach too)

perpendicular-flap/README.md
```diff
@@ -31,7 +31,7 @@ Solid participant:
 
 * FEniCS. The structural model is currently limited to linear elasticity. For more information, have a look at the [FeniCS adapter documentation](https://www.precice.org/adapter-fenics.html).
 
-* CalculiX. In order to allow a reasonable comparison to all solid codes, the geometrically non-linear solver has been disabled and only a linear model is used by default. For more information, have a look at the [CalculiX adapter documentation](https://www.precice.org/adapter-calculix-overview.html)
+* CalculiX. In order to allow a reasonable comparison to all solid codes, the geometrically non-linear solver has been disabled and only a linear model is used by default. For more information, have a look at the [CalculiX adapter documentation](https://www.precice.org/adapter-calculix-overview.html). Two cases are provided: one as a regular simulation, and one with modal dynamic simulations where a few eigenmodes are computed then used to simulate a reduced model.
```
MakisH, line 34: I think we could have 1-2 extra sentences here with more details on what/why/how. Right now, it is not clear to me what this does, I just understand that if I was already into this topic I could have understood, but still one needs to look at the code.

Is there any literature describing this case? Are the results supposed to be directly comparable to the `solid-calculix` case?

perpendicular-flap/solid-calculix-modal/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e -u
+
+ccx_preCICE frequency
+mv frequency.eig flap.eig
+ccx_preCICE -i flap -precice-participant Solid
```
MakisH, line 6: (Continuing the one-vs-two directories here, to actually have a thread)

Running a diff between the two cases, I see that the difference of the two cases is very small in terms of config:
- One extra file `frequency.inp` (which probably does not harm having in the non-modal case as well)
- Change in `flap.inp` from `*DYNAMIC, ALPHA=0.0, DIRECT` to `*MODAL DYNAMIC DIRECT`
- This `run.sh`

I would be in favor of a single folder, if that is possible, to avoid duplication. Please push your approach with the one configurable `run.sh` script.

MakisH:
I also ran the case and it completed, while the results look reasonable. I am not sure if I should look for any differences compared to the current `solid-calculix` case.

boris-martin:
> I also ran the case and it completed, while the results look reasonable. I am not sure if I should look for any differences compared to the current `solid-calculix` case.

Results should be very similar. If you reduce the number of frequencies (say 1 instead of 10) the difference becomes noticeable.

perpendicular-flap/README.md
boris-martin: It depends on how much CalculiX knowledge is expected. I always struggle to find a balance between "complete documentation" and "writing so much it becomes a CalculiX documentation and not an adapter documentation".

Since this is essentially a reduced order model, results should be comparable (but slightly less accurate).

Maybe add a link to the adapter doc mentioning this feature ? (Once there is one)

MakisH: Linking to the adapter documentation is something we should do, yes. We can explain more there. But here the question is not so much about CalculiX, but about modal analysis. Users need to know if that's relevant for them.

perpendicular-flap/solid-calculix/run.sh
```diff
@@ -1,4 +1,18 @@
 #!/bin/sh
 set -e -u
 
-ccx_preCICE -i flap -precice-participant Solid
+usage() { echo "Usage: run.sh [-modal]" 1>&2; exit 1; }
+
+if [ $# -ge 2 ] || [ $# -eq 1 -a $1 != "-modal" ]; then
+    usage
+fi
+#ccx_preCICE -i flap -precice-participant Solid
```
MakisH, line 9: Why is this commented out?

perpendicular-flap/solid-calculix/run.sh
```diff
@@ -1,4 +1,18 @@
 #!/bin/sh
 set -e -u
 
-ccx_preCICE -i flap -precice-participant Solid
+usage() { echo "Usage: run.sh [-modal]" 1>&2; exit 1; }
+
+if [ $# -ge 2 ] || [ $# -eq 1 -a $1 != "-modal" ]; then
```
MakisH, line 6: [shellcheck](https://www.shellcheck.net/) complains here (and my shell refuses to run without `-modal`):

```
In run.sh line 6:
if [ $# -ge 2 ] || [ $# -eq 1 -a $1 != "-modal" ]; then
                              ^-- SC2166 (warning): Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
                                 ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
if [ $# -ge 2 ] || [ $# -eq 1 -a "$1" != "-modal" ]; then


In run.sh line 12:
if [ ${1-nonmodal} = "-modal" ]; then
     ^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
if [ "${1-nonmodal}" = "-modal" ]; then

For more information:
  https://www.shellcheck.net/wiki/SC2166 -- Prefer [ p ] && [ q ] as [ p -a q...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

perpendicular-flap/solid-calculix/run.sh
```diff
@@ -1,4 +1,18 @@
...
+    usage
+fi
+#ccx_preCICE -i flap -precice-participant Solid
+
+
```
MakisH, line 11: We still need to extend the cleaning function to remove the files `WarnNodeMissMultiStage.nam` and `flap_modal.eig` (or maybe just files of these extensions).

perpendicular-flap/solid-calculix-modal/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e -u
+
```
MakisH, line 3: Don't forget to remove this directory at some point.

perpendicular-flap/solid-calculix/run.sh
boris-martin: Leftover from copy-paste, I'll remove it

perpendicular-flap/README.md
boris-martin: I don't get what you would want to add. From a CalculiX user's point of view, the only documentation is "you can use modal dynamic simulations with preCICE, and configuration is the same as with a regular simulation"

perpendicular-flap/solid-calculix/run.sh
boris-martin: Fixed it, does the new version looks good to you?

perpendicular-flap/README.md
MakisH: The updated state is already quite good, thanks!

MakisH (approved):

PR #285, opened 2022-07-27
Fix various typos

luzpaz:


MakisH (approved):
All changes look indeed like typos and the changes are as they were meant to be, thanks! :smile: 

I changed the target branch to `develop`.

PR #286, opened 2022-07-31
Add perpendicular-flap case solid-solids4foam and solid-openfoam

solids4foam:
(description by @MakisH)

This adds two solid participant cases in the [perpendicular flap](https://precice.org/tutorials-perpendicular-flap.html) tutorial:
- `solid-openfoam` (using the standard OpenFOAM solver solidDisplacementFoam)
- `solid-solids4foam` (using the solids4Foam solver of solids4foam)

Note that this depends on https://github.com/precice/openfoam-adapter/pull/236 and in the nextRelease development line of solids4foam.


perpendicular-flap/solid-solids4foam/0/D
```diff
@@ -0,0 +1,42 @@
+/*--------------------------------*- C++ -*----------------------------------*\
+| =========                 |                                                 |
+| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
+|  \\    /   O peration     | Version:  v2012                                 |
+|   \\  /    A nd           | Website:  www.openfoam.com                      |
+|    \\/     M anipulation  |                                                 |
+\*---------------------------------------------------------------------------*/
```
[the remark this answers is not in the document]
solids4foam: Understood. I will devise a simple script to remove these top 7 lines from each file.

One benefit of the first (and closing 7th line) is that it tells text editors which language highlight syntax to use; in this case, for example, my emacs knows to use C++ highlighting. One option would be to keep a modified version of line 1:
`/*--------------------------------*- C++ -*----------------------------------*/`
What do you think? Alternatively, for consistency (which I agree is very important), then I would just remove all 7 lines.

perpendicular-flap/solid-solids4foam/clean.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Source tutorial clean functions
+. $WM_PROJECT_DIR/bin/tools/CleanFunctions
+
+cleanCase
+rm -rf log.* *.log
+rm -f precice-Solid-events.json
+rm -rf history
```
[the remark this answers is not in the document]
solids4foam: Sure. I have made this change in the latest commit: 121cbc1e0487cd4b5b666a15565d7fcaf6cead47

perpendicular-flap/solid-solids4foam/constant/mechanicalProperties
```diff
@@ -0,0 +1,31 @@
...
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+planeStress no;
+
+mechanical
+(
+    rubber
+    {
+        type        linearElastic;
+        //type        neoHookeanElastic;
```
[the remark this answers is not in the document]
solids4foam: Currently, the case is set up to run as "linear geometry" like the other example solid solvers for this tutorial. If the user wants to run this case a "nonlinear geometry", then two changes are required:

> In constant/solidProperties, change "solidModel" from "linearGeometryTotalDisplacement" to "nonLinearGeometryTotalLagrangianTotalDisplacement"
> In constant/mechanicalProperties (file we are discussing here), change "type" from "linearElastic" to "neoHookeanElastic"

So, what to do here: for tidiness and consistency with other tutorials, I suggest I remove all of these commented options and then details on how to change the case (e.g. from linear to nonlinear geometry) can be given in a short tutorial page or similar. What do you think?

perpendicular-flap/solid-solids4foam/0/D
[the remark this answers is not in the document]
MakisH: I would just go with the consistent way for now, but good to know!

Let's try to find other ways to let editors know the type. I opened an issue: https://github.com/precice/tutorials/issues/288

(btw, you can tell GitHub to treat `.C` files as C++: https://github.com/precice/openfoam-adapter/blob/develop/.gitattributes)

perpendicular-flap/solid-solids4foam/constant/mechanicalProperties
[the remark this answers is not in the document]
MakisH: Since multiple changes are needed, I would offload that to the tutorial's `README.md`. Even better, if this is a common situation, I would offload it to the solids4foam documentation (is it maybe mentioned in some tutorial we can link to?).

perpendicular-flap/solid-solids4foam/system/controlDict
```diff
@@ -0,0 +1,62 @@
...
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      controlDict;
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+application     solids4Foam;
```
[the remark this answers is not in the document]
solids4foam: Good question and one that I have considered but have not made any rigid declarations on.

Here are my current thoughts:

- The solver executable name "solids4Foam", i.e. this is the command needed to run the solver
- I typically refer to the overall project (as opposed to the solver executable) as "solids4foam" (all lowercase).

From your nice documentation website, I have become motivated to set up a basic solids4foam page; that will be the ideal situation to clarify some of these points.

Related to this, in some (maybe all) versions of OpenFOAM I have noticed problems in Allrun scripts where the "application" has numbers in its name; for this reason, I hard-code the solids4Foam solver name in the run scripts. I suspect it should not be too difficult to use the "../../tools/run-openfoam.sh" script if you prefer.
Related to this, the "openfoam-remove-empty-dirs.sh" script needs a few minor updates to work with a solids4foam case. Let me know and I can make these changes.

perpendicular-flap/solid-solids4foam/system/controlDict
```diff
@@ -0,0 +1,62 @@
...
+    preCICE_Adapter
+    {
+        type preciceAdapterFunctionObject;
+        libs ("libpreciceAdapterFunctionObject.so");
+    }
+    interfaceForce
+    {
+        type solidForces;
+        historyPatch interface;
+    }
```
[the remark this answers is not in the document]
solids4foam: I have removed this now in 64bb3e7.

I was just using this when setting up the case to check the forces from both sides.

perpendicular-flap/solid-solids4foam/system/fvSchemes
```diff
@@ -0,0 +1,62 @@
...
+{
+    //default            Euler;
+    default            backward;
+}
+
+ddtSchemes
+{
+    //default            Euler;
+    default            backward;
+}
```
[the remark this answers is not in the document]
solids4foam: Agreed. Fixed in bfc935c.

perpendicular-flap/solid-solids4foam/system/controlDict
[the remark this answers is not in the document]
MakisH: > Related to this, the "openfoam-remove-empty-dirs.sh" script needs a few minor updates to work with a solids4foam case. Let me know and I can make these changes.

Feel free to directly contribute to that script. It is anyway very crude at the moment.

> From your nice documentation website, I have become motivated to set up a basic solids4foam page; that will be the ideal situation to clarify some of these points.

:smile: :partying_face: 

> I typically refer to the overall project (as opposed to the solver executable) as "solids4foam" (all lowercase).

I will use this convention everywhere, then. Unless there is a different policy on the future website.

perpendicular-flap/solid-solids4foam/0/D
```diff
@@ -0,0 +1,42 @@
...
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       volVectorField;
+    location    "0";
+    object      D;
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
```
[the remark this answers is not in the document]
solids4foam: Removed in 9ff832b.

perpendicular-flap/solid-solids4foam/system/fvSolution
```diff
@@ -0,0 +1,47 @@
...
+    {
+        //D   0.999;
+    }
+
+    // Under-relaxating the D field by 0.1-0.9 can improve convergence in some
+    // cases
+    fields
+    {
+        //D      0.9;
+    }
```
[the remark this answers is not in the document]
solids4foam: Ha, fair point. And understood.

perpendicular-flap/solid-solids4foam/0/D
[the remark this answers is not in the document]
solids4foam: OK, fixed in ac77e77.

Thanks, good to know more about gitattributes.

perpendicular-flap/solid-solids4foam/constant/mechanicalProperties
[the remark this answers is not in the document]
solids4foam: Fixed in 952d86f.

I agree that it would be better to describe this in the solids4foam documentation. When that is done then, I can come back here and update this.

perpendicular-flap/solid-solids4foam/system/controlDict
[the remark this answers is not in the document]
solids4foam: I have updated tools/run-openfoam.sh in 4dccace. The "if" lines are quite long now, but I am always scared when line wrapping bash commands.

Let me see if I can update the "run.sh" in perpendicular-flap/solid-solids4foam to be more similar to perpendicular-flap/fluid-openfoam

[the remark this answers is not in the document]
solids4foam: OK, I have now updated run.sh in a85fb9a to be consistent with the run.sh in fluid-openfoam.

perpendicular-flap/solid-solids4foam/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Currently, the case has only been checked with OpenFOAM-v2012
+if [[ "${WM_PROJECT}" != "OpenFOAM" || "${WM_PROJECT_VERSION}" != "v2012" ]]
+then
+    echo; echo "This case currently only runs in OpenFOAM-v2012"; echo
+    exit 0
+fi
```
MakisH, line 8: I am still not confident with this one, let's keep it open for now. I prefer to just document and link to the solids4foam documentation regarding compatibility. Otherwise, this will confuse users that will eventually have a compatible solids4foam, but older tutorials (that may need no updates).

tools/openfoam-remove-empty-dirs.sh
```diff
@@ -6,15 +6,15 @@ openfoam_remove_empty_dirs() {
 		echo "Looking for any time directories without results (e.g. stray functionObjectProperties files, see openfoam-adapter issue #26 on GitHub)..."
 
 		for f in [0-9]* [0-9]*.[0-9]*; do
-			if ! [ -f "${f}/U" ] && ! [ -f "${f}/T" ] && ! [ -f "${f}/U.gz" ] && ! [ -f "${f}/T.gz" ]; then
+                        if ! [ -f "${f}/U" ] && ! [ -f "${f}/T" ] && ! [ -f "${f}/U.gz" ] && ! [ -f "${f}/T.gz" ] && ! [ -f "${f}/D" ] && ! [ -f "${f}/pointD" ] && ! [ -f "${f}/DD" ] && ! [ -f "${f}/pointDD" ] && ! [ -f "${f}/D.gz" ] && ! [ -f "${f}/pointD.gz" ] && ! [ -f "${f}/DD.gz" ] && ! [ -f "${f}/pointDD.gz" ]; then
```
MakisH, line 9: Is there a situation where one will only have one of the `D`, `pointD`, `DD`, `pointDD`, `D.gz`, ...? Wouldn't it be enough to only include `D` and `D.gz`?

perpendicular-flap/solid-solids4foam/run.sh
solids4foam: Sure, understood.

If you agree, I can remove these lines.

tools/openfoam-remove-empty-dirs.sh
solids4foam: Unfortunately, some solid models in solids4foam use DD (increment of displacement) and do not use D; also, there is a solid model which uses pointD and does not use D or DD.

perpendicular-flap/README.md
```diff
@@ -37,6 +37,10 @@ Solid participant:
 
 * DUNE. For more information, have a look at the [experimental DUNE adapter](https://github.com/precice/dune-adapter) and send us your feedback.
 
+* OpenFOAM (solidDisplacementFoam). For more information, have a look at the [OpenFOAM plateHole tutorial](https://doc.cfd.direct/openfoam/user-guide-v6/platehole). The solidDisplacementFoam solver only supports linear geometry. For general solid mechanics procedures in OpenFOAM, see solids4foam.
```
MakisH, line 40: Since we are defaulting to the OpenCFD variant, let's also point to the respective documentation page:

```suggestion
* OpenFOAM (solidDisplacementFoam). For more information, have a look at the [OpenFOAM plateHole tutorial](https://www.openfoam.com/documentation/tutorial-guide/5-stress-analysis/5.1-stress-analysis-of-a-plate-with-a-hole). The solidDisplacementFoam solver only supports linear geometry. For general solid mechanics procedures in OpenFOAM, see solids4foam.
```

perpendicular-flap/solid-openfoam/constant/mechanicalProperties
```diff
@@ -0,0 +1,31 @@
...
+E
+{
+    type        uniform;
+    value       4e+6;
+}
+
+planeStress     no;
+
+
+// ************************************************************************* //
```
MakisH, line 31:
```suggestion
```

perpendicular-flap/solid-openfoam/run.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Currently, the case has only been checked with OpenFOAM-v2012
+if [[ "${WM_PROJECT}" != "OpenFOAM" || "${WM_PROJECT_VERSION}" != "v2012" ]]
+then
+    echo; echo "This case currently only runs in OpenFOAM-v2012"; echo
+    exit 0
+fi
+
```
MakisH, line 9: For the `solid-openfoam`, I guess we have no reason to restrict it to v2012:

```suggestion
```

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/darwin64ClangDPInt32Opt/Users/philipc/OpenFOAM/OpenFOAM-v2012/applications/solvers/stressAnalysis/solidDisplacementFoam/tractionDisplacement/tractionDisplacementFvPatchVectorField.C.dep
```diff
@@ -0,0 +1,578 @@
+$(OBJECTS_DIR)//Users/philipc/OpenFOAM/OpenFOAM-v2012/applications/solvers/stressAnalysis/solidDisplacementFoam/tractionDisplacement/tractionDisplacementFvPatchVectorField.C.dep: \
```
MakisH, line 1: I guess is automatically produced by WMake, so better remove it.

Same for the corresponding `.o` file.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/darwin64ClangDPInt32Opt/options
```diff
@@ -0,0 +1,30 @@
+# 1 "Make/options"
```
MakisH, line 1: Platform-specific, remove

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/darwin64ClangDPInt32Opt/solidDisplacementFoamForceFvPatchVectorField.C.dep
```diff
@@ -0,0 +1,579 @@
+$(OBJECTS_DIR)/solidDisplacementFoamForceFvPatchVectorField.C.dep: \
```
MakisH, line 1: Remove file (and the respective `.o` and platform-specific files).

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/lnInclude/solidDisplacementFoamForceFvPatchVectorField.C
```diff
@@ -0,0 +1 @@
+../solidDisplacementFoamForceFvPatchVectorField.C
```
MakisH, line 1: Remove (same for `.H`)

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.H
```diff
@@ -0,0 +1,191 @@
...
+    This is similar to tractionDisplacement except the absolute force per face
+    is specified.
+
+    Note: this applies the force per face, not to the entire patch.
+
+    Obviously, for a uniform force field, the total force applied to the patch
+    as the mesh is refined as the force per face stays constant.
+
+    The primary motivation for this condition is for use FSI cases using the
+    preCICE coupling library.
```
MakisH, line 41:
```suggestion
    The primary motivation for this condition is for FSI cases using 
    solidDisplacementFoam and the preCICE coupling library.
```

perpendicular-flap/solid-openfoam/system/fvSchemes
```diff
@@ -0,0 +1,58 @@
...
+{
+    default         linear;
+}
+
+snGradSchemes
+{
+    default         none;
+}
+
+// ************************************************************************* //
```
MakisH, line 58:
```suggestion
FoamFile
{
    version     2.0;
    format      ascii;
    class       dictionary;
    location    "system";
    object      fvSchemes;
}

d2dt2Schemes
{
    default         Euler;
}

ddtSchemes
{
    default         Euler;
}

gradSchemes
{
    default         leastSquares;
    grad(D)         leastSquares;
    grad(T)         leastSquares;
}

divSchemes
{
    default         none;
    div(sigmaD)     Gauss linear;
}

laplacianSchemes
{
    default         none;
    laplacian(DD,D) Gauss linear corrected;
    laplacian(DT,T) Gauss linear corrected;
}

interpolationSchemes
{
    default         linear;
}

snGradSchemes
{
    default         none;
}
```

perpendicular-flap/solid-openfoam/system/fvSolution
```diff
@@ -0,0 +1,38 @@
...
+
+stressAnalysis
+{
+    compactNormalStress yes;
+    nCorrectors     1000;
+    D               1e-06;
+}
+
+
+// ************************************************************************* //
```
MakisH, line 38:
```suggestion
FoamFile
{
    version     2.0;
    format      ascii;
    class       dictionary;
    location    "system";
    object      fvSolution;
}

solvers
{
    "(D|T)"
    {
        solver          GAMG;
        tolerance       1e-06;
        relTol          0.9;
        smoother        GaussSeidel;
        nCellsInCoarsestLevel 20;
    }
}

stressAnalysis
{
    compactNormalStress yes;
    nCorrectors     1000;
    D               1e-06;
}

```

perpendicular-flap/solid-openfoam/system/preciceDict
```diff
@@ -0,0 +1,50 @@
...
+{
+    solverType solid;
+
+    // Name of cell displacement field
+    nameCellDisplacement D;
+
+    // The solidDisplacementFoam does not have a point displacement field so we
+    // we use the special name "unused", which tells the adapter that it is not
+    // used
+    namePointDisplacement unused;
```
MakisH, line 46: @DavidSCN FYI, maybe you see a better solution.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/options
```diff
@@ -0,0 +1,25 @@
+ifeq ($(WM_PROJECT), foam)
+    VER := $(shell expr `echo $(WM_PROJECT_VERSION)` \>= 4.1)
+    ifeq ($(VER), 1)
+        VERSION_SPECIFIC_INC = -DFOAMEXTEND=41
+    else
+        VERSION_SPECIFIC_INC = -DFOAMEXTEND=40
+    endif
+else
```
MakisH, line 8: We anyway don't support foam-extend in the adapter, and I don't see `FOAMEXTEND` being used anywhere in the code.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.C
```diff
@@ -0,0 +1,282 @@
...
+    OpenFOAM is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+    for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with OpenFOAM; if not, write to the Free Software Foundation,
+    Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+
+\*---------------------------------------------------------------------------*/
```
MakisH, line 25: Duplication with `.H`, may be confusing.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.H
```diff
@@ -0,0 +1,191 @@
+/*---------------------------------------------------------------------------*\
+  =========                 |
+  \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox
+   \\    /   O peration     |
+    \\  /    A nd           | Copyright held by original author
+     \\/     M anipulation  |
+-------------------------------------------------------------------------------
+License
+    This file is part of OpenFOAM.
```
MakisH, line 9: Is it, really? If yes, which?
I see that the author is you, and the mesage that this comes from solids4foam, so I am not sure if we have to carry over the license in this form (we should of course do, if that's the right way).

Similar situation in another tutorial: https://github.com/precice/tutorials/blob/master/partitioned-heat-conduction/openfoam-solver/heatTransfer.C

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/options
```diff
@@ -0,0 +1,25 @@
...
+
+EXE_INC = \
+    $(VERSION_SPECIFIC_INC) \
+    -I$(FOAM_SOLVERS)/stressAnalysis/solidDisplacementFoam/tractionDisplacement \
+    -I$(LIB_SRC)/finiteVolume/lnInclude \
+    -I$(LIB_SRC)/meshTools/lnInclude
+
+LIB_LIBS = \
+    -lfiniteVolume \
+    -lmeshTools
```
MakisH, line 25: g++ 11.2 on Linux seems to not like something here:

```
ake/options:23:12: warning: backslash-newline at end of file
   23 | LIB_LIBS = \
      |             
    ln: ./lnInclude
```

Adding an empty line at the end seems to help.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.C
```diff
@@ -0,0 +1,282 @@
...
+        (
+            "solidDisplacementFoamForceFvPatchVectorField::solidDisplacementFoamForceFvPatchVectorField"
+        )   << "Only force or forceField can be "
+            << "specified, not both!"
+            << abort(FatalError);
+    }
+    else if (dict.found("forceField"))
+    {
+        Info<< "    force is specified as a field" << endl;
+        forceFieldPtr_.set
```
MakisH, line 105: A deprecation warning:

```
g++ -std=c++14 -m64 -pthread -DOPENFOAM=2206 -DWM_DP -DWM_LABEL_SIZE=32 -Wall -Wextra -Wold-style-cast -Wnon-virtual-dtor -Wno-unused-parameter -Wno-invalid-offsetof -Wno-attributes -Wno-unknown-pragmas -O3  -DNoRepository -ftemplate-depth-100  -DOPENFOAMESIORFOUNDATION -DOPENFOAMESI -I/usr/lib/openfoam/openfoam2206/applications/solvers/stressAnalysis/solidDisplacementFoam/tractionDisplacement -I/usr/lib/openfoam/openfoam2206/src/finiteVolume/lnInclude -I/usr/lib/openfoam/openfoam2206/src/meshTools/lnInclude -iquote. -IlnInclude -I/usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude -I/usr/lib/openfoam/openfoam2206/src/OSspecific/POSIX/lnInclude   -fPIC -c solidDisplacementFoamForceFvPatchVectorField.C -o Make/linux64GccDPInt32Opt/solidDisplacementFoamForceFvPatchVectorField.o
solidDisplacementFoamForceFvPatchVectorField.C: In constructor ‘Foam::solidDisplacementFoamForceFvPatchVectorField::solidDisplacementFoamForceFvPatchVectorField(const Foam::fvPatch&, const Foam::DimensionedField<Foam::Vector<double>, Foam::volMesh>&, const Foam::dictionary&)’:
solidDisplacementFoamForceFvPatchVectorField.C:106:9: warning: ‘void Foam::autoPtr<T>::set(T*) [with T = Foam::GeometricField<Foam::Vector<double>, Foam::fvPatchField, Foam::volMesh>]’ is deprecated: Since 2022-01; use "reset() - same behaviour" [-Wdeprecated-declarations]
  105 |         forceFieldPtr_.set
      |         ~~~~~~~~~~~~~~~~~~
  106 |         (
      |         ^
  107 |             new volVectorField
      |             ~~~~~~~~~~~~~~~~~~
  108 |             (
      |             ~
  109 |                 IOobject
      |                 ~~~~~~~~
  110 |                 (
      |                 ~
  111 |                     word(dict.lookup("forceField")),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  112 |                     patch().boundaryMesh().mesh().time().timeName(),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  113 |                     patch().boundaryMesh().mesh(),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  114 |                     IOobject::MUST_READ,
      |                     ~~~~~~~~~~~~~~~~~~~~
  115 |                     IOobject::AUTO_WRITE
      |                     ~~~~~~~~~~~~~~~~~~~~
  116 |                 ),
      |                 ~~
  117 |                 patch().boundaryMesh().mesh()
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  118 |             )
      |             ~
  119 |         );
      |         ~
In file included from /usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude/List.H:46,
                 from /usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude/wordList.H:50,
                 from /usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude/patchIdentifier.H:42,
                 from /usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude/polyPatch.H:45,
                 from /usr/lib/openfoam/openfoam2206/src/finiteVolume/lnInclude/fvPatch.H:42,
                 from /usr/lib/openfoam/openfoam2206/src/finiteVolume/lnInclude/fvPatchField.H:50,
                 from /usr/lib/openfoam/openfoam2206/src/finiteVolume/lnInclude/fvPatchFields.H:31,
                 from /usr/lib/openfoam/openfoam2206/applications/solvers/stressAnalysis/solidDisplacementFoam/tractionDisplacement/tractionDisplacementFvPatchVectorField.H:41,
                 from solidDisplacementFoamForceFvPatchVectorField.H:57,
                 from solidDisplacementFoamForceFvPatchVectorField.C:27:
/usr/lib/openfoam/openfoam2206/src/OpenFOAM/lnInclude/autoPtr.H:289:14: note: declared here
  289 |         void set(T* p) noexcept { reset(p); }
      |              ^~~
```

We can replace with the new solution, if this is actually working with v1812 and newer (or at least something not too much newer than that).

perpendicular-flap/solid-openfoam/system/controlDict
```diff
@@ -0,0 +1,50 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      controlDict;
+}
+
+libs            ("libsolidDisplacementFoamForce.so");
```
MakisH, line 10: When building with v2012 (and with v2206), I get:

```
Create time

--> FOAM Warning : 
    Could not load "libsolidDisplacementFoamForce.so"
/home/makish/OpenFOAM/makish-v2012/platforms/linux64GccDPInt32Opt/lib/libsolidDisplacementFoamForce.so: undefined symbol: _ZTIN4Foam38tractionDisplacementFvPatchVectorFieldE
Create mesh for time = 0
```

any clue why? `ldd /home/makish/OpenFOAM/makish-v2012/platforms/linux64GccDPInt32Opt/lib/libsolidDisplacementFoamForce.so` does not report any unresolved symbols:

```
~/tutorials/perpendicular-flap/solid-openfoam [solids4foam/master]$ ldd /home/makish/OpenFOAM/makish-v2012/platforms/linux64GccDPInt32Opt/lib/libsolidDisplacementFoamForce.so
	linux-vdso.so.1 (0x00007ffce2dd0000)
	libfiniteVolume.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/libfiniteVolume.so (0x00007f1e7a682000)
	libmeshTools.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/libmeshTools.so (0x00007f1e7a157000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f1e79eed000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1e79e06000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1e79de4000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1e79bbc000)
	libOpenFOAM.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/libOpenFOAM.so (0x00007f1e793c8000)
	libPstream.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/sys-openmpi/libPstream.so (0x00007f1e793ae000)
	libfileFormats.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/libfileFormats.so (0x00007f1e79273000)
	libsurfMesh.so => /usr/lib/openfoam/openfoam2012/platforms/linux64GccDPInt32Opt/lib/libsurfMesh.so (0x00007f1e790c7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1e7b86d000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f1e790a9000)
	libmpi.so.40 => /lib/x86_64-linux-gnu/libmpi.so.40 (0x00007f1e78f72000)
	libopen-rte.so.40 => /lib/x86_64-linux-gnu/libopen-rte.so.40 (0x00007f1e78eb5000)
	libopen-pal.so.40 => /lib/x86_64-linux-gnu/libopen-pal.so.40 (0x00007f1e78e02000)
	libhwloc.so.15 => /lib/x86_64-linux-gnu/libhwloc.so.15 (0x00007f1e78da6000)
	libevent_core-2.1.so.7 => /lib/x86_64-linux-gnu/libevent_core-2.1.so.7 (0x00007f1e78d6f000)
	libevent_pthreads-2.1.so.7 => /lib/x86_64-linux-gnu/libevent_pthreads-2.1.so.7 (0x00007f1e78d6a000)
	libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x00007f1e78d40000)
```

MakisH (commented):
Thanks a lot for adding a `solid-openfoam` case as well! I had some trouble getting the case to run, even though it compiled successfully with v2012 and v2206. Looks also like you accidentally pushed some build artifacts.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/darwin64ClangDPInt32Opt/options
solids4foam: Resolved in a6b8a09.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.C
solids4foam: Resolved in 7b9504d.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.H
solids4foam: Resolved in 7b9504d.

As you have seen, I have (probably incorrectly) not given much attention to the header license information in classes in solids4foam; however, I fully agree that it is confusing if they say they are (incorrectly) part of some version of OpenFOAM. I will fix this for the nextRelease in solids4foam.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/solidDisplacementFoamForceFvPatchVectorField.C
solids4foam: Resolved in bc18550.

perpendicular-flap/solid-openfoam/system/controlDict
solids4foam: Hmnn... so it is missing the definition of the tractionDisplacement class. Interestingly, this works for me when using clang. I will also try to use GCC to reproduce the error.

The new solidDisplacementFoamForce boundary condition derives from the tractionDisplacement boundary condition, where the tractionDisplacement condition is compiled directly with the solidDisplacementFoam solver. So the definition of tractionDisplacement should be available when calling the solidDisplacementFoam solver. Do you get this error from running solidDisplacementFoam?
I guess this error is to be expected when using other OpenFOAM solvers/utilities, e.g. blockMesh, but that's OK to ignore (unless you would rather not adopt that approach).

Initially, I tried also to compile tractionDisplacement with the solidDisplacementFoamForce boundary condition; however, in that case, solidDisplacementFoam complains that two definitions of tractionDisplacement are available.

Ideally, tractionDisplacement would be available within a library; however, currently, it builds with the solidDisplacementFoam solver.

So, what is the best solution... how about the following?

- I could copy code from the tractionDisplacement base class directly into the new solidDisplacementFoamForce class, such that solidDisplacementFoamForce no longer needs to compile from tractionDisplacement at all. In that way, it will be a stand-alone boundary condition, and we don't need to worry about the tractionDisplacement definition is available. The downside is that there will be some code duplication; however, this will probably only be 10-20 lines.

solids4foam: Resolved in f9da0cc.

You will see that the changes are not so major.

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/options
MakisH: Yes, please open an issue! :grimacing: I will then fix then and add a CI step that builds with Clang.

solids4foam: Sure. I just added https://github.com/precice/openfoam-adapter/issues/237

perpendicular-flap/solid-openfoam/system/fvSolution
```diff
@@ -0,0 +1,38 @@
...
+        relTol          0.9;
+        smoother        GaussSeidel;
+        nCellsInCoarsestLevel 20;
+    }
+}
+
+stressAnalysis
+{
+    compactNormalStress yes;
+    nCorrectors     1000;
```
MakisH, line 33: Do we really need such a high value here? I feel that the case is unreasonably slow because of this (still checking).

solids4foam: In short, yes.

However, this is the maximum number of correctors, so typically, this value is not reached, assuming the D outer loop tolerance is reached (given by `D               1e-06;`).

This is one of the disadvantages of the "segregated" approach for solids: in some cases, it can require large amounts of outer iterations and be slow. This has motivated the development of block-coupled solid solvers in solids4foam.

MakisH: Observations:
- With 1000 and 100, I get oscillations
- With 10, I get a steady-state

perpendicular-flap/solid-openfoam/solidDisplacementFoamForce/Make/options
MakisH: So, my suggestion here would be to remove these lines, to not confuse users.

solids4foam: Sure. Resolved in 1d0d863.

perpendicular-flap/solid-openfoam/system/fvSolution
solids4foam: OK, sure.

Shall I add the comment? Of alternatively, feel free to add it yourself.

solids4foam: Alternatively, the following changes could be made:

- reduce the solid mesh resolution in the blockMeshDict from `(6 15 1)` to `(3 15 1)`
- loosen the outer momentum tolerance in fvSolution from `D 1e-6` to `D 1e-5`
- [probably does not make much difference] change from GAMG to DIC solver in fvSolution:
```
solvers
{
    "(D|T)"
    {
        solver          PCG;
        preconditioner  DIC;
        tolerance       1e-06;
        relTol          0.1;
}
```
Then, the case runs in 66 s on my computer and produce reasonable oscillations.

MakisH: Here is what I would add:

```suggestion
    // Note: The accuracy of the solution can be significantly improved
    // by increasing this number of iterations, but this will impact the runtime.
    nCorrectors     30;
```

Thanks for trying different approaches. I guess this is a big box to open at some point (could even make a nice training talk at the OpenFOAM Workshop). Let's go for the easiest right now.

MakisH:
So, this looks more or less ready:

- Review is mostly ready to my perspective
- I was able to run the `solid-openfoam` case (no solids4foam)
- I was able to build solids4foam with OpenFOAM v2012 (as documented, but I still have issues running the case)

So, still todo:
- Run the `solid-solids4foam` case
- Add a PR description

I also triggered the CI (basic script/docs linting checks) and everything looks good there.

perpendicular-flap/solid-solids4foam/run.sh
solids4foam: Removed in c6f2b0b.

MakisH:
I currently have a very strange issue running the `solid-solids4foam` case, which I am not sure where it originates from. However, it looks related to the fact that both preCICE and solids4foam use Eigen.

Logs: https://gist.github.com/MakisH/26342b40980ea0c3260ef90110a15295

This run was with:
- Ubuntu 22.04, which includes Eigen 3.4.0.
- preCICE v2.4.0, built from source in Debug mode
- OpenFOAM v2012
- The OpenFOAM adapter state from https://github.com/precice/openfoam-adapter/pull/236 (commit a591b)
- solids4foam `nextRelease` branch (commit 50881), built with its own copy of Eigen 3.3.7.

Other tutorial cases work normally (e.g., the `solid-openfoam`).

I know that one can ask solids4foam to use the global Eigen installation. I did so by setting `EIGEN_DIR="/usr/include/eigen3"` and rebuilding solids4foam (after executing `Allwclean`). Looks like `Allwmake` discovered my installation and just created a symlink to that, so both preCICE and solids4foam should be using the same version. But even with that, I get the same error.

@solids4foam did you have any such issues?

@fsimonis any clue what could be wrong here, if any, in preCICE?

fsimonis:
@MakisH This definitely looks like conflicting versions. The callback of the tag uses an instance of `Eigen::MatrixXd`, so this all makes sense.

Can you somehow see which version of Eigen OpenFOAM is actually using (build logs, compiler invocations or similar maybe?)

Alternatively, you could build preCICE with the Eigen version used by OpenFOAM. This sounds like a shorter compilation time. :hot_face:

solids4foam:
> @MakisH This definitely looks like conflicting versions. The callback of the tag uses an instance of `Eigen::MatrixXd`, so this all makes sense.
> 
> Can you somehow see which version of Eigen OpenFOAM is actually using (build logs, compiler invocations or similar maybe?)
> 
> Alternatively, you could build preCICE with the Eigen version used by OpenFOAM. This sounds like a shorter compilation time. 🥵

As Eigen is a header-only library, I am surprised that it is causing a conflict since it is only looked up during compilation, not at run-time (e.g., as like a dynamic library).
Can you track down the actual lines in the Adapter and/or preCICE that cause the problem? Maybe, the `-DFULLDEBUG -g -O0` compilation flags would help for the adapter.

By the way, can you check if solids4foam works on its own, e.g. run `./Allrun` in `solids4foam-release/tutorials/solids/linearElasticity/plateHole/plateHoleTotalDisp`, just to rule out solids4foam-specific issues.

solids4foam:
> > @MakisH This definitely looks like conflicting versions. The callback of the tag uses an instance of `Eigen::MatrixXd`, so this all makes sense.
> > Can you somehow see which version of Eigen OpenFOAM is actually using (build logs, compiler invocations or similar maybe?)
> > Alternatively, you could build preCICE with the Eigen version used by OpenFOAM. This sounds like a shorter compilation time. 🥵
> 
> As Eigen is a header-only library, I am surprised that it is causing a conflict since it is only looked up during compilation, not at run-time (e.g., as like a dynamic library). Can you track down the actual lines in the Adapter and/or preCICE that cause the problem? Maybe, the `-DFULLDEBUG -g -O0` compilation flags would help for the adapter.
> 
> By the way, can you check if solids4foam works on its own, e.g. run `./Allrun` in `solids4foam-release/tutorials/solids/linearElasticity/plateHole/plateHoleTotalDisp`, just to rule out solids4foam-specific issues.

FYI, I have tried the following, and it works for me:

- Ubuntu 20.04.4 LTS
- OpenFOAM-v2012 from `apt-get`
- preCICE binary `libprecice2_2.4.0_focal.deb`
- solids4foam, nextRelease branch (commit 4e3f64b0): I did *not* set `EIGEN_DIR` and so solids4foam downloaded its own copy of Eigen (eigen-3.3.7)

I have also tried (and it works):

- macOS 11.4
- OpenFOAM-v2012 built from source using the patches from https://github.com/mrklein/openfoam-os-x and the clang compiler (Apple clang version 12.0.5)
- preCICE binaries from homebrew (thank you, very convenient!)
- solids4foam, nextRelease branch (commit 50188d0e - the more recent): once again, I did *not* set `EIGEN_DIR` and so solids4foam downloaded its own copy of Eigen (eigen-3.3.7)

I wonder if the debug mode of preCICE catches something that gets missed in the optimised version.

MakisH:
The Debug vs Release mode sounds indeed like something relevant. Building in Release mode, preCICE is initialized normally, but then [breaks the next time it uses Eigen](https://gist.github.com/MakisH/26342b40980ea0c3260ef90110a15295#file-with-precice-in-release-mode-without-petsc). This output is without PETSc, as the error handling seems to obfuscate issues, which also makes the RBF mapping use Eigen. But even [building in Release mode with PETSc gives again a segfault](https://gist.github.com/MakisH/26342b40980ea0c3260ef90110a15295#file-with-precice-in-release-mode-with-petsc).

Same situation when solids4foam is built with its own Eigen copy. solids4foam by itself runs fine.

> FYI, I have tried the following, and it works for me:

I will try to reproduce the first environment here, but to me this sounds like an actual issue that others will eventually also face.

MakisH:
Update: the case does run in the [preCICE Demo VM](https://precice.org/installation-vm.html), with OpenFOAM v2012. It completes without any issues and I get reasonable results.

I would suggest that:
- I open an issue summarizing our discussion
- Maybe, for visibility, @solids4foam creates a similar issue at the solids4foam repository (I this the most probably is that the issue originates from there, more in the issue I will open)
- We add a link to issue to the documentation (this PR)
- We merge this.

perpendicular-flap/README.md
```diff
@@ -37,6 +37,10 @@ Solid participant:
 
 * DUNE. For more information, have a look at the [experimental DUNE adapter](https://github.com/precice/dune-adapter) and send us your feedback.
 
+* OpenFOAM (solidDisplacementFoam). For more information, have a look at the [OpenFOAM plateHole tutorial](https://www.openfoam.com/documentation/tutorial-guide/5-stress-analysis/5.1-stress-analysis-of-a-plate-with-a-hole). The solidDisplacementFoam solver only supports linear geometry. For general solid mechanics procedures in OpenFOAM, see solids4foam.
+
+* solids4foam. Like for CalculuX, the geometrically linear solver is used by default. For more information, see the [solids4foam documentation](https://bitbucket.org/philip_cardiff/solids4foam-release/src/master/documentation/overview.md) This case currently (August 2022) only works with the `nextRelease` branch of solids4foam, which is only compatible with up to OpenFOAM v2012 and OpenFOAM 9 (as well as foam-extend, with which the OpenFOAM-preCICE adapter is not compatible).
```
MakisH, line 42:
```suggestion
* solids4foam. Like for CalculuX, the geometrically linear solver is used by default. For more information, see the [solids4foam documentation](https://bitbucket.org/philip_cardiff/solids4foam-release/src/master/documentation/overview.md) This case currently (August 2022) only works with the `nextRelease` branch of solids4foam, which is only compatible with up to OpenFOAM v2012 and OpenFOAM 9 (as well as foam-extend, with which the OpenFOAM-preCICE adapter is not compatible). In case you cannot run the case, have a look and comment at this [solids4foam-preCICE compatibility issue](https://github.com/precice/openfoam-adapter/issues/238).
```

fsimonis:
It is almost definitely an ODR issue with various versions of Eigen.

> As Eigen is a header-only library, I am surprised that it is causing a conflict since it is only looked up during compilation, not at run-time (e.g., as like a dynamic library).

The preCICE shared library exports all symbols, which is bad. We are working to change in the future. This means that the linker will treat both versions of Eigen as equal. This consequently leads to the observable issues with the destructor.

OpenFOAM downloads Eigen 3.3.7, while preCICE uses the system installation of Eigen.
Ubuntu 20.04 ships Eigen 3.3.7 and Ubuntu 22.04 ships Eigen 3.4.0.
Hence, Ubuntu 20.04 _works_ as both versions are identical.
The versions differ on Ubuntu 22.04, explaining the violation of the ODR and the crash.

Note that we most likely violate the ODR even with identical versions of Eigen, as we conditionally enable `EIGEN_INITIALIZE_MATRICES_BY_NAN` in Debug builds. Which of the compiled versions the linker finally picks, most likely doesn't dramatically influence the behaviour of Eigen, hence no crash.

perpendicular-flap/README.md
```diff
@@ -37,6 +37,10 @@ Solid participant:
...
 
+* OpenFOAM (solidDisplacementFoam). For more information, have a look at the [OpenFOAM plateHole tutorial](https://www.openfoam.com/documentation/tutorial-guide/5-stress-analysis/5.1-stress-analysis-of-a-plate-with-a-hole). The solidDisplacementFoam solver only supports linear geometry. For general solid mechanics procedures in OpenFOAM, see solids4foam.
+
+* solids4foam. Like for CalculuX, the geometrically linear solver is used by default. For more information, see the [solids4foam documentation](https://bitbucket.org/philip_cardiff/solids4foam-release/src/master/documentation/overview.md) This case currently (August 2022) only works with the `nextRelease` branch of solids4foam, which is only compatible with up to OpenFOAM v2012 and OpenFOAM 9 (as well as foam-extend, with which the OpenFOAM-preCICE adapter is not compatible). Note that, since both solids4foam and preCICE rely on Eigen, and due to some [solids4foam-preCICE compatibility issue](https://github.com/precice/openfoam-adapter/issues/238), you need to build preCICE and solids4foam with the same Eigen version.
```
MakisH, line 42:
```suggestion
* solids4foam. Like for CalculuX, the geometrically linear solver is used by default. For more information, see the [solids4foam documentation](https://bitbucket.org/philip_cardiff/solids4foam-release/src/master/documentation/overview.md) This case currently (August 2022) only works with the `nextRelease` branch of solids4foam, which is only compatible with up to OpenFOAM v2012 and OpenFOAM 9 (as well as foam-extend, with which the OpenFOAM-preCICE adapter is not compatible), as well as the `develop` branch of the OpenFOAM-preCICE adapter. Note that, since both solids4foam and preCICE rely on Eigen, and due to some [solids4foam-preCICE compatibility issue](https://github.com/precice/openfoam-adapter/issues/238), you need to build preCICE and solids4foam with the same Eigen version.
```

MakisH (approved):
Finally, I don't have any more comments to add, rather than that this was a really nice contribution and review! :smile: Thanks a lot!

raynoldtan:
We have recently tried to run the perpendicular flap case using percice 2.4 with solids4Foam NextRelease branch and the develop branch of the openFoam Adapter. We have encounter similar issue. We have used gdb to get the call stack. Attached is the screenshot of the error message.
Screenshot 2022-09-06 at 17 38 20
please advise how to fix.

fsimonis:
@raynoldtan Same issue. You need to tell solid4foam to use the system eigen. https://solids4foam.github.io/installation/installFromSource.html#eigen

This needs to point to the same version used by spack. Use `spack find --paths eigen` to see the path of the used version.

fsimonis:
@raynoldtan The EIGEN_DIR is expected to point to the include directory of Eigen. The Eigen repo follows an unusual directory structure. :wink:

So you need to append `/include` to the installation prefix.
```
export EIGEN_DIR=/home/raynoldtan/spack/opt/spack/linux-ubuntu20.04-cascadelake/gcc-9.4.0/eigen-3.4.0-4k2lkyslxqecxkljs4yh6t7rcbrsvh7d/include
```

raynoldtan:
@fsimonis 

Thanks for letting me know. I changed the export command to now append the /include into the path. 

export EIGEN_DIR=/home/raynoldtan/spack/opt/spack/linux-ubuntu20.04-cascadelake/gcc-9.4.0/eigen-3.4.0-4k2lkyslxqecxkljs4yh6t7rcbrsvh7d/include

And then I recompiled the solids4Foam. After that I go to ThirdParty folder to check,

Screenshot 2022-09-07 at 16 11 28


It seems correct now. Then I tried to rerun the perpendicular flap case again. But it still give the same issue as:

[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://petsc.org/release/faq/#valgrind
[0]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple MacOS to find memory corruption errors
[0]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run 
[0]PETSC ERROR: to get more information on the crash.
[0]PETSC ERROR: User provided function() at unknown file:0  
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 59.

I not sure if the version of precice used affect this? or the eigen version used will affect this?

Raynold

fsimonis:
@raynoldtan The path is still incorrect.

It has to look like this `ThirdParty/eigen3/Eigen/Core`, so only one `eigen3`.

I would manually remove the `eigen3` directory in `ThirdParty` and recompile from scratch.

raynoldtan:
@fsimonis 

I have removed the eigen3 folder. Then, I set the env variable using the following path:

export EIGEN_DIR=/home/raynoldtan/spack/opt/spack/linux-ubuntu20.04-cascadelake/gcc-9.4.0/eigen-3.4.0-4k2lkyslxqecxkljs4yh6t7rcbrsvh7d/include/eigen3

After that, I recompiled solids4Foam. Then I get the same directory path for eigen 3 as described by you above:

Screenshot 2022-09-08 at 10 44 11

~/solids4foam-release/ThirdParty/eigen3/Eigen

where the file Core is located.

Then I tried to rerun the case again. But I still get the same error as above.

fsimonis:
Ok, so there is something else going on.
Could you please open a new issue in this repository and describe your setup in additional detail?

raynoldtan:
Ok thanks for your help anyway. I have included the link below.

[https://github.com/precice/tutorials/issues/292]()

PR #290, opened 2022-08-12
Add gitignore for code and binaries, cleanup

MakisH:
We did not want to have a `.gitignore` in this repository, as the idea was that one could easily see what files are "dirty" and then use the cleaning scripts to clean up.

However, we also have now some cases that contain problem-specific solvers, i.e., code. This includes binaries, CMake files, etc, which are not cleaned up by the cleaning scripts, as one typically reuses them.

This PR:
- Adds an individual `.gitignore` file to each case directory that falls in the above category
- Cleans up an automatically generated mesh file for the `flow-over-heated-plate-two-meshes/solid-calculix` case.
- Removes the generation of VTU files from the `flow-over-heated-plate-two-meshes` tutorial.

@boris-martin please confirm that we don't need these files anymore.
@uekerman mostly FYI, please stop me if you think we should rather not add `.gitignore` files but rather clean also these auto-generated artifacts.


boris-martin (approved):
No reason to keep VTU nor mesh files indeed.

PR #291, opened 2022-08-24
Remove std::pow from constexpr

mennodeij:
using `std::pow` in a `constexpr` is not supported by Intel oneAPI C++ compiler:
```
rigid_body_solver.cpp(117): error: expression must have a constant value
    constexpr double inertia_moment = (1. / 12) * mass * (4 * std::pow(length, 2) + std::pow(height, 2));
                                      ^
/usr/include/c++/8/cmath(418): note: cannot call non-constexpr function "pow" (declared at line 140 of "/usr/include/bits/mathcalls.h")
        return pow(__type(__x), __type(__y));
               ^
```


DavidSCN (approved):
Indeed, seems to be a gcc extension, thanks!

PR #293, opened 2022-09-19
Tweak aste tutorial documentation

davidscn:
Follow-up to #244

PR #294, opened 2022-09-20
Update ASTE executable names

kursatyurt:
ASTE executable names have changed since tutorial files was added this PR updates the names of executables.


davidscn:
We already have a revision in progress https://github.com/precice/tutorials/pull/293

PR #295, opened 2022-09-27
refactoring partitioned-heat-conduction/nutils

gertjanvanzwieten:
Some suggested changes to the heat.py script to make it more efficient, more correct, and hopefully easier to read, while remaining backwards compatible with nutils 6 as well as forwards compatible with nutils 8. The commit messages explain the reasoning behind the individual modifications.

### DIFF
--- a/partitioned-heat-conduction/nutils/heat.py
+++ b/partitioned-heat-conduction/nutils/heat.py
@@ -16,6 +16,7 @@ def main(side='Dirichlet'):
     x_coupling = 1  # x coordinate of coupling interface
     degree = 1  # linear finite elements
     n = 10  # number of mesh vertices per dimension
+    timestep = 0.1
 
     if side == 'Dirichlet':
         x_grid = np.linspace(x_left, x_coupling, n)
@@ -96,21 +97,23 @@ def main(side='Dirichlet'):
 
     interface.initialize_data()
 
-    t = 0
+    t = 0.
+    istep = 0
 
     # initial condition
     sqr0 = domain.integral('(u - uexact)^2' @ ns, degree=degree * 2)
-    lhs0 = solver.optimize('lhs', sqr0, arguments=dict(t=t))
+    lhs = solver.optimize('lhs', sqr0, arguments=dict(t=t))
     bezier = domain.sample('bezier', degree * 2)
-    x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs0, t=t)
-    with treelog.add(treelog.DataLog()):
-        export.vtk(side + '-0', bezier.tri, x, Temperature=u, reference=uexact)
 
-    t += precice_dt
-    timestep = 0
-    dt = 0.1
+    while True:
 
-    while interface.is_coupling_ongoing():
+        # generate output
+        x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs, t=t)
+        with treelog.add(treelog.DataLog()):
+            export.vtk(side + "-" + str(istep), bezier.tri, x, Temperature=u, reference=uexact)
+
+        if not interface.is_coupling_ongoing():
+            break
 
         # read data from interface
         if interface.is_read_data_available():
@@ -118,13 +121,14 @@ def main(side='Dirichlet'):
 
         # save checkpoint
         if interface.is_action_required(precice.action_write_iteration_checkpoint()):
-            lhs_checkpoint = lhs0
-            t_checkpoint = t
-            timestep_checkpoint = timestep
+            checkpoint = lhs, t, istep
             interface.mark_action_fulfilled(precice.action_write_iteration_checkpoint())
 
-        # potentially adjust non-matching timestep sizes
-        dt = min(dt, precice_dt)
+        # prepare next timestep
+        lhs0 = lhs
+        istep += 1
+        dt = min(timestep, precice_dt)
+        t += dt
 
         # update (time-dependent) boundary condition
         cons = solver.optimize('lhs', sqr, droptol=1e-15, arguments=dict(t=t, readdata=readdata))
@@ -144,23 +148,10 @@ def main(side='Dirichlet'):
         # do the coupling
         precice_dt = interface.advance(dt)
 
-        # advance variables
-        t += dt
-        timestep += 1
-        lhs0 = lhs
-
         # read checkpoint if required
         if interface.is_action_required(precice.action_read_iteration_checkpoint()):
-            lhs0 = lhs_checkpoint
-            t = t_checkpoint
-            timestep = timestep_checkpoint
+            lhs, t, istep = checkpoint
             interface.mark_action_fulfilled(precice.action_read_iteration_checkpoint())
-        else:  # go to next timestep
-            bezier = domain.sample('bezier', degree * 2)
-            x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs, t=t)
-
-            with treelog.add(treelog.DataLog()):
-                export.vtk(side + "-" + str(timestep), bezier.tri, x, Temperature=u, reference=uexact)
 
     interface.finalize()
 
--- a/partitioned-heat-conduction/nutils/heat.py
+++ b/partitioned-heat-conduction/nutils/heat.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python3
 
 from nutils import cli, mesh, function, solver, export
+import functools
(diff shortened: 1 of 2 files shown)


IshaanDesai:
Thanks for the refactoring @gertjanvanzwieten! Is this ready for reviewing?

gertjanvanzwieten:
I noticed that autopep8 isn't happy, in part because of the double space indent on line 101 to avoid double indentation of 103-149, but I can make it compliant if you prefer. Otherwise it's ready for review!

gertjanvanzwieten:
Please be aware that Nutils may trigger [an issue](https://github.com/precice/precice/issues/1443) unrelated to this PR that we are currently in the process of fixing. The problem is exacerbated in the development version but exists in stable as well, unnoticed until now.

partitioned-heat-conduction/nutils/heat.py
```diff
@@ -1,173 +1,152 @@
...
-    ns.alpha = 3  # parameter of problem
-    ns.beta = 1.3  # parameter of problem
+    ns.alpha = alpha  # parameter of problem
+    ns.beta = beta  # parameter of problem
     ns.u = 'basis_n ?lhs_n'  # solution
     ns.dudt = 'basis_n (?lhs_n - ?lhs0_n) / ?dt'  # time derivative
     ns.flux = 'basis_n ?fluxdofs_n'  # heat flux
     ns.f = 'beta - 2 - 2 alpha'  # rhs
-    ns.uexact = '1 + x_0 x_0 + alpha x_1 x_1 + beta ?t'  # analytical solution
+    ns.uexact = '1 + (x_0 + 1)^2 + alpha x_1^2 + beta ?t'  # analytical solution
```
IshaanDesai, line 40: The analytical solution is changed here. I think it does not matter in the end, but is there a reason to do this?

partitioned-heat-conduction/nutils/heat.py
```diff
@@ -1,173 +1,152 @@
...
     # top and bottom boundary are non-coupling for both sides
-    sqr0 = domain.boundary['top'].integral('(u - 1 - x_0 x_0 - alpha - beta ?t)^2 d:x' @ ns, degree=degree * 2)
-    sqr0 += domain.boundary['bottom'].integral('(u - 1 - x_0 x_0 - beta ?t)^2 d:x' @ ns, degree=degree * 2)
-    if side == 'Dirichlet':  # left boundary is non-coupling
-        sqr0 += domain.boundary['left'].integral('(u - 1 - alpha x_1 x_1 - beta ?t)^2 d:x' @ ns, degree=degree * 2)
-    elif side == 'Neumann':  # right boundary is non-coupling
-        sqr0 += domain.boundary['right'].integral('(u - 1 - x_0 x_0 - alpha x_1 x_1 - beta ?t)^2 d:x' @ ns,
-                                                  degree=degree * 2)
+    sqr = domain.boundary['top,bottom,left' if side == 'Dirichlet'
+                      else 'top,bottom,right'].integral('(u - uexact)^2 d:x' @ ns, degree=degree * 2)
```
IshaanDesai, line 50: In the previous formulation the `top` and `bottom` boundary conditions were different than the `left` and `right` non-coupling boundaries. I understand that what you have done now makes it simpler, but just to confirm, is that the only reason?

partitioned-heat-conduction/nutils/heat.py
```diff
@@ -1,173 +1,152 @@
...
-        else:  # go to next timestep
-            bezier = domain.sample('bezier', degree * 2)
-            x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs, t=t)
-
-            with treelog.add(treelog.DataLog()):
-                export.vtk(side + "-" + str(timestep), bezier.tri, x, Temperature=u, reference=uexact)
+        checkpoint = lhs, t
+        interface.mark_action_fulfilled(precice.action_write_iteration_checkpoint())
+
+        while interface.is_coupling_ongoing() and not interface.is_action_required(precice.action_write_iteration_checkpoint()):
```
IshaanDesai, line 116: I personally understand the order of the time loop right now, but I am not sure if this change of ordering makes it easier for a new user to quickly see how the implicit coupling has been done. Having the checkpoint writing at the start necessitates the break statement above and also this statement with two conditions. I personally prefer to keep it like:

- read checkpoint
- read data
- solve system
- write data
- write checkpoint or check for convergence

@uekerman what do you think?

uekerman: Sth went wrong with the order of actions here. The checkpoint is now only written once before the time loop.

There are multiple options, but the classical order would be:

* write checkpoint if required
* read data
* solve
* write data
* advance
* read checkpoint if required

https://precice.org/couple-your-code-implicit-coupling.html

partitioned-heat-conduction/nutils/heat.py
gertjanvanzwieten: The boundary conditions should not have changed; the former terms simply had `x_0 = 0` substituted at the left boundary and `x_1 = 0` and `x_1 = 1` and the bottom and top boundaries. Unless I am overlooking something?

partitioned-heat-conduction/nutils/heat.py
gertjanvanzwieten: This modification compensates for shifting the domain, so together the analytical solution should not be different. The rational behind the shift is explained in the commit message of eec74b6.

partitioned-heat-conduction/nutils/heat.py
gertjanvanzwieten: The time loop is the outer loop, corresponding to Precice's time windows and checkpoints. The inner loop is for iterations and sub-timesteps. The rationale behind this structure is described in the commit message of aee4700 as being largely cosmetic, but I personally find it insightful to see checkpointing reflected in a loop rather than a read action that acts rather like a goto statement. Not criticizing the API, mind you, I just think there is something to be said for giving the script a clearly identifiable structure in its own right. That said, this is a Precice tutorial, and you should obviously use the structure that best serves that purpose. Let me know if you want me to drop the commit.

IshaanDesai: I now understand the point of having two loops and this does make the checkpointing more *explicit*, but I would say that as this is a preCICE tutorial, lets stick to the usual form of having one time loop which has the operations in the order that @uekerman has shown above.

partitioned-heat-conduction/nutils/heat.py
```diff
@@ -1,173 +1,147 @@
...
+    lhs = solver.optimize('lhs', sqr0, arguments=dict(t=t))
     bezier = domain.sample('bezier', degree * 2)
-    x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs0, t=t)
-    with treelog.add(treelog.DataLog()):
-        export.vtk(side + '-0', bezier.tri, x, Temperature=u, reference=uexact)
 
-    t += precice_dt
-    timestep = 0
-    dt = 0.1
+    while True:
```
IshaanDesai, line 98:
```suggestion
    while interface.is_coupling_ongoing()
```

partitioned-heat-conduction/nutils/heat.py
```diff
@@ -1,173 +1,147 @@
...
-    while interface.is_coupling_ongoing():
+        # generate output
+        x, u, uexact = bezier.eval(['x_i', 'u', 'uexact'] @ ns, lhs=lhs, t=t)
+        with treelog.add(treelog.DataLog()):
+            export.vtk(side + "-" + str(istep), bezier.tri, x, Temperature=u, reference=uexact)
 
-        # update (time-dependent) boundary condition
-        cons0 = solver.optimize('lhs', sqr0, droptol=1e-15, arguments=dict(t=t))
+        if not interface.is_coupling_ongoing():
+            break
```
IshaanDesai, line 106:
```suggestion

```

IshaanDesai (commented):
Thank you for the changes and also catching the incompatibility, it totally slipped my view too. Just a minor suggestion regarding the main time loop.

partitioned-heat-conduction/nutils/heat.py
gertjanvanzwieten: This would necessitate generating output for the initial condition before entering the while loop, as was the case prior to this commit. Personally I think a break is cleaner than duplicate code but if that is what you prefer then I'll revert it, let me know.

IshaanDesai: Okay, having one less output does look cleaner, so lets keep this design.

PR #296, opened 2022-09-28
Rename force field name for solid participants

MakisH:
According to https://github.com/precice/openfoam-adapter/pull/252

Does not need a changelog entry (not previously released).

No clue why the PEP8 CI check fails, this does not touch any Python files.


davidscn (approved):
Pep 8 fails due to `./channel-transport-reaction/plot.py`, which is unrelated.

PR #301, opened 2022-10-24
Simplified heat exchanger tutorial

MakisH:
This adds a simplified heat exchanger tutorial, essentially extending the OpenFOAM-CalculiX flow-over-heated-plate tutorial with another channel below the plate.

```text
Cold inlet -> ====== Fluid1 ========== -> warmer outlet
                   == Solid ==
Hot inlet  -> ====== Fluid2 ========== -> cooler outlet
```

Compared to the heat exchanger tutorial:
- instead of Robin-Robin coupling, this demonstrates Dirichlet-Neumann coupling
- instead of composing two explicit coupling schemes, this uses a multi-coupling scheme
- instead of a complex geometry, this uses just three rectangles (2D case)

This is currently a draft. Open points:
- [ ] Fluid2: Move the interface from the bottom to the top.
- [ ] Fluid2: Set a temperature at the inlet different than the one of Fluid1
- [ ] Solid: Convert the bottom plate to an interface -> I need some help from some CalculiX expert here @boris-martin @KyleDavisSA 
- [ ] Extend `README.md` (suggestions welcome)

The configuration file looks like this:

precice-config

To test, it requires the bugfix implemented in https://github.com/precice/calculix-adapter/pull/108


boris-martin:
> Compared to the heat exchanger tutorial:
> 
>   -  instead of Robin-Robin coupling, this demonstrates Dirichlet-Neumann coupling
>   -  instead of composing two explicit coupling schemes, this uses a multi-coupling scheme
>   -  instead of a complex geometry, this uses just three rectangles (2D case)

Also, this one is transient whereas the other was a steady state computation. It might be worth it to rename the old case like `heat-exchanger-steady-state` or something like that.

As for the bottom interface, you'd need an additional face set and nodes set.

For the node set, you named it `ystart` in the YAML file but it seems the `ybegin` in the mesh generator should do the job.
For the faces set, something like the last part of the script but with bottom faces instead of top faces. I think you'd need to change the element indices and write S3 instead of S5 (according to https://web.mit.edu/calculix_v2.7/CalculiX/ccx_2.7/doc/ccx/node247.html and https://web.mit.edu/calculix_v2.7/CalculiX/ccx_2.7/doc/ccx/node26.html )

MakisH:
@boris-martin thanks a lot for the suggestions! I think I made it work. Probably I still have a heat source in CalculiX that I need to disable (how?).

This is how it looks at the moment, it still needs some tuning:

Screenshot from 2022-10-25 15-56-26

The top channel has an inlet of 300K and the bottom channel an inlet of 310K.

boris-martin:
What do you mean by heat sources in CalculiX ? there is no volumetric heat source in the case.
BCs should probably be cleaned up though, we say "310" on top and "0" at the end. Both should be overriden by preCICE but I think the convention is to set them at either 0 or the initial condition. But I don't know what is the optimal initial condition, a constant one is bound to have a discontinuity at one of the interfaces.

Unless we manage to get a linear (300 to 310) initial condition through a steady state step (similar to https://github.com/precice/calculix-adapter/pull/101) but that's out of my skill set. 

(The geometry looks like a spaceship to me :D )

MakisH:
> BCs should probably be cleaned up though, we say "310" on top and "0" at the end. Both should be overriden by preCICE but I think the convention is to set them at either 0 or the initial condition. But I don't know what is the optimal initial condition, a constant one is bound to have a discontinuity at one of the interfaces.

I changed both BCs to "0", I did not observe any difference. The interface is not continuous, but the fluid cases look like they make sense. The top fluid gets hotter, the bottom fluid gets cooler.

I am confused on why I don't need to synchronize the output files in the same way as we do in the two-participant OpenFOAM-CalculiX tutorial. This hints me towards visualization issues.

If I make both fluids have 300K inlets, it looks like they are both cooling the hot solid (~~and the iterations do not converge, at least till 0.13~~ this is also the case for the normal case).

Setting the initial condition to 305K leads to a bit smoother results, but still not continuous.

```
** Initial temperature constraint
** written by write_constraints_initialtemperature function
*INITIAL CONDITIONS,TYPE=TEMPERATURE
Nall,305.0
```

I guess what we really need is a better initial state (a linear distribution, as you described). But why does the two participant case behave better?

MakisH:
Running longer (t=5s) and initializing the solid domain with the mid temperature between the two fluid domains (T=305K) gives better results, even though still not completely smooth. The bottom interface seems to be more affected than the top:

Screenshot from 2022-11-08 14-54-24

boris-martin:
That's surprising, I would expect a fully symmetric result. Are the two fluids meshes similar in terms of resolution etc?

MakisH:
> That's surprising, I would expect a fully symmetric result. Are the two fluids meshes similar in terms of resolution etc?

Good point. Indeed, we have a much finer mesh next to the interface in the top participant, I forgot to update that. I will try again.

MakisH:
After flipping the mesh gradient for Fluid2, the results look much better:

Screenshot from 2022-11-08 19-13-08

Looking closer around the interface, this looks like a mapping issue to me:

Screenshot from 2022-11-08 19-13-42

We are currently using nearest-neighbor mapping, I will also check with RBF.

MakisH:
@IshaanDesai could you please have a look at this and check if you see anything suspicious?

Consider the following:
- I started from the flow-over-heated-plate-two-meshes case
- I changed the solid thickness to 0.4 (does it make any difference if we set it to 1, as in the original tutorial? I will also let it run and compare)
- I used our new naming conventions for tutorials with multiple participants: https://precice.org/community-contribute-to-precice.html#guidelines-and-hints

MakisH:
Corresponding PR in the website: https://github.com/precice/precice.github.io/pull/221

IshaanDesai (approved):
I ran the case till the end (`t = 5`) and the results look correct and so do the case files.

BenjaminRodenberg:
> I ran the case till the end (`t = 5`) and the results look correct and so do the case files.

I think the pictures provided in this PR are very useful and it would be nice to also have them somewhere in the documentation. @IshaanDesai , @davidscn , @MakisH : Does anybody of you by chance have some pictures of the result that we could add in the `README.md` for this case?

IshaanDesai:
Sadly do not have pictures already, but I guess it should be straightforward to run the case and create some good pictures?

MakisH:
I should have included a picture already when I contributed, but missed it. I also need this often, so I will add one.

PR #303, opened 2022-10-24
Rename multiple-perpendicular-flaps mesh names

### LINKED ISSUE #302: Confusing mesh names in multiple-perpendicular-flaps
Reading the configuration file of the [multiple perpendicular flaps tutorial](https://precice.org/tutorials-multiple-perpendicular-flaps.html), I get the impression that there are two fluid participants, `Fluid1` and `Fluid2` (which is not the case, the only fluid participant is `Fluid`):

https://github.com/precice/tutorials/blob/eaf55b17ae88f8b37c20ae8791467f16ed1e2f0e/multiple-perpendicular-flaps/precice-config.xml#L16-L40

We have as a [guideline](https://precice.org/community-contribute-to-precice.html#naming-conventions):
> Mesh names start with the participant/domain name, e.g. Fluid-Mesh.

The guideline seems to be followed, even if the example does not apply here.

I suggest that we name the branches `Fluid-Solid1-Mesh-Centers` (and similarly for the rest), as this makes it clear what the mesh is used for.

MakisH:
Fixes #302.

@uekerman how do you like this change? Does it fit our naming scheme?


uekerman (commented):
The suggested change makes things clearer, yes, but I don't particularly like to use other participants names. 
An alternative could be to use geometric names: `Fluid-Upstream-Mesh-Centers` and `Fluid-Downstream-Mesh-Centers`. Then, probably `Solid1` should also become `Solid-Upstream` etc.
We have no conventions yet for such multi-coupling setups.

MakisH:
One rule could be `ParticipantName-InterfaceName-Mesh-Modifier`, where `InterfaceName` the name of the interface/patch for the participant that defines that mesh. We don't always name these, but we do name them in OpenFOAM.

In this case, that would be `Fluid-Interface1-Mesh-Centers`.

In #301, where the defining participant is CalculiX, we can take the `patch:` name from the YAML file. This would lead to the not so elegant `Solid-flux_interface1-Mesh-Faces`. Which probably means we should cleanup the CalculiX patch names everywhere, if we want to follow this rule.

uekerman:
Yes, sounds good. But then we could also rename `Interface1` to something meaningful, right? Or is this name fixed?

MakisH:
> Yes, sounds good. But then we could also rename `Interface1` to something meaningful, right? Or is this name fixed?

These names are currently arbitrary and not used anywhere. We just need a name for each list in this dictionary format.

MakisH:
I renamed the participants, meshes, and data fields:

- `Solid1` becomes `Solid-Left` and `Solid2` becomes `Solid-Right`, following the directory names. We could further rename these, together with the directory names. Do we even worry about such "breaking" changes in the tutorials?
- For the Fluid meshes, I just used `Upstream` and `Downstream` as indicators.
- For the Solid meshes, I followed the usual `Participant-Mesh` name.
- The data fields are named based on the Upstream/Downstream indicator.

Should I further rename the solid participants to `Solid-Upstream` and `Solid-Downstream`? I have mixed feelings about this. On the one hand, it is not well motivated why we are using left/right and upstream/downstream. On the other hand, it helps one understand that the other participant's name is not part of the name. I tend towards renaming.

MakisH:
I documented the conventions in https://github.com/precice/precice.github.io/pull/216.

multiple-perpendicular-flaps/README.md
```diff
@@ -92,14 +92,14 @@ The scenario settings are implemented similarly for the nonlinear case.
    The solid participants are only designed for serial runs. To run the `Solid1` participant, execute the corresponding deal.II binary file e.g. by:
```
MakisH, line 92: We still have `Solid1` and `Solid2` here.

PR #308, opened 2022-11-16
AxialGeoMultiscale Tutorial

ezonta:
The goal of this PR is to have a tutorial case for the newly developed AxialGeoMultiscale mapping feature.
The tutorial consists of a 1D and a 3D pipe that exchange velocity and pressure values.

This PR is still a draft, because the tutorial case is not yet satisfactory - TO DO:
* [x] Fix data not being exchanged and Fluid2 running independently
* [x] Change substitute 1D script into correct Nutils implementation
* [x] Clean up case and write README.md


1D-3D-partitioned-pipe/precice-config.xml
```diff
@@ -0,0 +1,66 @@
...
+        <use-data name="Velocity"/>
+        <use-data name="Pressure"/>
+    </mesh>
+
+    <mesh name="Fluid2-Mesh">
+        <use-data name="Velocity"/>
+        <use-data name="Pressure"/>
+    </mesh>
+
+    <participant name="Fluid1">
```
MakisH, line 25: These participant names are a bit arbitraty. Let's rename them to `Fluid1D` and `Fluid3D`.

The directory names would then become `fluid1d-python` and `fluid3d-openfoam`.

MakisH: Also, maybe a better name for the tutorial itself would be `partitioned-pipe-multiscale` (fits with `partitioned-pipe`).

Alternatively, we could have `multiscale-partitioned-pipe`, similar to `multiple-perpendicular-flaps`, hinting towards the feature. I am not sure what is better, but I think that `partitioned-pipe-multiscale` is good for now.

ezonta:
There's still a bug in the tutorial regarding the run scripts. Running the tutorial from the participant folders with 
´´´
cd Fluid3D && ./run.sh
´´´
does not work. However, executing both from the tutorial folder (using the runFluid3D script) works fine. I am not sure why that is.

1D-3D-partitioned-pipe/precice-config.xml
ezonta: Renamed them according to the conventions.

partitioned-pipe-multiscale/fluid3d-openfoam/system/controlDict
```diff
@@ -0,0 +1,45 @@
...
+
+startFrom       startTime;
+
+startTime       0;
+
+stopAt          endTime;
+
+endTime         10.0;
+
+deltaT          0.01;
```
MakisH, line 20: The `coupling-window-size` defined in `precice-config.xml` is `0.005`, which is smaller than the solver time step `deltaT`. Maybe increase the `coupling-window-size` a bit, so that we only couple every few solver time steps (investigate how important this is for the results). We can also couple every solver time step, in which case we should really set the same value to both solvers. preCICE will overwrite these timestep sizes, which can be misleading if we set them as too small.

partitioned-pipe-multiscale/fluid3d-openfoam/system/controlDict
```diff
@@ -0,0 +1,45 @@
...
+
+writeControl    adjustableRunTime;
+
+writeInterval   0.01;
```
MakisH, line 24: For debugging, we probably do want to write the results after every time step. But in the final tutorial, I think we should aim for 10-20 result files in total. The user can always adjust this.

partitioned-pipe-multiscale/fluid3d-openfoam/system/decomposeParDict
```diff
@@ -0,0 +1,16 @@
...
+    format ascii;
+}
+
+numberOfSubdomains 2;
+
+method          simple;
+
+simpleCoeffs
+{
+    n               (2 1 1);
```
MakisH, line 14: With this, we are splitting per x-direction, meaning that only one rank will be participating in the coupling (assuming that the axis of the pipe is x).

It would be interesting (not necessarily in this PR) to check if the mapping also works in parallel when the interface is split between MPI ranks. I expect surprises.

partitioned-pipe-multiscale/fluid3d-openfoam/system/fvSolution
```diff
@@ -0,0 +1,74 @@
...
+solvers
+{
+
+    p
+    {
+            
+    solver           PCG;
+    preconditioner   DIC;
+    tolerance        1e-8;
+    relTol 	         1.0e-3;
```
MakisH, line 19:
```suggestion
    tolerance        1e-8;
    relTol 	     1.0e-3;
```

partitioned-pipe-multiscale/fluid3d-openfoam/system/preciceDict
```diff
@@ -0,0 +1,33 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      preciceDict;
+}
+
+preciceConfig "./precice-config.xml";
```
MakisH, line 10: This should probably be

```suggestion
preciceConfig "../precice-config.xml";
```

(combined with the new run script)

partitioned-pipe-multiscale/precice-config.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0"?>
+
+<precice-configuration>
```
MakisH, line 3: Please format this config file with the config formatter. See https://precice.org/community-contribute-to-precice.html#optional-help-us-with-some-checks

MakisH (commented):
This is already quite clean, thank you for the careful work! I added a few comments, mostly on details.

I have not yet tried to run it.

partitioned-pipe-multiscale/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+```bash
+cd fluid3d-openfoam && ./run.sh
+```
+
+## Results
+
+You should be able to see an established laminar profile at the inlet of the 3D participant.
+
+![Expected Result](images/tutorials-partitioned-pipe-multiscale-profile.png)
```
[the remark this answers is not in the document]
ezonta: That's a good idea, will do that once the Nutils script runs.

partitioned-pipe-multiscale/fluid1d-python/Fluid1D.py
```diff
@@ -0,0 +1,169 @@
...
+# TODO: uncomment for vtk output
+# from evtk.hl import pointsToVTK
+import numpy as np
+import precice
+
+def main():
+    
+    # number of nodes, length of domain and space interval
+
+    length = 10
```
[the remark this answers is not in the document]
ezonta: Will do that for the Nutils script!

partitioned-pipe-multiscale/README.md
```diff
@@ -0,0 +1,34 @@
...
+We exchange velocity data from the upstream 1D to the downstream 3D participant and for the pressure data vice versa. The config looks as follows:
+
+![Config visualization](images/tutorials-partitioned-pipe-multiscale-config.png)
+
+## How to run
+
+In two different terminals execute
+
+```bash
+cd fluid1d-python && ./run.sh
```
MakisH, line 23: We should probably rename this `-python` directory to `-nutils`.
We also need a section on available solvers, including the nutils version required, and the respective model.

partitioned-pipe-multiscale/fluid1d-python/Fluid1D.py
```diff
@@ -0,0 +1,112 @@
...
+            p_read = interface.read_scalar_data(pressure_id, vertex_ids)
+            p_read = numpy.maximum(0, p_read)     # filter out unphysical negative pressure values
+        else:
+            p_read = 0
+        usqr = domain.boundary['left'].integral('(u_0 - 1)^2 dS' @ ns, degree=2)
+        diricons = solver.optimize('u', usqr, droptol=1e-15)
+        ucons = diricons
+        stringintegral = '(p - ' + str(numpy.rint(p_read)) + ')^2 dS'
+        psqr = domain.boundary['right'].integral(stringintegral @ ns, degree=2)
+        pcons = solver.optimize('p', psqr, droptol=1e-15)
```
MakisH, line 67: I barely understand what is going on here, but a tolerance of `1e-15` sounds too strict to me.

MakisH (commented):
Some comments to get you started.

I want to look again into the Nutils code after you add some more comments, and I still need to run it. Since I assume you are still based on preCICE v2, please don't yet port it to v3.

partitioned-pipe-multiscale/fluid3d-openfoam/system/decomposeParDict
MakisH: Documented in an issue: https://github.com/precice/precice/issues/1667

partitioned-pipe-multiscale/fluid1d-python/Fluid1D.py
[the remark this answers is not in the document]
MakisH: Not sure if anything changed in the meantime, but the current indentation seems to be consistent with other Nutils tutorials we have. All good.

MakisH:
In the current configuration file, the logger is configured to show debug messages from the mapping:

```xml
  <log>
    <sink
      type="stream"
      output="stdout"
      filter="(%Severity% > debug) or (%Severity% >= debug and %Module% contains Mapping)"
      enabled="false" />
  </log>
```

This might be useful later, but I am going to make the file consistent with all other tutorials.

MakisH:
After porting the code to preCICE v3, the simulation runs but I don't seem to get any inlet in the 3D pipe.

Suspecting that I messed up something in https://github.com/precice/tutorials/pull/308/commits/6e622a98d0a7f6fe77a748d07d920890b25d6ab0

partitioned-pipe-multiscale/images/tutorials-partitioned-pipe-multiscale-profile.png
MakisH, line 1: To demonstrate that the mapping is what does the work, and not ParaView's interpolation, let's replace this picture with one showing the actual cell (face) values.

partitioned-pipe-multiscale/images/tutorials-partitioned-pipe-multiscale-config.png
MakisH, line 1: Let's also add a schematic to show the two domains and their boundary conditions.

MakisH:
I could also imagine that something went wrong when porting the old Nutils script to the new Nutils version. Find the old one in https://github.com/precice/tutorials/pull/308/commits/aa8fe2f7c59547612fd54dc6ab7f65ce5c17df70

MakisH:
Found the issue: I was setting the coupling interface to (0,0,0) intead of (0,0,1):

Screenshot from 2025-07-03 12-02-51

MakisH:
This PR has been superseded by https://github.com/precice/tutorials/pull/677

We now have two geometric multiscale tutorials:

- partitioned pipe: https://precice.org/tutorials-partitioned-pipe-multiscale.html
- water hammer: https://precice.org/tutorials-water-hammer.html

PR #315, opened 2022-12-05
Channel transport with OpenFOAM

tirgendetwas:
Channel transport tutorial using OpenFOAM for the fluid participant (and Nutils for the transport participant).

Tested with `serial-explicit` and `parallel-explicit` coupling, with sequential and parallel (MPI) runs for the OpenFOAM solver.

- [x] updated README.md


uekerman:
Good to have this new case!
Is there a motivation for adding it as a separate case and not "simply" extending https://precice.org/tutorials-channel-transport.html that I don't see?

MakisH:
> Good to have this new case! Is there a motivation for adding it as a separate case and not "simply" extending https://precice.org/tutorials-channel-transport.html that I don't see?

There is no particular motivation. This is still a draft and @tirgendetwas is planning to move the case there.

channel-transport/fluid-openfoam/system/decomposeParDict
```diff
@@ -0,0 +1,18 @@
+FoamFile {
+    version     2.0;
+    class       dictionary;
+    object      decomposeParDict;
+    format      ascii;
+}
+
+numberOfSubdomains 2;
```
MakisH, line 8: The parallel version does not work due to the selected preCICE mapping:

```
---[precice] ERROR:  For a parallel participant, only the mapping combinations read-consistent and write-conservative are allowed
```

Irrelevant to this PR, but we may want to consider that we have not yet tested the volume coupling in parallel.

MakisH (approved):
I ran the case and it looks good. The structure is correct and the results look at least qualitatively right (I have not compared to the fluid-nutils).

The only thing missing is an entry in the tutorial's `README.md`.

channel-transport/fluid-openfoam/system/preciceDict
```diff
@@ -0,0 +1,32 @@
...
+participant Fluid;
+
+modules (FF);
+
+interfaces
+{
+  Interface1
+  {
+    mesh              Fluid-Mesh;
+    patches           (obstacle);
```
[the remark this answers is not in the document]
tirgendetwas: Actually, no ^^

channel-transport/fluid-openfoam/system/decomposeParDict
tirgendetwas: Updated and tested with serial-explicit and parallel-explicit coupling

channel-transport/README.md
```diff
@@ -28,6 +28,8 @@ Fluid participant:
 
 * Nutils. For more information, have a look at the [Nutils adapter documentation](https://www.precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0.
 
+* OpenFOAM (pimpleFoam): tested with OpenFOAM v2112. For more information, have a look at the [OpenFOAM adapter documentation](https://precice.org/adapter-openfoam-overview.html).
```
MakisH, line 31: Is the version important? We should always try to check with the latest version anyway.

```suggestion
* OpenFOAM (pimpleFoam): For more information, have a look at the [OpenFOAM adapter documentation](https://precice.org/adapter-openfoam-overview.html).
```

Seems to be running with OpenFOAM v2212 anyway, using the current `volume-coupling` branch.

channel-transport/precice-config.xml
```diff
@@ -20,19 +20,19 @@
...
       <use-mesh name="Transport-Mesh" provide="yes" />
+      <use-mesh name="Fluid-Mesh" from="Fluid" />
       <read-data name="Velocity" mesh="Transport-Mesh" />
+      <mapping:rbf-compact-tps-c2
+        direction="read"
+        support-radius="0.4"
+        from="Fluid-Mesh"
+        to="Transport-Mesh"
+        constraint="consistent" />
     </participant>
```
MakisH, line 36: Why do we need to change the `precice-config.xml` to add the OpenFOAM case?
Why do we need to move the mapping to the `Transport` participant and exchange the `Fluid-Mesh`?

Note that, in any case, we will need to update to preCICE v3 before merging.

channel-transport/fluid-openfoam/system/decomposeParDict
```diff
@@ -0,0 +1,18 @@
...
+
+method          simple;
+
+simpleCoeffs
+{
+    n               (2 2 1);
+    delta           0.001;
+}
+
+// TODO test and find correct number
```
MakisH, line 18: I checked (with `./run.sh -parallel`) and the results look reasonable (did not compare values). Four subdomains are also reasonable for a tutorial, and cutting both horizontally and vertically helps us identify issues.

channel-transport/README.md
tirgendetwas: I added the version in order to document what version this has been tested with. I guess since it's the newest version I can leave it out, although it might be helpful in the future when newer OpenFOAM versions come out?

channel-transport/precice-config.xml
tirgendetwas: The mapping was moved so that parallel participants can work (following [this](https://precice.org/configuration-mapping.html#restrictions-for-parallel-participants) explanation)

channel-transport/README.md
MakisH: This would be good to document in the IDP report.
It would definitely be nice for the user to know it, but we avoid documenting the version in the tutorials. Reasons include:

1. We typically check that everything keeps running with the latest version supported the adapter, and add notes in the respective tutorial case if not.
2. We anyway don't do thorough validation of all tutorial cases, so as long as it runs and gives reasonable results, this is enough. Of course, we would document every dependency in detail in a publication, especially when we perform some validation.
3. It is considerable effort to update the version compatibility of every case regularly, and cases tend to look outdated, while they are not. Users may even assume the cases don't work with the latest version (or any version they have), and immediately try to install a different one. The compatibility should be dictated by this page: https://precice.org/adapter-openfoam-support.html

Note that we used to document the versions, but we don't do this anymore. If you find any such tutorial, please directly edit to remove the version, or let me know.

channel-transport/precice-config.xml
MakisH: Very good point, thanks! Then, this is relevant and we should keep it.
It only became apparent here, since our Nutils example does not run in parallel.

channel-transport/precice-config.xml
```diff
@@ -27,9 +27,8 @@
       <provide-mesh name="Transport-Mesh" />
       <receive-mesh name="Fluid-Mesh" from="Fluid" />
       <read-data name="Velocity" mesh="Transport-Mesh" />
-      <mapping:rbf-compact-tps-c2
+      <mapping:nearest-neighbor
```
MakisH: Note for the future reader: I think we need this change, since this case does not have non-conforming meshes. With preCICE v2.5, the initialization takes quite a while with RBF.

@BenjaminRodenberg @davidscn is there any reason to keep the RBF mapping in the channel transport tutorial?

uekerman: Meshes are not non-confirming, but non-matching, no? At least for the Nutils case. `nearest-neighbor` could lead to bad accuracy. Maybe the support radius needs tweaking.

MakisH: I would then suggest to at least run both cases and compare the results.

channel-transport/transport-nutils/transport.py
```diff
@@ -74,17 +74,17 @@ def main():
...
 
-        # read velocity values from interface
-        if interface.is_read_data_available():
-            velocity_values = interface.read_block_vector_data(velocity_id, vertex_ids)
+        precice_dt = participant.get_max_time_step_size()
+        # read velocity values from participant
+        velocity_values = participant.read_data(mesh_name, data_name, vertex_ids, precice_dt)
 
         # potentially adjust non-matching timestep sizes
         dt = min(dt, precice_dt)
```
MakisH, line 90: I think you need to read data at `dt`, not at `precice_dt`. This is the actual time step size your solver is performing.

With time interpolation now enabled by default, you can read values at any requested time. preCICE will interpolate between coupling time windows and give you a value that corresponds to this dt.

MakisH (changes requested):

channel-transport/precice-config.xml
MakisH: I ran both the Nutils-Nutils and the OpenFOAM-Nutils combinations, both with RBF and NN.
I did not notice any issues with the initialization time. The slowest step by far was the receiving Nutils participant's simulation, in both mapping cases.

The results I got look the same, but somehow I cannot open the (binary) Nutils VTK results with fieldcompare, at least not with the latest release from PIP.

I suggest reverting back to the RBF mapping, just in case. If runtime becomes an issue, we can adapt.

MakisH (approved):

PR #316, opened 2022-12-16
Fix broken link and minor update

BenjaminRodenberg:
* Update broken link
* Remove redundancy for less maintainment

### DIFF
--- a/partitioned-heat-conduction/README.md
+++ b/partitioned-heat-conduction/README.md
@@ -27,13 +27,13 @@ You can either couple a solver with itself or different solvers with each other.
 
 * FEniCS. Install [FEniCS](https://fenicsproject.org/download/) and the [FEniCS-adapter](https://github.com/precice/fenics-adapter). The code is largely based on this [fenics-tutorial](https://github.com/hplgit/fenics-tutorial/blob/master/pub/python/vol1/ft03_heat.py) from [1].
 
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
Maybe a good point to trigger this discussion: As far as I understand links are now checked manually. I guess this might be the reason that we now have a broken link here: If something has to be done manually it's often just not done at all.

In https://github.com/precice/tutorials/pull/230 we moved to a CRON job and then in https://github.com/precice/tutorials/pull/252 to the manual execution. As far as I understand In https://github.com/precice/tutorials/pull/230 the main motivation for the CRON job was to save computational ressources, but this turned out problematic in https://github.com/precice/tutorials/pull/252. Is there a reason for not going back to the `on: [push, pull_request]` execution? To me this looks less risky.

MakisH:
The reason for not having automatic links at the moment is that there were too many false positives, see https://github.com/precice/precice.github.io/pull/66

If we enabled the same checks (i.e., on the markdown level) on push and pull_request, I guarantee that they will always be failing, no matter if we have broken links or not. Instead, I suggested checking links with an external tool that will directly look at the website, such as (arbitrarily) https://www.deadlinkchecker.com/

I am not aware of any better (i.e., usable/reliable) solution that we could integrate in the CI at the moment.

MakisH (approved):
Thanks for fixing!

PR #318, opened 2023-01-04
Replaced all <use-mesh> tags by <receive-mesh> and <provide-mesh> tag…

Timo-Schrader:
This PR changes all `<use-mesh>` tags to `<provide-mesh>` and `<receive-mesh>` in */precice-config.xml which makes it compatible with current preCICE versions.


davidscn (approved):
Looks sane, thanks a lot. May I ask you to add a changelog entry as well?

PR #319, opened 2023-01-12
Updated perpendicular-flap/fluid-su2 to work with updated SU2 adapter for v7.5

j-signorelli:
Config file and initial flow field updated to work with updated SU2 adapter for v7.5


uekerman:
Closed in favor of #324

PR #320, opened 2023-01-23
Fixed a docker-compose race condition

valentin-seitz:
Sometimes the cleanup service ran faster than the openfoam-adapter container leading to a missing "eth0" entry in the network configuration of the `precice-config.xml`

Now the [startup order](https://docs.docker.com/compose/startup-order/) is enforced via conditions.


MakisH (approved):
Good catch, thank you!

PR #328, opened 2023-03-21
Add nightly tests/add field compare

valentin-seitz:
I integrated fieldcompare into the docker-compose services and also added a sample reference output via git lfs to compare the results to.

### DIFF
--- a/flow-over-heated-plate/tests/docker-compose.yml
+++ b/flow-over-heated-plate/tests/docker-compose.yml
@@ -53,7 +53,9 @@ services:
         condition: service_completed_successfully
       solid-openfoam:
         condition: service_completed_successfully
-    command: fieldcompare dir preCICE-output reference-output
+    command:
+      - /tests/flow-over-heated-plate/preCICE-output 
+      - /tests/flow-over-heated-plate/reference-output
 
   cleanup:
     depends_on:
--- a/flow-over-heated-plate/tests/docker-compose.yml
+++ b/flow-over-heated-plate/tests/docker-compose.yml
@@ -38,12 +38,27 @@ services:
     command: >
       /bin/bash -c "id && 
       cd /tests/flow-over-heated-plate/solid-openfoam &&
-      openfoam2112 ./run.sh | tee solid-openfoam.log 2>&1"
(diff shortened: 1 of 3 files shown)


flow-over-heated-plate/tests/docker-compose.yml
```diff
@@ -40,14 +40,29 @@ services:
     command: >
       /bin/bash -c "id && 
       cd /tests/flow-over-heated-plate/solid-openfoam &&
-      openfoam2112 ./run.sh | tee solid-openfoam.log 2>&1"
+      openfoam2112 ./run.sh | tee solid-openfoam.log 2>&1 &&
+      cd /tests/flow-over-heated-plate/ &&
+      tar -zxf reference-output.tar.gz"
```
MakisH, line 45: Wouldn't this step better fit into the `field-compare` service?
Why always call the results `reference-output`? This could be confusing.

flow-over-heated-plate/tests/docker-compose.yml
```diff
@@ -40,14 +40,29 @@ services:
...
+      - /etc/group:/etc/group:ro
+      - ../..:/tests
     depends_on:
       fluid-openfoam:
         condition: service_completed_successfully
       solid-openfoam:
         condition: service_completed_successfully
+    command:
+      - /tests/flow-over-heated-plate/preCICE-output 
+      - /tests/flow-over-heated-plate/reference-output
```
MakisH, line 60: I don't get this point. These are result directories, not commands.
The [field-compare action](https://github.com/dglaeser/fieldcompare-action) provides specific options:

```yaml
- name: regression-test
  uses: dglaeser/fieldcompare-action@v1.0
  with:
    result: "file.csv"
    reference: "reference_file.csv"
    cli-options: "-rtol 1e-2 -atol 1e-6 -rtol pressure:1e-5 -atol pressure:1e-5"
```

MakisH (commented):
I meant to look at this before merging, sorry for the delay.
Is the comparison supposed to work already? (in following PRs, please help me by describing how to check that the solution works)

I added two comments, please consider them in the follow-up PRs.

PR #332, opened 2023-03-27
oscillator: update to v3 API

BenjaminRodenberg:
Updates the oscillator case to the latest state of the preCICE v3 API (https://github.com/precice/python-bindings/commit/166c6e10a92049ef37f7caeba83d1899f358ac92 etc.)


BenjaminRodenberg:
Not sure whether we already want to break and update the tutorials, but I guess having a PR for this case already makes sense, because I had these changes locally anyway.

IshaanDesai (approved):
I ran this tutorial with the latest precice:develop and everything seems to be in order.

Issue #311, opened 2022-11-22
Tutorial case metadata


fsimonis:
## Problem description

The layout of the preCICE tutorials are designed to be easy to understand: case directories including participant directories for various solvers including scripts to clean and run the case.
This layout is sometimes insufficient as some cases use a single solver to run multiple participants such as the paritioned-heat-equation or the aste-turbine case.

This leads to issues where it becomes unclear on how to actually run a scenario, or even which solvers are available for which participant.

This can become tedious when running some tutorials in all combinations.

## Proposed solution 1 - case meta data

We add a small metadata file which encodes:
1. which participants are required
2. the options per participant
3. how to run them (default to ./run)

Example for the elastic-tube-3d which all provide a run-script:
```yml
# elastic-tube-3d/.case.yml
fluid:
  - fluid-openfoam
solid:
 - solid-calculix
 - solid-fenics
```

Example for paritioned-heat-equation which run scripts require arguments:
```yml
# partitioned-heat-equation/.case.yml
dirichlet:
  - fencis: ./run -d
  - nutils: ./run -d
  - openfoam-dirichlet
neumann:
  - fenics: ./run -n
  - nutils: ./run -n
  - openfoam-neumann
```

## Proposed solution 2 - case and solver meta data

We add solver metadata which encodes:
1. what the solver is based on (say the openfoam-adapter)
2. all participants the solver provides
3. how to run them (default to ./run)

We could then also provide case metadata which encodes the participants of the case.

Example for the elastic-tube-3d:
```yml
# elastic-tube-3d/.case.yml
participants:
  - fluid
  - solid
```
```yml
# elastic-tube-3d/fluid-openfoam/.solver.yml
needs: openfoam-adapter
provides:
  - fluid
```
```yml
# elastic-tube-3d/solid-calculix/.solver.yml
needs: calculix-adapter
provides:
  - solid
```
```yml
# elastic-tube-3d/solid-fenics/.solver.yml
needs: fenics-adapter
provides:
  - solid
```

Example for paritioned-heat-equation:
```yml
# partitioned-heat-equation/.case.yml
participants:
  - dirichlet
  - neumann
```
```yml
# partitioned-heat-equation/fencis/.solver.yml
needs: fenics-adapter
provides:
  - dirichlet: ./run -d
  - neumann: ./run -n
```
```yml
# partitioned-heat-equation/openfoam-dirichlet/.solver.yml
# same for openfoam-neumann
needs: openfoam-adapter
provides:
[...]

PR #334, opened 2023-03-30
Adding metadata yaml files in the tutorial cases

valentin-seitz:
The main idea is stolen from https://github.com/precice/tutorials/issues/311 but adapted to a more test centered language. 

This is just a sketch branch to see if this direction makes sense, would be easy to maintain and would ease the reviving of the systemstest :)


sytemstests/print_combinations.py
```diff
@@ -0,0 +1,47 @@
+import glob
+import os
+import yaml
+import itertools
+
+# Set the directory containing the tutorial folders
+tutorial_dir = '../'
```
MakisH, line 7: There is a typo in the name of the current directory:

```diff
- sytemtests
+ systemtests
```

but why not use the already specified in https://github.com/precice/tutorials/pull/220 directory `tests/`?

sytemstests/print_combinations.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+# Use globbing to find all the test.yaml files in the directory
+yaml_files = glob.glob(f'{tutorial_dir}/*/tests.yaml')
+
+print("Welcome to preCICE systemtest")
```
MakisH, line 14: Confusing :smirk: 

```suggestion
```

flow-over-heated-plate/tests.yaml
```diff
@@ -0,0 +1,8 @@
+name: flow-over-heated-plate
+couples: 
+  solid: 
+    - openfoam
+    - nutils
+    - fenics
+  fluid: 
+    - openfoam
```
MakisH, line 8: OpenFOAM can also be executed in parallel (`./run.sh -parallel`), we should also consider this attribute.

.github/workflows/systemtests-commit-trigger.yml
```diff
@@ -0,0 +1,12 @@
...
+  push:
+    branches:
+      - sketch-systest-framework
+
+jobs:
+  call-workflow-2-in-local-repo:
+    uses: ./.github/workflows/systemtests-playground.yml
+    with:
+      precice-version: develop
+      adapter-versions: openfoam=develop,fenics=v1.4.0
```
MakisH, line 12: Is the `with: adapter-versions: XYZ` supposed to override the default `adapter-versions` in the systemtests-playground? (just trying to understand the mechanism and what you have in mind)

.github/workflows/systemtests-playground.yml
```diff
@@ -0,0 +1,27 @@
+name: Systemtests Playground
+on:
+  workflow_call:
+    inputs:
+      precice-version:
+        type: string
+        description: 'Version of the Precice library to use (SHA or tag)'
+        required: true
+        default: 'master'
+        # Validation pattern can be added here if needed
```
MakisH, line 10: An example or some documentation link would be useful eventually.

.github/workflows/systemtests-playground.yml
```diff
@@ -0,0 +1,27 @@
+name: Systemtests Playground
+on:
+  workflow_call:
+    inputs:
+      precice-version:
+        type: string
+        description: 'Version of the Precice library to use (SHA or tag)'
```
MakisH, line 7: 😱 (shame! 🔔 shame! 🔔 shame! 🔔) 😅

```suggestion
        description: 'Version of the preCICE library to use (SHA or tag)'
```

MakisH (commented):
I understand that this is a draft, but I am trying to comment here to keep the rock rolling.

valentin-seitz:
> I understand that this is a draft, but I am trying to comment here to keep the rock rolling.

Thanks for the really fast feedback!
So the main idea in the playground is to write a reusable "execute me the tests" workflow. 
This way we can have smaller workflows for example for Adam Adapter which will be in their repository and trigger this workflow here. 
I will implement some interface which should work for both adapter and the upstream main library.

Therefore nobody up or downstream does not need to know about the actual tutorial cases it wants to run. 

I did not fully figure out how to actually build the adapter given a commit-sha, as i want to avoid pushing to the container registry just for 1 test build, but i need the container possibly in multiple workers, as a matrix job might make sense here.

MakisH:
> I did not fully figure out how to actually build the adapter given a commit-sha, as i want to avoid pushing to the container registry just for 1 test build, but i need the container possibly in multiple workers, as a matrix job might make sense here.

From my understanding, we would have to push to the container registry, but we would need some versioning scheme and some expiration policy.

I could imagine that Actions of different repositories could also share some cache, but this would be yet another additional mechanism.

In the long run, we would mostly have new images in the container registry for every new feature of an adapter, or for compatibility branches (e.g., for preCICE v3).

valentin-seitz:
> I think this should already help a lot.
> 
> Something not yet covered is that tutorials that contain a complete solver code do not follow the same structure (e.g., https://precice.org/tutorials-partitioned-heat-conduction.html). The reason is that, by enforcing the structure, we introduce duplication.
> 
> In those cases, the additional commands that @fsimonis described in https://precice.org/tutorials-partitioned-heat-conduction.html are important. The directory names are also different.
> 
> We could also discuss changes in the tutorials structure, as having exactly the same pattern everywhere (without introducing duplication) could be a simpler solution.
> 
> I appreciate the new tool `print_combinations.py` (we may later need to add more features, in which case the name `precice-tutorials` that @fsimonis suggested would make sense). We could, for example, use that as a way to run any tutorial and comparing with the reference results. More or less similar to the `tests/run-tests.sh` tool introduced in #220 I guess that, for the tests, we would also need the script to return something that can be used by another script.

My initial idea here was to built a system similar to what @fsimonis described in #311 but reuse the system to enable easy systemtests. 
I will try to extend the current solution a bit to incorporate the non-standard cases.

sytemstests/print_combinations.py
valentin-seitz: should be fixed in the next commit. 

Mainly as i am currently trying to have the docker-compose file not per case, but one for all tutorials. This should make it much more maintable as you only have to update one file.

MakisH:
Good direction with the `variations:`! :+1:

heat-exchanger/tutorial.yaml
```diff
@@ -0,0 +1,8 @@
+name: heat-exchanger
+couples:
```
uekerman, line 2:
```suggestion
participants: 
```
could be an alternative?

partitioned-heat-conduction/tutorial.yaml
```diff
@@ -0,0 +1,19 @@
+name: partitioned-heat-conduction
+couples: 
+  heat-conduction:
```
uekerman, line 3: Why is this level needed?

uekerman (commented):
Looks interesting!

Let's also think about naming alternatives to `tutorial.yaml`.
Maybe `metadata.yml` could already be easier to grasp? Not sure yet.

partitioned-heat-conduction/tutorial.yaml
valentin-seitz: From my understanding and how other tutorials work is that you have 1 `./run.sh` for each participant. Since we use the same solvers for both participants this reduces to just one `./run.sh` for both participants. 
One could also modify the case a bit and make a yaml file similar to this: 
```yaml

name: partitioned-heat-conduction
participants: 
  left-section:
    fenics:
      variations:
        neumann:
          run: ./fenics/run.sh -n 
        dirichlet: 
          run: ./fenics/run.sh -n 
    nutils:
      variations:
        neumann:
          run: ./nutils/run.sh -n 
        dirichlet:
          run: ./nutils/run.sh -n 
  right-section:
    fenics:
      variations:
        neumann:
          run: ./fenics/run.sh -n 
        dirichlet: 
          run: ./fenics/run.sh -n 
    nutils:
      variations:
        neumann:
          run: ./nutils/run.sh -n 
        dirichlet:
          run: ./nutils/run.sh -n 
```
This would be consistent with other tutorials and also allow for the easy execution of the [combinations](https://precice.org/tutorials-partitioned-heat-conduction.html#note-on-the-combination-of-nutils--fenics) 
On the other hand it gets a bit verbose.

heat-exchanger/tutorial.yaml
valentin-seitz: totally! This is more in agreement with the naming scheme i guess :)

partitioned-heat-conduction/tutorial.yaml
MakisH: Actually, if we have the `participant`, we don't need the `variations`:

```yaml
name: partitioned-heat-conduction
participants: 
  left-section:
    fenics:
        run: ./fenics/run.sh -n
```

(`variations` could still be helpful for parallel runs, but I am not sure for what else, then)

valentin-seitz: Maybe i am wrong, but the way i understood it is, that there are two ways to couple the left-section with the right one: 
One via Neumann BC and one via Dirichlet BCs. So i think we still need them. 
But i think the name `variants` is a bit more preCISE here. 
I think it makes sense to keep the `./run.sh` as a common interface to start a participant and then just select the `variant` with a flag like this:
```yaml
name: partitioned-heat-conduction
variants: 
  neumann-bc:
    additional-flags: -n
  dirichlet-bc:
    additional-flags: -d
participants: 
  left-section:
    - fenics
  right-section:
    - nutils
```
By design we would generate for each `variant` each combination of participants. A `variant` then would enforce the same flags on the `./run.sh` for each participant. 

Parallel runs on the other hand is more a participant specific thing. I would keep it more the participants themself if they are capable of running parallel like this:
```yaml
participants: 
  left-section:
    openfoam:
      parallel: true
    nutils:
```

A launch would then add the `-parallel` flag to the `./run.sh` command

MakisH: > Maybe i am wrong, but the way i understood it is, that there are two ways to couple the left-section with the right one: One via Neumann BC and one via Dirichlet BCs.

At the moment, this is not the case. In all tutorials that are like this, who is Neumann and who is Dirichlet (left or right) is hard-coded.

fsimonis:
@MakisH @valentin-seitz Do you plan to embed the parallelism into the metadata?
If you just care to globally turn this on or off, you could also use an env to set the default behaviour. Example being to set `PRECICE_TUTORIALS_PARALLEL` to make every `./run.sh` start the solver in parallel if supported.

Is there any reason to prefer one tutorial-metadata file over a tutorial- and a case-metadata file?

As a final note, @valentin-seitz your proposed metadata doesn't contain information on which case requires which adaptor/binding. This may be required to filter solver combinations.

MakisH:
> @MakisH @valentin-seitz Do you plan to embed the parallelism into the metadata?

Yes: https://github.com/precice/tutorials/pull/334/files#r1159848901

> If you just care to globally turn this on or off, you could also use an env to set the default behaviour. Example being to set `PRECICE_TUTORIALS_PARALLEL` to make every `./run.sh` start the solver in parallel if supported.

A global switch could turn out a bit inconvenient here:
- I would normally test some tutorials in serial and some in parallel, to cover a larger spectrum with fewer combinations
- Running some cases in serial and some in parallel could also overlap runs in time, saving in total runtime
- Not all cases use the same number of ranks. Most are easy to run in four ranks (or even in serial), while other may need many more (such as the FSI3). It is a good question whether we want to include such large cases in the system tests. I think we should, but maybe in a different frequency or system.

> Is there any reason to prefer one tutorial-metadata file over a tutorial- and a case-metadata file?

Not really, we could indeed have one per tutorial and one per (solver) case. But since the one per tutorial is already rather short, is there any reason to prefer the multiple files?

> As a final note, @valentin-seitz your proposed metadata doesn't contain information on which case requires which adaptor/binding. This may be required to filter solver combinations.

This is currently implicitly defined by the [name of the directories](https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial), I am not sure if we need an additional explicit declaration.

fsimonis:
> Not really, we could indeed have one per tutorial and one per (solver) case. But since the one per tutorial is already rather short, is there any reason to prefer the multiple files?

Some _very_ weak reasons for separate files:
* it may be easier to generate a case setup skeletons if we ever decide to add such a feature.
* it may be less prone to merge conflicts as the tutorial file only defines cases and the cases contain the majority of the description

Working and changing many files can be a huge pain though.

> This is currently implicitly defined by the [name of the directories](https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial), I am not sure if we need an additional explicit declaration.

Isn't it simpler to make this explicit to avoid guesswork with things like:
* perpendicular-flap/solid-solids4foam (there is no adapter for solids4foam, correct?)
* partitioned-pipe/fluid1-openfoam-sonicliquidfoam (compound names)
* partitioned-heat-conduction/nutils (which doesn't have an adapter as such)

MakisH:
> it may be less prone to merge conflicts as the tutorial file only defines cases and the cases contain the majority of the description

I like this argument. I would trust @valentin-seitz here, based on what feels easier to write and maintain.

> Isn't it simpler to make this explicit to avoid guesswork with things like:

Good point, we do have these cases. So, an explicit label would be nice.

valentin-seitz:
Hi all, 
a lot changed since i last contributed here. I discussed a new metadata concept with @MakisH  
We tried to focus on verbosity without to much duplication. Nobody likes implicitly named stuff :)

We basically come up with a few building blocks:

- Participants: Which correspond to the "domains" beeing coupled. For FSI its [Fluid, Solid]. We need minimum one case of each participant to run the coupled simulation.
- Cases: Correspond to a participant "implementation". Each of them is one folder. They provide run and clean commands and require a component to run. 
- Components: These things correspond to adapters, or anything that would make a case run. They will be globally defined in a `components.yaml` describing the components and their required input params as e.g GIT Refs. At some point they will be used to 'render' a docker-compose file which corresponds to the systemtest

elastic-tube-3d/metadata.yaml
```diff
@@ -0,0 +1,36 @@
+name: Elastic tube 3D
+path: elastic-tube-3d # relative to git repo 
+url: https://precice.org/tutorials-elastic-tube-3d.html
+clean: ./clean-tutorial.sh 
+tags: # just exemplary here 
+  - nearest-projection
+  - IMVJ
+  - FSI
```
MakisH, line 8: A better example here would be:

```suggestion
tags: # just exemplary here 
  - long
  - nearest-projection
  - IMVJ
```

The idea of the tags is that we will be able to run specific tests on demand (e.g., only run the "long" tutorials nightly/weekly), or even on specific runners (e.g., for cases requiring licensed codes).

We could also use these tags for specific features (e.g., `nearest-projection` or `IMVJ`), but I see this as secondary in priority.

Tags such as `FSI` probably do not help us much, as they would make the corresponding checks quite fine-grained, without a clear use case.

elastic-tube-3d/metadata.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - Fluid 
+  - Solid 
+
+cases:
+  fluid-openfoam:
+    particpant: Fluid
+    directory: ./fluid-openfoam
+    run: run.sh 
+    clean: clean.sh 
+    component: openfoam-adapter
```
MakisH, line 20: Is `component` a list or does it only take one value?
If it is a list, then `components` would make more sense.

flow-over-heated-plate/metadata.yaml
```diff
@@ -0,0 +1,50 @@
+name: Flow over heated plate
+path: flow-over-heated-plate # relative to git repo 
+url: https://precice.org/tutorials-flow-over-heated-plate.html
+clean: ./clean-tutorial.sh 
+tags: # just exemplary here 
+  - CHT 
+  - conjugate-heat transfer
```
MakisH, line 7:
```suggestion
tags:
  - Aitken
```

flow-over-heated-plate/metadata.yaml
```diff
@@ -0,0 +1,50 @@
...
+    run: run.sh 
+    clean: clean.sh 
+    component: openfoam-adapter 
+
+tests:
+  openfoam-openfoam:
+    suite: nightly
+    cases:
+      - fluid-openfoam
+      - solid-openfoam
```
MakisH, line 47: I am not sure where we landed on the `tags` and `suite`: there is some overlap in terms of use cases.

At the same time, `tags` belong to the tutorial, which `suite` to the `tests`:
1. We could inquire which tutorials use certain features (also interesting for the user and the documentation).
2. We could also use the `suite` (better: `test-suites`) to add certain combinations of cases to multiple test suites.
3. The `long` tag I proposed in a separate comment would then kind-of duplicate the use case of the `nightly` suite. Would it make any sense to keep the `long` tag if we already have a `nightly` suite?

systemtests/components.yaml
```diff
@@ -0,0 +1,16 @@
+openfoam-adapter:
+  repository: https://github.com/precice/openfoam-adapter
```
MakisH, line 2: Detail: This is a GitHub page URL, not the URL of the repository.

To make this more useful: How could we integrate, e.g., cases from private GitLab repositories? Not something to solve now, but the design needs to be extensible towards that direction.

If I give an SSH address to such a private repository, and I have already registered the public key of the runner on GitLab, will it be able to fetch the component?

Same holds for the tutorials/tests.

systemtests/components.yaml
```diff
@@ -0,0 +1,16 @@
...
+  docker-service-template: components/openfoam-adapter.yml
+  requires:
+    - openfoam:REF
```
MakisH, line 5: What can `REF` be? Any GitHub branch/commit/tag?

systemtests/components.yaml
```diff
@@ -0,0 +1,16 @@
...
+fenics-adapter:
+  repository: https://github.com/precice/fenics-adapter
+  docker-service-template: components/nutils-adapter.yml
+  requires:
+    - fenics:REF
+    - pythonbindings:REF
+
+## maybe respostiryx
+
+#systemtests(components_with_versions{},true)
```
MakisH, line 16: What is this for?

systemtests/components.yaml
```diff
@@ -0,0 +1,16 @@
+openfoam-adapter:
+  repository: https://github.com/precice/openfoam-adapter
+  docker-service-template: components/openfoam-adapter.yml
+  requires:
+    - openfoam:REF
+    - pythonbindings:REF
+fenics-adapter:
+  repository: https://github.com/precice/fenics-adapter
+  docker-service-template: components/nutils-adapter.yml
```
MakisH, line 9: FEniCS is not Nutils :smirk:

systemtests/components/openfoam.yaml
```diff
@@ -0,0 +1,10 @@
+image: "ghcr.io/precice/openfoam-adapter:{{ TAG_PRECICE }}-{{ TAG_OPENFOAM_ADAPTER }}"
+user: ${MY_UID}:${MY_GID}
+volumes:
+  - /etc/passwd:/etc/passwd:ro #TODO: Maybe move those into a more general section ? 
+  - /etc/group:/etc/group:ro
```
MakisH, line 5: I am not sure if we can use the same for arbitrary images (probably we can), so let's just leave it here for now. It also makes things more transparent and easy to process.

systemtests/print_metadata.py
```diff
@@ -0,0 +1,17 @@
...
+from classes import Tutorials , Components
+
+available_components = Components.from_yaml("./components.yaml")
+
+tutorial_dir = '../'
+available_tutorials = Tutorials.from_path("../")
+
+print("The following tutorials are described by the metadata")
+for tutorial in available_tutorials.tutorials:
+    print(tutorial)
```
MakisH, line 17: Since we provide tags, we should also be able to filter by tags. I would expect that writing:

```shell
python3 print_metadata.py --tag=nearest-projection
```

would only select the Elastic tube 3D.

Or would this only come to a separate script that triggers the tests?

If it makes things more complicated, I would rather skip any functionality related to tags for now.

MakisH (commented):
Getting into a nice shape!

Don't forget to add some documentation. Even a short `README.md` would really help.

Also, I still don't like the name `systemtests/`. I would rather prefer `tests/`, as it may be clearer and easier to spot for more people.

partitioned-heat-conduction/tutorial.yaml
MakisH: Not relevant anymore since https://github.com/precice/tutorials/pull/334/commits/0aaff578a34f485fcebd1eec36a018e13c97b783

MakisH:
Some of my comments above may be a bit too complicated for now, feel free to skip and open issues to remember.

I think the schema looks already quite good. We should then apply this to all tutorials (even with no tags), and focus on making a manual triggering with only `develop` branches functional. We will soon need it quite much for testing together all v3 components.

systemtests/components.yaml
valentin-seitz: Thats a good question: I imagine it to be any valid git ref. Anything you can checkout basically. 
But this of course complicates the instantiation of the docker-container a bit. We bascially need a mechanism to check if a container with this ref combination is already pushed to a registry (`latest` would be a example here) or if its a certain branch we just pushed to (in case of a feature PR).
In that case i think building the Docker container locally makes a bit more sense.

systemtests/components.yaml
valentin-seitz: will be removed ;)

systemtests/components.yaml
valentin-seitz: Currently this is just a nice piece of information. 
Its not even read into python. But going forward this might be important together with the location of the Dockerfile of that component, which we might need to pull in order to build the service.

systemtests/print_metadata.py
valentin-seitz: Yes i also feel, that tags might be helpful later but i would not consider it bare minimum functionality.

valentin-seitz:
Again, a small eternity has passed, but here i am again: 
This time with a somehow and only maybe working version v.0.1.beta of the systemtests. 

I added a lot of Code to make it work. This definitely needs some cleanup. 
Maybe @MakisH  you can trying running it on your machine. with `python systemtests.py --components=openfoam-adapter`

This should only execute the flow-over-heated-plate right now.

systemtests/classes.py
```diff
@@ -0,0 +1,366 @@
...
+
+        current_dir = os.getcwd()
+        tutorial_path = f"../{self.tutorial.path}"
+        os.chdir(tutorial_path)
+        with open("docker-compose.yaml", 'w') as file:
+            file.write(docker_compose_content)
+
+        try:
+            # Execute docker-compose command
+            process = subprocess.Popen(['docker-compose', 'up'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
MakisH, line 343: This seems to depend on `docker-compose`, which is the old version of the now `docker compose`: https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose

systemtests/docker-compose.template.yaml
```diff
@@ -0,0 +1,50 @@
+version: "3.9"
+services:
+  prepare:
+    image: ubuntu:latest
+    user: ${MY_UID}:${MY_GID}
```
MakisH, line 5: Where should `MY_UID` and `MY_GID` be defined?
I was previously using a shell script just to take care of this environment preparations.

systemtests/systemtests.py
```diff
@@ -0,0 +1,51 @@
+import glob
+import os
+import yaml
+import itertools
+import argparse
+
+
```
MakisH, line 7: Something does not work for me:

```
$ python3 systemtests.py --components=openfoam-adapter
parsing the following yaml files. ['../elastic-tube-3d/metadata.yaml', '../flow-over-heated-plate/metadata.yaml']
About to run the following systemtests [Flow over heated plate [fluid-openfoam, solid-openfoam]]
Now running Flow over heated plate with [fluid-openfoam, solid-openfoam]

The MY_UID variable is not set. Defaulting to a blank string.
The MY_GID variable is not set. Defaulting to a blank string.
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 394, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1037, in _send_output
    self.send(msg)
  File "/usr/lib/python3.10/http/client.py", line 975, in send
    self.connect()
  File "/usr/lib/python3/dist-packages/docker/transport/unixconn.py", line 30, in connect
    sock.connect(self.unix_socket)
PermissionError: [Errno 13] Permission denied

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 755, in urlopen
    retries = retries.increment(
  File "/usr/lib/python3/dist-packages/urllib3/util/retry.py", line 532, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/lib/python3/dist-packages/six.py", line 718, in reraise
    raise value.with_traceback(tb)
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 394, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1037, in _send_output
    self.send(msg)
  File "/usr/lib/python3.10/http/client.py", line 975, in send
    self.connect()
  File "/usr/lib/python3/dist-packages/docker/transport/unixconn.py", line 30, in connect
    sock.connect(self.unix_socket)
urllib3.exceptions.ProtocolError: ('Connection aborted.', PermissionError(13, 'Permission denied'))

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/docker/api/client.py", line 214, in _retrieve_server_version
    return self.version(api_version=False)["ApiVersion"]
  File "/usr/lib/python3/dist-packages/docker/api/daemon.py", line 181, in version
    return self._result(self._get(url), json=True)
  File "/usr/lib/python3/dist-packages/docker/utils/decorators.py", line 46, in inner
    return f(self, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/docker/api/client.py", line 237, in _get
    return self.get(url, **self._set_request_timeout(kwargs))
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 555, in get
    return self.request('GET', url, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 655, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/adapters.py", line 498, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', PermissionError(13, 'Permission denied'))

During handling of the above exception, another exception occurred:

  File "/usr/bin/docker-compose", line 33, in <module>
    sys.exit(load_entry_point('docker-compose==1.29.2', 'console_scripts', 'docker-compose')())
  File "/usr/lib/python3/dist-packages/compose/cli/main.py", line 81, in main
    command_func()
  File "/usr/lib/python3/dist-packages/compose/cli/main.py", line 200, in perform_command
    project = project_from_options('.', options)
  File "/usr/lib/python3/dist-packages/compose/cli/command.py", line 60, in project_from_options
    return get_project(
  File "/usr/lib/python3/dist-packages/compose/cli/command.py", line 152, in get_project
    client = get_client(
  File "/usr/lib/python3/dist-packages/compose/cli/docker_client.py", line 41, in get_client
    client = docker_client(
  File "/usr/lib/python3/dist-packages/compose/cli/docker_client.py", line 170, in docker_client
    client = APIClient(use_ssh_client=not use_paramiko_ssh, **kwargs)
  File "/usr/lib/python3/dist-packages/docker/api/client.py", line 197, in __init__
    self._version = self._retrieve_server_version()
  File "/usr/lib/python3/dist-packages/docker/api/client.py", line 221, in _retrieve_server_version
    raise DockerException(
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', PermissionError(13, 'Permission denied'))

Docker Compose process failed with exit code: 1
```

flow-over-heated-plate/precice-config.xml
```diff
@@ -31,7 +31,8 @@
         from="Solid-Mesh"
         to="Fluid-Mesh"
         constraint="consistent" />
-    </participant>
+      <export:vtk directory="../preCICE-output" />
```
MakisH, line 34: These exports should not always be enabled. Instead, the script running the system tests should modify the preCICE config to add this line.

systemtests/docker-compose.template.yaml
valentin-seitz: i just set them prior to executing the systemtests command. A bit of a workaround.

systemtests/README.md
```diff
@@ -1,6 +1,26 @@
 # Systemtests
-## How to use: 
-The intented use of this python script hodgepodge is to ease the execution of systemtests. 
+## What do they solve?
+Since the last Systemtest got deprecated at some point a new precice release involves a lot of manual system testing to ensure stable operations between the different adapters. 
+
+## Why is this in the tutorial repository?
+Most of the cases we currently test new precice releases on are the tutorial cases. They cover quite a broad spectrum of different solvers and domains. Putting everything into a seperate repository would make maintance harder as you would split up things that belong together.
```
MakisH, line 6: OK for now, but the `README.md` has the perspective of someone that discovers a project for the first time. The "last system tests" comment is something better fit for the PR description, or some "history" / "previous work" section in the README.

elastic-tube-3d/metadata.yaml
valentin-seitz: One case only can be realized using 1 component. So the singular here is correct i guess?

flow-over-heated-plate/metadata.yaml
valentin-seitz: This is now a bit redudant as we have `tests.yaml` instead of defining tests in each and every tutorial file

systemtests/README.md
valentin-seitz: I added some more helpful comments about running. But since everything is still a bit unstable i just wanted to provide some hints and explain the terminology i use.

elastic-tube-3d/metadata.yaml
MakisH: It is still a list, which in that special case it is a list of one element. In my view, singular here is confusing ("which component do I choose? How do I specify the other components?"), while the plural is kind of standard practice (but I cannot bring examples right now).

flow-over-heated-plate/metadata.yaml
MakisH: Let's drop all discussions regarding tags now. We will see the use case later.

systemtests/README.md
MakisH: Alright, let's keep it rolling, just remember to update this README once in a while (and before the next review).

valentin-seitz:
@MakisH I feel we should merge this as i think its a good base to work upon with some more specific PRs.

systemtests/components.yaml
```diff
@@ -0,0 +1,72 @@
...
+      default: "v1"
+    precice:
+      git-ref: True
+      description: Version of precice to use
+      default: "latest"
+
+
+calculix-adapter:
+  repository: https://github.com/precice/fenics-adapter
+  template: component-templates/nutils-adapter.yaml
```
MakisH, line 38: typo

systemtests/components.yaml
```diff
@@ -0,0 +1,72 @@
...
+      description: Version of precice to use
+      default: "latest"
+    openfoam-adapter-ref:
+      git-ref: True
+      description: Ref of the actual openfoam adapter
+      default: "latest"
+    
+fenics-adapter:
+  repository: https://github.com/precice/fenics-adapter
+  template: component-templates/nutils-adapter.yaml
```
MakisH, line 20: typo

systemtests/components.yaml
```diff
@@ -0,0 +1,72 @@
...
+      default: "v1"
+    precice:
+      git-ref: True
+      description: Version of precice to use
+      default: "latest"
+
+
+
+nutils-adapter:
+  repository: https://github.com/precice/fenics-adapter
```
MakisH, line 57: typo

systemtests/components.yaml
```diff
@@ -0,0 +1,72 @@
...
+  parameters:
+    pythonbindings:
+      type: git
+      semnantic: Version of the pythonbindings to use
+
+    fenics:
+      git-ref: False
+      options: ["v1","v2"]
+      semantic: Version of FENICS to use
+      default: "v1"
```
MakisH, line 48: Why FEniCS here?

PR #337, opened 2023-05-15
Update fvSolution to fit OpenFOAM 9 and newer

theBelpo:
Adds outerCorrectorResidualControl dictionary adapted to fit OpenFOAM 9 requirements. Also modifies wrong dictionary name for OpenFOAM 6, 7, 8 ("outerResidualControl" changed to "outerCorrectorResidualControl")

I should specify I am not sure this change does not apply to older OpenFOAM versions, but I can confirm based on experience that OpenFOAM 9 will not run without the modified dictionary.

### DIFF
--- a/elastic-tube-3d/fluid-openfoam/system/fvSolution
+++ b/elastic-tube-3d/fluid-openfoam/system/fvSolution
@@ -209,6 +209,25 @@ outerResidualControl
 }
 */
 
+// OpenFOAM 9 (.org) or newer
+// see also https://github.com/OpenFOAM/OpenFOAM-dev/commit/4c8122783aedaa7dadf0486163a98350e625db32
+/*
+outerCorrectorResidualControl
+{
+U
+{
+    tolerance 1e-5;
+    relTol	  0;
+}
+
+p
+{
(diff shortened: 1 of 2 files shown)


MakisH:
Thank you! We also have a branch specific to OpenFOAM10, as the changes needed for that are quite extensive: https://github.com/precice/tutorials/tree/OpenFOAM10

We should also add a hint here: https://precice.org/adapter-openfoam-support.html#supported-openfoam-versions

We are currently inconsistent with where/how we document and support different OpenFOAM versions in the tutorials. For OpenFOAM 6, it looks like we have this note in the turek-hron-fsi3 and in the elastic-tube-3d cases, but not anywhere else.

theBelpo:
Thank you! I'm just a beginner but I hope I could be of help!
I am currently using OpenFOAM 9 on WSL for my uni project but this issue seems indeed to apply also to the OpenFOAM 10 and develop branches, hope it can be fixed so that other students do not stumble on the same problem.

Have a nice day!

MakisH (approved):

MakisH (approved):

PR #342, opened 2023-06-30
Add trixi participant to channel transport tutorial

uekerman:
Open TODOs:

- [ ] Change geometry to include step
- [ ] Get proper cell coordinates (currently first-order average) 
- [ ] Allow arbitrary polynomial degree (currently hacked to 2)

Joint work with @sloede


MakisH:
This code is written for preCICE v2, but it should be straight-forward to upgrade it to preCICE v3: https://precice.org/couple-your-code-porting-v2-3.html

PR #347, opened 2023-07-20
DRAFT: Add Systemtests

valentin-seitz:


MakisH:
Is there something I can already try out / review here?

valentin-seitz:
From my side this is good enough to be merged to develop and iterate there in smaller PRs when adding adapters etc.

tools/tests/component-templates/nutils-adapter.yaml
```diff
@@ -0,0 +1,13 @@
+image: "ghcr.io/precice/openfoam-adapter:{{ params["openfoam-adapter-ref"] }}" # TODO: Update
```
MakisH, line 1: Remember to overwrite this file later

MakisH (approved):
I have previously reviewed most of the changes in other PRs. This looks good enough for now, let's merge it and start contributing in parallel in smaller PRs.

Of course, it needs some cleanup here and there, but mostly to replace placeholders with their actual implementations.

I could run the tests locally with:

```bash
python3 systemtests.py --suites=openfoam-adapter-release
```

which fails because of mismatch between tutorials branch (based on develop) and preCICE version (v2.5.0). Checking out the case files from `master` helped:

```bash
git checkout master -- ../../flow-over-heated-plate
```

Before running again, one needs to remove the temporary `runs/` folder in the tutorials root. These files are currently owned by root (@valentin-seitz please open an issue).

I was also able to trigger the workflow from the GitHub CLI (which succeeded).

PR #350, opened 2023-07-31
Volume-coupled flow tutorial

tirgendetwas:
Tutorial to validate volume coupling on a mesh region *and* the reading of internal fields in OpenFOAM. Flow is introduced into the system through a square region of cells. The dummy participant writes Velocity at a constant value, the fluid participant reads that value on a `cellSet` and enforces it through `fvOptions`.

This tutorial could be a basis for a more complex scenario where the dummy solver is replaced or expanded.

TODO:

- [x] README.md


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
MakisH, line 1: This file should not be here

channel-transport/fluid-openfoam/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -e -u
+
+blockMesh
+touch fluid-openfoam-pimplefoam.foam
```
MakisH, line 5:
```suggestion
touch fluid-openfoam.foam
```

(probably copy-paste from the partitioned pipe tutorial, where we have more than one OpenFOAM cases)

channel-transport/fluid-openfoam/system/fvSolution
```diff
@@ -0,0 +1,74 @@
...
+solvers
+{
+
+    p
+    {
+            
+    solver           PCG;
+    preconditioner   DIC;
+    tolerance        1e-8;
+    relTol 	         1.0e-3;
```
MakisH, line 19:
```suggestion
    tolerance        1e-8;
    relTol 	     1e-3;
```

channel-transport/precice-config.xml
```diff
@@ -19,17 +19,19 @@
     </mesh>
 
     <participant name="Fluid">
-      <provide-mesh name="Fluid-Mesh" />
-      <receive-mesh name="Transport-Mesh" from="Transport" />
+      <use-mesh name="Fluid-Mesh" provide="true"/>
```
MakisH, line 22: This PR reverts to the preCICE v2 config format. We need to keep it in the v3 format.

volume-coupled-flow/dummy-nutils/dummy.py
```diff
@@ -0,0 +1,70 @@
...
+from nutils import function, mesh, cli, solver, export
+import treelog as log
+import numpy as np
+import precice
+from mpi4py import MPI
+
+
+def main():
+
+    print("Running utils")
```
MakisH, line 13:
```suggestion
    print("Running Nutils")
```

volume-coupled-flow/dummy-nutils/dummy.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # cloud of Gauss points
+    uniform_grid = domain.sample("uniform", degree=1)
+
+    # Nutils namespace
+    ns = function.Namespace(fallback_length=2)
+    ns.x = geom
+
+    # preCICE setup
+    interface = precice.Interface("Dummy-Velocity", "../precice-config.xml", 0, 1)
```
MakisH, line 33: Is "Dummy" the right name? We usually name the participants based on the kind of domain they solve (fluid, solid, etc).

volume-coupled-flow/fluid-openfoam/0/U_vol
```diff
@@ -0,0 +1,40 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       volVectorField;
+    object      U_vol;
```
MakisH, line 6: Is this file required, or is it created by the adapter if it does not already exist?
In case of the latter, better remove it.

volume-coupled-flow/fluid-openfoam/system/controlDict
```diff
@@ -0,0 +1,46 @@
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    object      controlDict;
+}
+
+application     buoyantPimpleFoam;
```
MakisH, line 9: Is this the right solver?
By the way, the case also needs a README.md

volume-coupled-flow/fluid-openfoam/system/topoSetDict
```diff
@@ -0,0 +1,29 @@
+/*--------------------------------*- C++ -*----------------------------------*\
+| =========                 |                                                 |
+| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
+|  \\    /   O peration     | Version:  v2212                                 |
+|   \\  /    A nd           | Website:  www.openfoam.com                      |
+|    \\/     M anipulation  |                                                 |
+\*---------------------------------------------------------------------------*/
```
MakisH, line 7: Remove these lines (and optionally the ones below).

volume-coupled-flow/precice-config.xml
```diff
@@ -0,0 +1,46 @@
...
+    <mesh name="Fluid-Mesh">
+      <use-data name="Velocity" />
+    </mesh>
+
+    <mesh name="Dummy-Mesh">
+      <use-data name="Velocity" />
+    </mesh>
+
+    <participant name="Fluid">
+      <use-mesh name="Fluid-Mesh" provide="yes" />
```
MakisH, line 22: Also needs to be ported to preCICE v3.

MakisH (commented):
Some first comments for the next iteration. Keep it rolling!

volume-coupled-flow/fluid-openfoam/0/U_vol
tirgendetwas: It is required, I'm guessing to give the variable an initial state

volume-coupled-flow/fluid-openfoam/system/controlDict
tirgendetwas: Why not? This is the solver we have been experimenting with so far

volume-coupled-flow/dummy-nutils/dummy.py
tirgendetwas: I thought that participant is so basic that it couldn't really be described in a better way

volume-coupled-flow/fluid-openfoam/system/controlDict
MakisH: Since we are not simulating any heat transfer (zero temperature gradients in the domain), pimpleFoam would have been enough. Not really an issue, it just looks strange. I think I will change the case to use pimpleFoam, instead.

volume-coupled-flow/dummy-nutils/dummy.py
MakisH: Good point... In other cases, the participant name is based on the domain. I will rename this to "Source", I think this will be clearer.

volume-coupled-flow/fluid-openfoam/system/controlDict
tirgendetwas: Ok. But I think the case looks different in pimpleFoam, so the screenshot in the README has to change as well.

MakisH: Then let's keep the buoyantPimpleFoam, as it also can be a testbed/starting point for writing temperature.

MakisH:
Status: This case currently stumbles upon this preCICE assertion on the Nutils side:

```
---[precice]  iteration: 1, time-window: 201, time: 1 of 4, time-window-size: 0.005, max-time-step-siz
e: 0.005, ongoing: yes, time-window-complete: yes, 
ASSERTION FAILED
Location:   void precice::cplscheme::BaseCouplingScheme::sendData(const PtrM2N&, const DataMap&, bool)
File:       /home/gc/repos/precice/precice/src/cplscheme/BaseCouplingScheme.cpp:119
Expression: math::equals(timesAscending(nTimeSteps - 1), time::Storage::WINDOW_END)
Rank:       0
Arguments:  
  0: timesAscending(nTimeSteps - 1) == 0.9999999999999787
  1: nTimeSteps == 2
```

which is discussed in https://github.com/precice/precice/issues/1751.

Since:

- The case was previously running correctly with preCICE v2.5.0
- The results we obtain till it hangs (at t=0.35) look reasonable
- It could be a test case to evaluate if the original issue is resolved (@BenjaminRodenberg)

I suggest merging this already.

MakisH (approved):

Issue #354, opened 2023-08-02
openfoam adapter writes as root


valentin-seitz:
invoking the systemtests, currently the docker container of the openfoam adapter writes some files with root file permission. 
This should be fixed by mapping the current UID and GID into the container.

MakisH:
Closed by #355.

PR #355, opened 2023-08-02
Fixed Docker compose root user problem #354

valentin-seitz:
This solves #354

Solution: 
- first get the User & Group ID of the current user via a subprocess call
- write out a .env file 
- docker compose will then use this and substitute ${UID} and ${GID}

It did not break the action:  see https://github.com/precice/tutorials/actions/runs/5740269267


MakisH (approved):
The changes make sense.
I tried running the tests locally, and I can now remove the `./run/` directory as my regular user.

Issue #356, opened 2023-08-03
Automatically delete any previous test `runs/` directories


MakisH:
Currently, running the system tests manually generates a `runs/` directory. When running the tests again, the tests complain.

I suggest removing this directory automatically if it already exists, or alternatively having time-stamped subdirectories.

The directory should also be specified in the `.gitignore` file.

valentin-seitz:
I Think for the Github Action its necessary to have it not automatically removed. So i would prefer the timestamp option.

MakisH:
Closed by #358.

PR #358, opened 2023-08-06
Added Timestamps to not override the folder

valentin-seitz:
In #356 it became obvious that we need to delete the folders if they have the same names as the script will fail. 
I opted against deletion because we want to be able to have them as artifacts later on. 
This PR solves #356 by introducing a timestamp to be added to the foldername. 

It also introduced a small .gitignore file to be places in the `run_directory` of the systemtests to not have 500 changed files show up in git.


MakisH (approved):
Works perfectly! :ok_hand:

PR #362, opened 2023-08-09
Fixed docker user problems in openfoam-adapter

### LINKED ISSUE #359: Systemtests: OpenFOAM adapter is not actually loaded
Looking at the results of the `flow-over-heated-plate/fluid-openfoam` case, they looked strange. Then, looking at the `fluid-openfoam.log`, at the beginning of the solver execution, the issue becomes clear:

```text
--> FOAM Warning : 
    Could not load "libpreciceAdapterFunctionObject.so"
libpreciceAdapterFunctionObject.so: cannot open shared object file: No such file or directory
--> FOAM Warning : 
Unknown function type preciceAdapterFunctionObject

Valid function types :
8(ObukhovLength patchProbes probes psiReactionThermoMoleFractions rhoReactionThermoMoleFractions setTimeStepFaRegion sets surfaces)


    From static Foam::autoPtr<Foam::functionObject> Foam::functionObject::New(const Foam::word&, const Foam::Time&, const Foam::dictionary&)
    in file db/functionObjects/functionObject/functionObject.C at line 129.
--> loading function object 'preCICE_Adapter'
```

This typically means that OpenFOAM cannot find the adapter library where it is supposed to be. The expected path is `$FOAM_USER_LIBBIN`, which depends on the current user and the OpenFOAM version (e.g., on my system right now, it is `~/OpenFOAM/gc-v2306/platforms/linux64GccDPInt32Opt/lib`).

@valentin-seitz Any first clue on what could be going wrong? I have not started debugging yet.

valentin-seitz:
This PR fixes #359 

I removed the user mapping and modified the corresponding docker file to use the user `USER precice` which will then also be used to run the container

### DIFF
--- a/tools/tests/component-templates/openfoam-adapter.yaml
+++ b/tools/tests/component-templates/openfoam-adapter.yaml
@@ -4,13 +4,10 @@ build:
     {% for key, value in build_arguments.items() %}
       - {{key}}={{value}}
     {% endfor %}
-user: ${UID}:${GID}
 depends_on:    
   prepare:
     condition: service_completed_successfully
 volumes:
-  - /etc/passwd:/etc/passwd:ro
-  - /etc/group:/etc/group:ro
   - {{ run_directory }}:/runs
 command: >
   /bin/bash -c "id && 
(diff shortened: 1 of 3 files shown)


MakisH (approved):
The changes make sense and it works on my machine. Thanks! :rocket:

PR #364, opened 2023-08-13
Have one Dockerfile instead of multiples 

valentin-seitz:
As discussed in https://github.com/precice/precice/issues/1700 this PR introduces a small, but working protoype of the ansatz described in https://github.com/precice/precice/issues/1700 using just one Dockerfile

To execute the test use `python systemtests.py --suites=openfoam-adapter-release`
or `python systemtests.py --suites=calculix_test`

It currently supports the following adapters:
- fenics
- nutils
- openfoam
- calculix

But this approach is more flexible as we can easily "install" a certain precice/precice version by using the `PRECICE_REF` build argument without the need of publishing a tag to a docker registry first.


tools/tests/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,159 @@
+FROM ubuntu:22.04 as base_image
+USER root
+SHELL ["/bin/bash", "-c"]
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PATH="${PATH}:/home/precice/.local/bin:/opt/precice/installation/bin"
```
MakisH, line 5: What is the `/home/precice/.local/bin` for? I understand from the rest that the goal is to have everything under `/opt/`. Same in the `openfoam_adapter` step, and maybe elsewhere.

And do we really need the `./installation/` in the name? That's a bit unusual.

tools/tests/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+        libxml2-dev \
+        lintian \
+        lsb-release \
+        petsc-dev \
+        python3-dev \
+        python3-numpy \
+        python3-pip \
+        python3-venv \
+        pkg-config \
+        wget
```
MakisH, line 42: I think we don't need all these dependencies (e.g., `ccache lcov lintian`). These are mostly for other CI tools.

See the list in https://precice.org/installation-source-dependencies.html#ubuntu-2204-lts-jammy-jellyfish

tools/tests/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+USER precice
+RUN python3 -m pip install --user --upgrade pip
+
+
+FROM precice_dependecies as precice
+# Install & build precice into /opt/precice/installation
+ARG PRECICE_REF=main
+USER precice
+WORKDIR /opt/precice
+RUN git clone https://github.com/precice/precice.git -b ${PRECICE_REF} precice && \
```
MakisH, line 52: We could save some space and time by only fetching the last commit (with `--depth=1`, see https://github.com/precice/vm/blob/a37b13a50ade7658e63bd17bbe3230df085bba7f/provisioning/install-precice.sh#L11). You can generally copy from the provisioning scripts of the VM for many components.

tools/tests/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+    cd precice && \
+    mkdir build && cd build &&\
+    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/precice/installation -DPRECICE_PETScMapping=OFF -DBUILD_TESTING=OFF && \
+    make all install -j $(nproc)
+
+FROM precice_dependecies as openfoam_adapter
+ARG OPENFOAM_EXECUTABLE=openfoam2112
```
MakisH, line 59:
```suggestion
ARG OPENFOAM_EXECUTABLE=openfoam2306
```

tools/tests/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+
+FROM precice_dependecies as python_bindings
+COPY --from=precice /opt/precice/installation /opt/precice/installation
+ENV PATH="${PATH}:/home/precice/.local/bin:/opt/precice/installation/bin"
+ENV LD_LIBRARY_PATH="/opt/precice/installation/lib:${LD_LIBRARY_PATH}"
+ENV CPATH="/opt/precice/installation/include:$CPATH"
+# Enable detection with pkg-config and CMake
+ENV PKG_CONFIG_PATH="/opt/precice/installation/lib/pkgconfig:$PKG_CONFIG_PATH"
+ENV CMAKE_PREFIX_PATH="/opt/precice/installation:$CMAKE_PREFIX_PATH"
+ENV PYTHONPATH="/opt/precice/installation/python_bindings:$PYTHONPATH"
```
MakisH, line 92: I don't like that we are duplicating these paths everywhere. Can we just have an environment file to load?

MakisH (commented):
Looks good generally, how/what should I test?

What is the problem with CalculiX? The steps look very similar to the VM provisioning script, which works: https://github.com/precice/vm/blob/develop/provisioning/install-calculix.sh

I guess you probably stumbled upon the GCC-related issue, since you are basing on Ubuntu 22.04, and the VM on 20.04: https://precice.org/adapter-calculix-get-adapter.html#compiling-with-gcc-10-or-newer

Here is how we solve this when preparing Debian packages: https://github.com/precice/calculix-adapter/blob/9057470c6b7018a28895c24c964f65364a7fbd04/.github/workflows/ubuntu_build.yml#L16

MakisH:
Closes https://github.com/precice/openfoam-adapter/pull/184

tools/tests/dockerfiles/Dockerfile
valentin-seitz: Ah shit! Forgot to do it here! 
Thanks!

tools/tests/dockerfiles/Dockerfile
valentin-seitz: I also added some from other containers, to ensure everything goes smoothly. I will remove them and see if it its still worky :)

MakisH: Removing would help both with build time and with size, so let's not install packages we don't really need.

valentin-seitz:
@MakisH if you could maybe have a quick look if it also runs smoothly on your machine? 

Mainly `python systemtests.py --suites=calculix_test` 
and maybe `python systemtests.py --suites=fenics_test`

tools/tests/components.yaml
```diff
@@ -2,39 +2,77 @@ openfoam-adapter:
...
+      default: "main"
+    PLATFORM:
+      description: Dockerfile platform used
+      default: "ubuntu_2204"
+    TUTORIALS_REF:
+      description: Tutorial git reference to use
+      default: "master"
     OPENFOAM_EXECUTABLE:
-      options: ["openfoam2112"]
+      options: ["openfoam2306","openfoam2212",openfoam2112]
```
MakisH, line 15: I guess we need the quotes everywhere:

```suggestion
      options: ["openfoam2306","openfoam2212","openfoam2112"]
```

flow-over-heated-plate/reference_results.md
```diff
@@ -8,14 +8,22 @@ This file contains an overview of the results over the reference results as well
```
MakisH, line 8: Would it make sense to rename these files from `*.md` to `*.metadata`?
This way, when someone does a find-all & replace, they could easily skip these.

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+RUN pip3 install --user fenics-ufl
+ARG FENICS_ADAPTER_REF
+# Building fenics-adapter
+RUN pip3 install --user git+https://github.com/precice/fenics-adapter.git@${FENICS_ADAPTER_REF}
+
+
+FROM precice_dependecies as nutils_adapter
+COPY --from=python_bindings /home/precice/.local /home/precice/.local
+USER precice
+# Building fenics-adapter
```
MakisH, line 96:
```suggestion
# Installing nutils - There is no adapter
```

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+WORKDIR /home/precice
+RUN wget http://www.dhondt.de/ccx_${CALULIX_VERSION}.src.tar.bz2 && \
+    tar xvjf ccx_${CALULIX_VERSION}.src.tar.bz2 && \
+    rm -fv ccx_${CALULIX_VERSION}.src.tar.bz2
+
+ARG CALULIX_ADAPTER_REF
+WORKDIR /home/precice
+RUN git clone --depth 1 --branch ${CALULIX_ADAPTER_REF} https://github.com/precice/calculix-adapter.git && \
+    cd calculix-adapter && \
+    make CXX_VERSION=${CALULIX_VERSION} ADDITIONAL_FFLAGS="-std=legacy" -j $(nproc) && \
```
MakisH, line 117: Is the `-std=legacy` a better alternative that we should document in https://precice.org/adapter-calculix-get-adapter.html#compiling-with-gcc-10-or-newer ?
Do we already have this somewhere, or did you have to solve it yourself?

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+    mkdir build && cd build &&\
+    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/home/precice/.local/ -DPRECICE_PETScMapping=OFF -DBUILD_TESTING=OFF && \
+    make all install -j $(nproc)
+
+FROM precice_dependecies as openfoam_adapter
+ARG OPENFOAM_EXECUTABLE
+USER root
+RUN apt-get update &&\
+    wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
+    apt-get -qq install ${OPENFOAM_EXECUTABLE}-dev &&\
```
MakisH, line 59: I don't really like that the `OPENFOAM_EXECUTABLE` is misused as a package name, but I understand that this is significantly simpler. For the cases we currently cover, this is fine.

MakisH (commented):
> @MakisH if you could maybe have a quick look if it also runs smoothly on your machine?
> 
> Mainly `python systemtests.py --suites=calculix_test` and maybe `python systemtests.py --suites=fenics_test`

The `calculix_test` times out:

```shell
~/repos/precice/tutorials [chonky-dockerfile]$ python tools/tests/systemtests.py --suites=calculix_test
Using log-level: INFO
INFO: About to run the following systemtest in the directory /home/gc/repos/precice/tutorials/runs:
 [Perpendicular flap (fluid-openfoam, solid-calculix)]
INFO: Started running Perpendicular flap (fluid-openfoam, solid-calculix),  0/1
CRITICAL: Systemtest Perpendicular flap (fluid-openfoam, solid-calculix) had serious issues executin the docker compose command about to kill the docker compose command. Please check the logs! Command '['docker', 'compose', '--file', 'docker-compose.tutorial.yaml', 'up', '--build']' timed out after 360 seconds
CRITICAL: Docker Compose failed, skipping fieldcompare and writing stdout.log and stderr.log
INFO: Running Perpendicular flap (fluid-openfoam, solid-calculix) took 925.802237306998 seconds
ERROR: Failed to run Perpendicular flap (fluid-openfoam, solid-calculix)
```

even though the run completes and both solvers exit with zero:

[stderr.log](https://github.com/precice/tutorials/files/12505660/stderr.log)
[stdout.log](https://github.com/precice/tutorials/files/12505661/stdout.log)

The `fenics_test` succeeded:

```shell
~/repos/precice/tutorials [chonky-dockerfile]$ python tools/tests/systemtests.py --suites=fenics_test
Using log-level: INFO
INFO: About to run the following systemtest in the directory /home/gc/repos/precice/tutorials/runs:
 [Flow over heated plate (fluid-openfoam, solid-fenics)]
INFO: Started running Flow over heated plate (fluid-openfoam, solid-fenics),  0/1
INFO: Running Flow over heated plate (fluid-openfoam, solid-fenics) took 269.88722395299556 seconds
INFO: Success running Flow over heated plate (fluid-openfoam, solid-fenics)
```

However, in contrast to the failed case, I don't see any `stderr.log` and `stdout.log` files in the archive. These would always be useful (to check that indeed the test did what it was supposed to), especially since we now only print the overview.

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
valentin-seitz: This is i guess similar to the error you made me aware of. But i just forgot about that hint. So i solved it myself. I think i can adopt to your proposed solution. It should be the same error. :)

flow-over-heated-plate/reference_results.md
valentin-seitz: Good idea. I will change the script accordingly :)

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
MakisH: I am not asking to change your solution. In fact, it looks like it may be a better solution. But then, let's also document it!

valentin-seitz: I think its better to have the most specific solution to the problem in this case. 
As [already discussed](https://groups.google.com/g/comp.lang.fortran/c/vITMz5e1nHQ/m/8j5--lt9AAAJ) `-std=legacy` might imply other, not really expected things.

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
valentin-seitz: Me neither, but i cannot really think of a very good alternative for now.

valentin-seitz:
> The `calculix_test` seems to be a false negative (it fails after a long time, but it seems to work actually):

Yes and no :)
The script will also fail once it times out. Since you probably had to built the calculix image from scratch it probably timed out. 
The newest version of the systemtests addresses it such that the actual solver time is the one that has a timeout as a deadlock is way more common there. (if one participant segfaults etc)

> However, in contrast to the failed case, I don't see any stderr.log and stdout.log files in the archive. These would always be useful (to check that indeed the test did what it was supposed to), especially since we now only print the overview.

Is added, we now write always the logs into the .log files

MakisH:
Works now, great job!

```
~/repos/precice/tutorials/tools/tests [develop]$ python systemtests.py --suites=calculix_test
Using log-level: INFO
INFO: About to run the following systemtest in the directory /home/gc/repos/precice/tutorials/runs:
 [Perpendicular flap (fluid-openfoam, solid-calculix)]
INFO: Started running Perpendicular flap (fluid-openfoam, solid-calculix),  0/1
INFO: Running Perpendicular flap (fluid-openfoam, solid-calculix) took 191.29625335900005 seconds
INFO: Success running Perpendicular flap (fluid-openfoam, solid-calculix)
+-------------------------------------------------------+---------+-------------------+-----------------+-----------------------+
| systemtest                                            | success | building time [s] | solver time [s] | fieldcompare time [s] |
+-------------------------------------------------------+---------+-------------------+-----------------+-----------------------+
| Perpendicular flap (fluid-openfoam, solid-calculix)   |    1    |       71.39       |     104.51      |         15.22         |
+-------------------------------------------------------+---------+-------------------+-----------------+-----------------------+
```

PR #365, opened 2023-08-14
Add precice-events to the cleaning scripts

MakisH:
Deletes the `./precice-events/` directory in each case.

Also removes the previously specified `precice-*-events.json` and `precice-*-events-summary.log`, as well as the relic `precice-postProcessingInfo.log`.

### DIFF
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -30,6 +30,7 @@ clean_precice_logs() {
             ./precice-*-convergence.log \
             ./precice-*-watchpoint-*.log \
             ./precice-*-watchintegral-*.log \
+            ./events.json \
             ./core
         rm -rfv ./precice-events/
     )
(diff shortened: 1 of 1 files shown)


tools/cleaning-tools.sh
```diff
@@ -28,12 +28,10 @@ clean_precice_logs() {
         echo "---- Cleaning up preCICE logs in $(pwd)"
         rm -fv ./precice-*-iterations.log \
             ./precice-*-convergence.log \
-            ./precice-*-events.json \
-            ./precice-*-events-summary.log \
-            ./precice-postProcessingInfo.log \
             ./precice-*-watchpoint-*.log \
             ./precice-*-watchintegral-*.log \
             ./core
+        rm -rfv ./precice-events/
```
davidscn, line 34: That's correct in principle. If people start using the generated `precice-events`, e.g., `precice-events merge <...>`, a file called `events.json` will be generated. I would consider to also delete this file, though this is a bit of a philosophical question. There are even further post-processing facilities generating files in the `precice-events`, but deleting the aggregated json should be fine.

MakisH: Good catch! Added `events.json` as well.

PR #370, opened 2023-08-18
Port nutils cases to preCICE v3

### LINKED ISSUE #369: Compatibility of Nutils cases with preCICE v3
Already ported:

- [channel-transport](https://github.com/precice/tutorials/tree/develop/channel-transport)
- [volume-coupled-flow](https://github.com/precice/tutorials/tree/develop/volume-coupled-flow)

Still missing:

- [x] flow-over-heated-plate
- [x] partitioned-heat-conduction
- [x] partitioned-heat-conduction-direct
- [x] perpendicular-flap

I will work on this.

MakisH:
Closes #369 

@IshaanDesai could you please check the diff and check that it makes sense? Not having a compiler to watch my back stresses me a bit.

@davidscn I got a bit confused with the following regarding direct mesh access:

- I need to call `initialize()` before I can access the region, but
- I now need to write data to the region before I call `initialize()`, since we have merged it with `initialize_data()`.

What do I misunderstand?

Please have a look at the `partitioned-heat-conduction-direct` case. Does the order of calls around initialization make sense?

All cases work on my system and the results look reasonable.

### DIFF
--- a/partitioned-heat-conduction-direct/nutils/heat.py
+++ b/partitioned-heat-conduction-direct/nutils/heat.py
@@ -58,10 +58,6 @@ def main(side='Dirichlet', n=10, degree=1, timestep=.1, alpha=3., beta=1.3):
     vertex_ids_read = participant.set_mesh_vertices(mesh_name_read, read_sample.eval(ns.x))
     participant.set_mesh_access_region(mesh_name_write, [.9, 1.1, -.1, 1.1])
 
-    # write initial data
-    if participant.requires_initial_data():
-        precice_write(write_sample.eval(0.))
-
     participant.initialize()
     precice_dt = participant.get_max_time_step_size()
     dt = min(timestep, precice_dt)
--- a/partitioned-heat-conduction-direct/nutils/heat.py
+++ b/partitioned-heat-conduction-direct/nutils/heat.py
@@ -50,22 +50,31 @@ def main(side='Dirichlet', n=10, degree=1, timestep=.1, alpha=3., beta=1.3):
         res += read_sample.integral('basis_n readfunc d:x' @ ns)
 
     # preCICE setup
-    interface = precice.Interface(side, "../precice-config.xml", 0, 1)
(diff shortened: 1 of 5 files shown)


partitioned-heat-conduction-direct/nutils/heat.py
```diff
@@ -50,22 +50,31 @@ def main(side='Dirichlet', n=10, degree=1, timestep=.1, alpha=3., beta=1.3):
...
 
-    vertex_ids_read = interface.set_mesh_vertices(mesh_id_read, read_sample.eval(ns.x))
-    interface.set_mesh_access_region(mesh_id_write, [.9, 1.1, -.1, 1.1])
+    vertex_ids_read = participant.set_mesh_vertices(mesh_name_read, read_sample.eval(ns.x))
+    participant.set_mesh_access_region(mesh_name_write, [.9, 1.1, -.1, 1.1])
 
-    precice_dt = interface.initialize()
+    # write initial data
+    if participant.requires_initial_data():
+        precice_write(write_sample.eval(0.))
```
davidscn, line 63: I think these lines need to be removed. I understand that our tutorials want and should offer some kind of generality, but the initial data is not used in this setup https://github.com/precice/tutorials/blob/0482e5b4d3e1604dd031d2b113a2aaa63733c9a8/partitioned-heat-conduction-direct/precice-config.xml#L45-L46

In fact, using the direct access with initial data is not possible anymore, see https://github.com/precice/precice/issues/1583
When configuring initial-data in this example, (add `initialize="true"` in the exchange tag), preCICE should complain about it. You can also keep for educational reasons to showcase that initial data writing is not possible in this context.

IshaanDesai (approved):
I went through all the changes but did not run the cases. Everything looks to be in order :+1:

partitioned-heat-conduction-direct/nutils/heat.py
MakisH: Thanks for the clarification! Removed the data initialization.

PR #375, opened 2023-09-30
Port two-scale-heat-conduction case from DuMuX-Adapter

Fujikawas:
- [ ] Port the case to DuMuX-Adapter new version for precice V3.0.0;
- [ ] Rearrange the folder structure to cooperate with the Nutil-Nutil coupling;
- [ ] Set the environment to run the case directly from current folder;
- [ ] Update relevant documentations and guides;


uekerman:
replaced by #376

PR #379, opened 2023-10-04
Use beta consistent with literature

BenjaminRodenberg:
Closes https://github.com/precice/precice/commit/b8ba60bf8b6034d676aeec9e73bb366bebcfe848

Comments for the review: Am I missing some tutorial?

### DIFF
--- a/partitioned-heat-conduction-complex/fenics/heat.py
+++ b/partitioned-heat-conduction-complex/fenics/heat.py
@@ -72,7 +72,7 @@ fenics_dt = .1
 # Error is bounded by coupling accuracy. In theory we can obtain the analytical solution.
 error_tol = 10 ** -6
 alpha = 3  # parameter alpha
-beta = 1.3  # parameter beta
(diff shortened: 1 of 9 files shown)


IshaanDesai (approved):
I think you have got all the cases.

davidscn (commented):
Hm, the proposed changes here contradict what we do in the OpenFOAM adapter paper. There, we use 1.3 and even refer to the tutorials

> We select α = 3 and β = 1.3, as used in the preCICE tutorials with other solvers

We selected these parameters due to their value in the tutorial. @MakisH opinions?

MakisH:
I remember discussing this a while ago. I think it was wrong, but kept it to stick with the tutorials.

In the OpenFOAM-preCICE paper, we refer to a specific version of the tutorials, so I think we should not be restricted by that. Consistency is more important.

BenjaminRodenberg:
Ok. So approved? @MakisH @davidscn

MakisH (approved):

PR #382, opened 2023-10-06
Use `mapping:nearest-neighbor` in partitioned heat conduction

BenjaminRodenberg:
We should use nearest neighbor mapping here, because this is sufficient for this simple case with matching meshes. Also consistent with the description in the OpenFOAM adapter paper:

> Figure 6. ... The coupled scenario uses a nearest-neighbor mapping with matching interface nodes.


IshaanDesai (approved):

MakisH (approved):
Totally! And I think we may have more such cases where we use RBF while we don't really need it.

PR #383, opened 2023-10-10
Fix update of Dirichlet BC and RHS

BenjaminRodenberg:
Fixes bug, if time step changes inside window. See https://github.com/precice/fenics-adapter/issues/20.

For the purpose of documentation, if will add some comments in the diff of this PR.

With this bugfix the waveform version of the tutorial provided via #281 also works for non-matching subcycling, like described in https://github.com/precice/fenics-adapter/issues/20.

### DIFF
--- a/partitioned-heat-conduction-complex/fenics/heat.py
+++ b/partitioned-heat-conduction-complex/fenics/heat.py
@@ -178,11 +178,6 @@ ranks << mesh_rank
 error_total, error_pointwise = compute_errors(u_n, u_ref, V)
 error_out << error_pointwise
 
-# set t_1 = t_0 + dt, this gives u_D^1
-# call dt(0) to evaluate FEniCS Constant. Todo: is there a better way?
-u_D.t = t + dt(0)
-f.t = t + dt(0)
-
 flux = Function(V_g)
 flux.rename("Flux", "")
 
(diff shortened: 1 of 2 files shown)


partitioned-heat-conduction-complex/fenics/heat.py
```diff
@@ -239,9 +241,5 @@ def determine_gradient(V_g, u, flux):
         ref_out << u_ref
         error_out << error_pointwise
 
-    # Update Dirichlet BC
-    u_D.t = t + float(dt)
-    f.t = t + float(dt)
```
BenjaminRodenberg, line 244: This `dt` is actually the `dt` from the timestep that was just completed, not the timestep that one is going to compute next. But the `u_D` and `f` here is used for the future timestep. Meaning: If `dt` changes from one timestep to the next, `u_D` and `f` will use the `dt` corresponding to the last time step and not corresponding to the next.

partitioned-heat-conduction-complex/fenics/heat.py
```diff
@@ -194,7 +189,14 @@ def determine_gradient(V_g, u, flux):
 
     precice_dt = precice.get_max_time_step_size()
     dt.assign(np.min([fenics_dt, precice_dt]))
+
+    # Dirichlet BC and RHS need to point to end of current timestep
+    u_D.t = t + float(dt)
+    f.t = t + float(dt)
+
+    # Coupling BC needs to point to end of current timestep
     read_data = precice.read_data(dt)
```
BenjaminRodenberg, line 198: As a general advice, one should set all boundary conditions ("real" boundary conditions, RHS and coupling boundary conditions) before calling `solve` and before calling `advance`.

BenjaminRodenberg (commented):
Explaining the pitfall that has lead to this bug in the comments below.

BenjaminRodenberg:
@NiklasKotarsky and @uekerman might be interesting for you. I think this also nicely illustrates how `get_max_time_step_size()` improves the API, because before it was much harder to not get confused with the `dt` returned by preCICE.

PR #386, opened 2023-10-25
Sympy expressions in partitioned-heat-conduction(-complex)

NiklasVin:
I added in both examples Sympy expressions in the heat.py files and replaced the string representations of `u_D`, `f_N` and `f` with them.
If you want to change the analytical solution `u_D`, you now don't need to change anything else despite `u_D`. 
Before, it was necessary to change `f_N` and `f` but the expressions of both are now deduced from `u_D` automatically.


BenjaminRodenberg:
Looks good! Thanks @NiklasVin for the contribution!

BenjaminRodenberg (approved):

PR #388, opened 2023-11-03
Oscillator: Add higher-order time stepping schemes.

BenjaminRodenberg:
* Add Runge Kutta 4 and Radau IIA scheme
* Use subcycling to achieve higher order

Moved some features from my research repository https://github.com/BenjaminRodenberg/oscillator-example over here. Question is similar to #30: Do we want to have these more advanced settings here or rather keep this PR until we have a proper solution?


oscillator/python/timeSteppers.py
```diff
@@ -0,0 +1,159 @@
...
+
+class RadauIIA():
+    def __init__(self, ode_system) -> None:
+        self.ode_system = ode_system
+        pass
+
+    def rhs_eval_points(self, dt) -> List[float]:
+        return np.linspace(0, dt, 5)  # will create an interpolant from this later
+
+    def do_step(self, u, v, a, f, dt) -> Tuple[float, float, float]:
```
BenjaminRodenberg, line 125: Instead of `f` being an array of rhs values at the rhs evaluation points, we could also request a lambda function `f`. This makes `oscillator.py` slightly more complicated, but also allows us to skip the interpolation below by just wrapping `precice.read_data`

oscillator/python/oscillator.py
```diff
@@ -78,7 +80,7 @@ class Participant(Enum):
 
 participant.initialize()
 precice_dt = participant.get_max_time_step_size()
-my_dt = precice_dt  # use my_dt < precice_dt for subcycling
+my_dt = precice_dt / 4  # use my_dt < precice_dt for subcycling
```
uekerman, line 83: Shouldn't we document this in the README?

uekerman (approved):
Looks like a meaningful addition to me. For me, the oscillator is anyway more a numerical testcase for things in preCICE than a case that users copy to learn an adapter etc. So, having higher-order time integrators here is beneficial also in the long run.
Shouldn't we add something in the README about the choices?

PR #390, opened 2023-11-09
Bugfix: Remove stray mesh from flow-over-heated-plate-two-meshes

MakisH:
This was making preCICE complain when the Fluid participant was running in parallel.

See https://precice.discourse.group/t/problem-to-run-openfoam-in-parallel/1652

Previously:

precice-config


uekerman (approved):
Correct.

PR #381, opened 2023-10-06
Partitioned heat conduction with Schwarz-type domain decomposition

BenjaminRodenberg:
Adds a tutorial illustrating how Schwarz-type domain decomposition could be implemented with preCICE.

General question similar to #30: How can we publish these kind of academic and advanced tutorials in a proper way?

PR #391, opened 2023-11-15
Use overlapping schwarz for oscillator tutorial

BenjaminRodenberg:
Exchange displacement instead of forces. This leads to identical results, but makes the modelling a bit clearer and fits better into the domain decomposition framework.

What we did before:

Screenshot from 2023-11-15 22-01-49

What we do now:

Screenshot from 2023-11-15 22-02-31

Inside of each participant, we still translate the displacement of the other participant to a force, because this fits better into the generalized alpha framework. This, however, is an implementation detail of the solver and does not really interfere with the coupling scheme.

### DIFF
--- a/oscillator/README.md
+++ b/oscillator/README.md
@@ -15,7 +15,7 @@ This tutorial solves a simple mass-spring oscillator with two masses and three s
 
 ![Schematic drawing of oscillator example](images/tutorials-oscillator-schematic-drawing.png)
 
-Note that this case applies a Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We will refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`. For more information, please refer to [1].
+Note that this case applies an overlapping Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We will refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`. For more information, please refer to [1].
 
 ## Available solvers
 
--- a/oscillator/precice-config.xml
+++ b/oscillator/precice-config.xml
@@ -7,30 +7,30 @@
     <sink filter="%Severity% > debug" />
   </log>
 
-  <data:scalar name="Force-Left"  />
-  <data:scalar name="Force-Right"  />
+  <data:scalar name="Displacement-Left"  />
+  <data:scalar name="Displacement-Right"  />
 
   <mesh name="Mass-Left-Mesh" dimensions="2">
-    <use-data name="Force-Left" />
-    <use-data name="Force-Right" />
+    <use-data name="Displacement-Left" />
+    <use-data name="Displacement-Right" />
   </mesh>
 
   <mesh name="Mass-Right-Mesh" dimensions="2">
-    <use-data name="Force-Left" />
-    <use-data name="Force-Right" />
+    <use-data name="Displacement-Left" />
+    <use-data name="Displacement-Right" />
   </mesh>
 
   <participant name="Mass-Left">
     <provide-mesh name="Mass-Left-Mesh" />
-    <write-data name="Force-Left" mesh="Mass-Left-Mesh" />
-    <read-data  name="Force-Right" mesh="Mass-Left-Mesh" />
+    <write-data name="Displacement-Left" mesh="Mass-Left-Mesh" />
+    <read-data  name="Displacement-Right" mesh="Mass-Left-Mesh" />
   </participant>
 
   <participant name="Mass-Right">
     <receive-mesh name="Mass-Left-Mesh" from="Mass-Left"/>
     <provide-mesh name="Mass-Right-Mesh" />
-    <write-data name="Force-Right" mesh="Mass-Right-Mesh" />
-    <read-data  name="Force-Left" mesh="Mass-Right-Mesh" />
+    <write-data name="Displacement-Right" mesh="Mass-Right-Mesh" />
+    <read-data  name="Displacement-Left" mesh="Mass-Right-Mesh" />
     <mapping:nearest-neighbor direction="write" from="Mass-Right-Mesh" to="Mass-Left-Mesh" constraint="consistent" />
     <mapping:nearest-neighbor direction="read"  from="Mass-Left-Mesh" to="Mass-Right-Mesh" constraint="consistent" />
   </participant>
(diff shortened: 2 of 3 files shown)


oscillator/README.md
```diff
@@ -15,7 +15,7 @@ This tutorial solves a simple mass-spring oscillator with two masses and three s
 
 ![Schematic drawing of oscillator example](images/tutorials-oscillator-schematic-drawing.png)
 
-Note that this case applies a Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We will refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`. For more information, please refer to [1].
+Note that this case applies an overlapping Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We will refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`. For more information, please refer to [1].
```
uekerman, line 18: Is it really overlapping?

BenjaminRodenberg:
Decision after discussion with @uekerman: Similar to https://github.com/precice/tutorials/pull/381 this modification of the case should go into `tutorials/oscillator-schwarz`. This will result in some amount of code duplication, but there are the following arguments for keeping both version of the tutorial:

* #322 uses the coupling scheme that exchanges forces, not displacements and would therefore require a larger update
* Providing examples where the same scenario is coupled with different coupling schemes is generally valuable

I will also modify #388 to work with the overlapping schwarz-version of the tutorial, because this is also the case I'm investigating for my thesis.

oscillator/README.md
BenjaminRodenberg: Yes. $k_{12}$ is duplicated and this makes the region between $[u_1, u_2]$ the overlap region. As we are talking about an ODE here and actually all DOFs are part of the overlap region now, the whole situation is a bit on the border, but I think for illustrative purposes everything is fine.

MakisH:
> Decision after discussion with @uekerman: Similar to #381 this modification of the case should go into `tutorials/oscillator-schwarz`. This will result in some amount of code duplication, but there are the following arguments for keeping both version of the tutorial:

Isn't the "overlapping" part easier to understand from the top-view? So, `tutorials/oscillator-overlap`?

BenjaminRodenberg:
> > Decision after discussion with @uekerman: Similar to #381 this modification of the case should go into `tutorials/oscillator-schwarz`. This will result in some amount of code duplication, but there are the following arguments for keeping both version of the tutorial:
> 
> Isn't the "overlapping" part easier to understand from the top-view? So, `tutorials/oscillator-overlap`?

Probably yes. I think "Schwarz" is also a broader term and might refer to completely different setups. With "Overlap" the main feature is clear.

oscillator/python/__pycache__/problemDefinition.cpython-38.pyc
MakisH, line 1: 🚨

This folder should not be here.

oscillator/precice-config.xml.orig
```diff
@@ -0,0 +1,92 @@
...
+      from="Mass-Left"
+      to="Mass-Right"
+      initialize="true" />
+    <exchange
+      data="Force-Right"
+      mesh="Mass-Left-Mesh"
+      from="Mass-Right"
+      to="Mass-Left"
+      initialize="true" />
+>>>>>>> develop
```
MakisH: 🚨

Merge conflicts still here. Also in more places.

(what is the `.orig` for?)

oscillator/images/tutorials-oscillator-schematic-drawing.aux
```diff
@@ -0,0 +1 @@
+\relax
```
MakisH: 🚨

LaTeX auxiliary files? Thesis side-products? :sweat_smile: 

More files in this folder should not be here.

MakisH (changes requested):
Some cleanup is needed before we continue

Edit: I was somehow looking at an older commit. Abort.

BenjaminRodenberg:
> Some cleanup is needed before we continue
> 
> Edit: I was somehow looking at an older commit. Abort.

Yeah. Felt like a déjà vu to me :) Then all good? Can you :heavy_check_mark: ?

oscillator-overlap/README.md
```diff
@@ -0,0 +1,42 @@
...
+summary: We solve an oscillator with two masses in a partitioned fashion with overlapping domain decomposition. Each mass is solved by an independent ODE.
+---
+
+{% note %}
+Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/oscillator-overlap). Read how in the [tutorials introduction](https://www.precice.org/tutorials.html).
+{% endnote %}
+
+## Setup
+
+This tutorial solves the same problem as `oscillator`, but applies a different domain decomposition strategy. See the `oscillator` case for details on the general setup. The partitioning of the mass-spring system is shown here:
```
uekerman, line 14:
```suggestion
This tutorial solves the same problem as [oscillator](tutorials-oscillator.html), but applies a different domain decomposition strategy. See the [oscillator](tutorials-oscillator.html) case for details on the general setup. The partitioning of the mass-spring system is shown here:
```

oscillator-overlap/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+![Schematic drawing of oscillator example with overlapping domain decomposition](images/tutorials-oscillator-overlap-dd.png)
+
+Note that this case applies an overlapping Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We will refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`. For more information, please refer to [1].
```
uekerman, line 18:
```suggestion
Note that this case applies an overlapping Schwarz-type coupling method and not (like most other tutorials in this repository) a Dirichlet-Neumann coupling. This results in a symmetric setup of the solvers. We refer to the solver computing the trajectory of $m_1$ as `Mass-Left` and to the solver computing the trajectory of $m_2$ as `Mass-Right`.
```
There is no [1] on this page and the paper does also not discuss this case. We could maybe refer to BR's thesis here later.

uekerman (approved):
Name and README is good to go IMO (besides last edits below).
I did not run the case.

oscillator-overlap/python/oscillator.py
```diff
@@ -0,0 +1,190 @@
...
+    write_data_name = 'Displacement-Right'
+    mesh_name = 'Mass-Right-Mesh'
+
+    this_mass = problemDefinition.MassRight
+    this_spring = problemDefinition.SpringRight
+    connecting_spring = problemDefinition.SpringMiddle
+    other_mass = problemDefinition.MassLeft
+
+else:
+    raise Exception(f"wrong participant name: {participant_name}")
```
MakisH, line 53: Same comment as in #381: I know it needs effort to port old tutorials, but for new tutorials, we can already stick to the documented structure we are trying to use everywhere: https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial

Here, I would suggest:

- `solver-python/`
   - `oscillator.py` 
   - `problemDefinition.py`
- `mass-left-python/`
   - `config.py` or `config.json` 
   - `run.sh` (without options)
   - `clean.sh`
- `mass-right-python/`
   - `config.py` or `config.json`
   - `run.sh` (without options)
   - `clean.sh`

oscillator-overlap/README.md
MakisH: This is the tutorials repository, so URLs to webpages will not work when looking at the `README.md` outside the website. That's why we use full links here:

```suggestion
This tutorial solves the same problem as the [oscillator tutorial](https://precice.org/tutorials-oscillator.html), but applies a different domain decomposition strategy. See the oscillator tutorial for details on the general setup. The partitioning of the mass-spring system is shown here:
```

MakisH (changes requested):
I have not tried to run it, but it looks clean. :+1:
I would just really like that, since we already converged in different occasions on why it makes sense that we stick to the standard tutorial structure, that we try to enforce it.

oscillator-overlap/README.md
MakisH: Not a concrete suggestion, but a brainstorming idea: In principle, you can even use references to the literature cards we render on the website. They will not render correctly outside the repository, but we could later find a solution for that.

oscillator-overlap/python/oscillator.py
BenjaminRodenberg: Thanks! Should be fine now.

MakisH (approved):
The new structure makes looks nice! It is fine to not have dedicated config files for such a simple solver. The command-line arguments in the `run.sh` scripts are enough of a configuration.

I have ran the tutorial and I get the following errors:
- Left: `0.01,0.02597530032202172`
- Right: `0.01,0.025058100518940307`

PR #392, opened 2023-11-16
Moving the actual running of the systemtests into the percice-tests vm 

valentin-seitz:
This PR modifies the underlying workflow file to execute the tests on the vm using the LFS mirror locally. 

After some life debugging (as can be seen in the git commits) the systemtests [now run](https://github.com/precice/tutorials/actions/runs/6892775066/job/18750824403) locally with a local LFS repo. 

I also added a flag to disable the default upload of the artifacts: The behavior is now: if the tests fail-> upload; if not->dont. 
Some tutorials generate up to 80M meshes in calculix, so this can be rather big at some point. 

@MakisH Should the documentation of the VM config etc into this PR?


.github/workflows/run_testsuite_manual.yml
```diff
@@ -26,6 +26,14 @@ on:
             - 'WARNING'
             - 'ERROR'
             - 'CRITICAL'
+      upload_artifacts:
+        description: 'Upload artifacts always (not only on failure)'
```
MakisH, line 30:
```suggestion
        description: 'Upload artifacts also on success (not only on failure)'
```

.github/workflows/run_testsuite_workflow.yml
```diff
@@ -20,42 +20,45 @@ on:
...
+        echo "Systemtests branch: ${{ inputs.systests_branch }}"
+        echo "Uploading the runs folder on success: ${{ inputs.upload_artifacts }}"
+    - name: Move LFS URL to local LFS server
+      run: |
+        /home/precice/runners_root/scripts/make_lfs_local.sh
+    - name: 'Cleanup the folder'
+      run: |
+        ls -la ./
+        rm -rf ./* || true
+        rm -rf ./.??* || true
```
MakisH, line 44: What is the `|| true` for? The `-f` should anyway return 0 if there is nothing to delete.

.github/workflows/run_testsuite_workflow.yml
```diff
@@ -20,42 +20,45 @@ on:
...
-    - name: Log directory
-      run: |
-        ls -al
-        pwd
-    - uses: actions/setup-python@v4
-      with:
-        python-version: '3.11'
-    - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v2
     - name: install python depencies
```
MakisH, line 52:
```suggestion
    - name: Install Python dependencies
```

MakisH (approved):
Looks good overall, and apparently it works. We cannot yet check the UI, since the workflow is not yet on master. But we could eventually make a cherry-pick.

.github/workflows/run_testsuite_workflow.yml
valentin-seitz: removed :)

Issue #394, opened 2023-11-18
Test suite with multiple tests only executes last-defined test


MakisH:
When defining multiple tests in a test suite, only the last one is used/printed. For example:

`tests.yaml`:

```yaml
  elastic_tube_1d_test:
    tutorials:
      - path: elastic-tube-1d
        case_combination:
          - fluid-cpp
          - solid-cpp
        reference_result: ./elastic-tube-1d/reference-data/fluid-cpp_solid-cpp.tar.gz
      - path: elastic-tube-1d
        case_combination:
          - fluid-python
          - solid-python
        reference_result: ./elastic-tube-1d/reference-data/fluid-python_solid-python.tar.gz
      - path: elastic-tube-1d
        case_combination:
          - fluid-cpp
          - solid-python
        reference_result: ./elastic-tube-1d/reference-data/fluid-cpp_solid-python.tar.gz
```

`print_test_suites.py`:

```shell
~/repos/precice/tutorials/tools/tests [develop]$ python3 print_test_suites.py 
Using log-level: INFO
Test suite: openfoam_adapter_pr contains:
    Flow over heated plate
        cases: [(fluid-openfoam, solid-openfoam)]
        reference_results: [/home/gc/repos/precice/tutorials/flow-over-heated-plate/reference-data/fluid-openfoam_solid-openfoam.tar.gz]

Test suite: openfoam_adapter_release contains:
    Flow over heated plate
        cases: [(fluid-openfoam, solid-openfoam)]
        reference_results: [/home/gc/repos/precice/tutorials/flow-over-heated-plate/reference-data/fluid-openfoam_solid-openfoam.tar.gz]

Test suite: fenics_test contains:
    Flow over heated plate
        cases: [(fluid-openfoam, solid-fenics)]
        reference_results: [/home/gc/repos/precice/tutorials/flow-over-heated-plate/reference-data/fluid-openfoam_solid-fenics.tar.gz]

Test suite: nutils_test contains:
    Flow over heated plate
        cases: [(fluid-openfoam, solid-nutils)]
        reference_results: [/home/gc/repos/precice/tutorials/flow-over-heated-plate/reference-data/fluid-openfoam_solid-nutils.tar.gz]

Test suite: calculix_test contains:
    Perpendicular flap
        cases: [(fluid-openfoam, solid-calculix)]
        reference_results: [/home/gc/repos/precice/tutorials/perpendicular-flap/reference-data/fluid-openfoam_solid-calculix.tar.gz]

[...]

PR #408, opened 2023-11-21
Fix 394

valentin-seitz:
This PR adresses #394

What happend: During the parsing of the yaml file, the list was emptied every iteration, so only the last iteration survived


MakisH (approved):
Makes sense to me, have not tried it yet though. Let's merge and test everything together.

PR #409, opened 2023-11-23
Remove extrapolation order from config files

davidscn:
..which is no longer supported IIRC


BenjaminRodenberg (approved):
Yes. Extrapolation has been removed: https://github.com/precice/precice/pull/1738

With extraplation, these configuration files should also trigger some error. If not, then we have a bug in preCICE. Is this the case?

davidscn:
> With extraplation, these configuration files should also trigger some error. If not, then we have a bug in preCICE. Is this the case?

Yes, I just didn't know whether we have any equivalent replacement..

Issue #405, opened 2023-11-19
Trigger generation of systemtests reference _results


MakisH:
We need a workflow that generates the reference results based on a provided `reference_results.yaml`.

It would be nice if this workflow could (optionally) generate reference results only for the cases it does not yet have reference results, instead of rerunning every test.

Issue #401, opened 2023-11-19
Report git hash of each repository


MakisH:
Related to #400.

The actual git commits of each repository are not reported anywhere.

While just `develop` is nice and easy when triggering the tests, it does not let me reproduce the tests. How do I know that my commit is there?

Even something dirty, such as `git show | head -n 5` for each repository would be nice.

valentin-seitz:
What do you mean by repository?

The corresponding solvers+precice versions are "fixed" in the docker-compose.tutorial.yaml. 
The ambiguity of using a branch name can be fixed by providing tags as build_args. 
I can see why this might be a bit unhandy to do (one doesnt want to create a lot of tags)

One thing i could do is to place a file there that resolves the git heads of each of the branches used. (This however would mean i have to checkout the 

But re-triggering this via the github CLI would probably be difficult as its assumed that `git clone -b $REF` is a working command for most of the dockerfile. For the openfoam-adapter this looks like. `git clone --depth 1 -b ${OPENFOAM_ADAPTER_REF} https://github.com/precice/openfoam-adapter.git` 

Ofc, we can introduce the notion of only using commits, this would however, mean a little refactor on the dockerfile side as it will probably not work out of the box. (this would probably also solve some caching issues, e.g hitting a stale cache)

MakisH:
> What do you mean by repository?

Well... Each component.

> The corresponding solvers+precice versions are "fixed" in the docker-compose.tutorial.yaml.

I can pass build arguments to each component, including which version to use. It is important for me to know, when looking at arbitrary results, which state of each component was used to generate those results.

> One thing i could do is to place a file there that resolves the git heads of each of the branches used. (This however would mean i have to checkout the

Why is something like `git show | head -n 5` just after the `git clone` not enough?

Note that this issue is only about reporting, not about passing all kinds of git references to `git clone`.

valentin-seitz:
>I can pass build arguments to each component, including which version to use. It is important for me to know, when looking at arbitrary results, which state of each component was used to generate those results.

I totally agree on this. The problem is: The actual commit is only known by the docker container.  We treat that as a black box right now. We will just stuff some arguments in and in the end we have a container that will happily execute the `./run.sh`.  What we could do is execute a git show | head -n 5 in each of the the build stages and try to make the systemtests print that.


> Note that this issue is only about reporting, not about passing all kinds of git references to git clone.

kind of, but we also run into the issue of stale caches: 
Imagine the following requests to the systemtests: 
[...]

Issue #400, opened 2023-11-19
System tests: Contents of the archive


MakisH:
Example of the contents of an archive:

![Screenshot from 2023-11-19 21-48-16](https://github.com/precice/tutorials/assets/4943683/875abadc-2bd1-4a84-bac1-5fe8a52c5e0f)

I think we still need some changes:

- [ ] `reference_results` and `reference-data` seems to be confusing (which is what?). One of them needs to be renamed, maybe to `results_to_compare`.
- [ ] Do we need the `metadata.yaml`, `docker-compose.tutorial.yaml`, and `docker-compose.field_compare.yaml`? Are they only useful for debugging the system tests themselves? I would suggest to either delete them, or move them to a subfolder (e.g., `system-tests-debug`).
- [ ] What is the `tutorials_ref`? It currently contains `master`, but in that job, I had asked for `develop` (see #399).
- [ ] The actual git commits of each repository are not reported anywhere. While just `develop` is nice and easy when triggering the tests, it does not let me reproduce the tests. How do I know that my commit is there? Even something dirty, such as `git show | head -n 5` for each repository would be nice.

cc @valentin-seitz

valentin-seitz:
>     * [ ]  Do we need the `metadata.yaml`, `docker-compose.tutorial.yaml`, and `docker-compose.field_compare.yaml`? Are they only useful for debugging the system tests themselves? I would suggest to either delete them, or move them to a subfolder (e.g., `system-tests-debug`).

One of the big benefits of using containers is that its very easy to "debug" a problem with a precice component, without having to install all of them manually. So having the docker-compose.tutorial.yaml and the fieldcompare.yaml is a crucial step in reproducing the failed pipeline in your local machine. So i would def. keep them around. A related Issue is: #387 where we need to move some absolute paths into relative ones to allow this "debug the pipeline in your local machine" feature.

MakisH:
Then, let's focus on what is currently not included (or how it is named), and remove stuff later.

PR #410, opened 2023-11-28
Add Workflows for generating reference results 

valentin-seitz:
This PR adds a reusable workflow to automagically add reference results to a certain branch. 
It can be either be triggered manual or another trigger can be used to trigger the reusable workflow (e.g on creation of PR of a certain form)

It also adds more information on the machine (Still open to discussion if we acually need it). But especially the CPU ISA Extensions like AVX-512 etc are of importance. 

The Pipeline now runs and is able to add commits:  https://github.com/precice/tutorials/actions/runs/7181615150


valentin-seitz:
This PR is now ready. 
The CI/CD Pipeline works as expected (see: https://github.com/precice/tutorials/actions/runs/7181615150)
After https://github.com/precice/tutorials/pull/420 is merged, i will rebase and then the tests should go through so we can merge.

valentin-seitz:
I also added small changes to the Dockerfile to allow for commit-ish things to be requested.

tools/tests/systemtests/Systemtest.py
```diff
@@ -354,13 +354,12 @@ def _run_field_compare(self):
...
                 process.kill()
-                # process.send_signal(9)
                 raise KeyboardInterrupt from k
             except Exception as e:
                 logging.critical(
-                    f"Systemtest {self} had serious issues executin the docker compose command about to kill the docker compose command. Please check the logs! {e}")
+                    f"Systemtest {self} had serious issues executing the docker compose command about to kill the docker compose command. Please check the logs! {e}")
                 process.kill()
-                stdout, stderr = process.communicate()
+                process.communicate(timeout=10)
```
MakisH, line 362: What is this magic 10? Should it be:

```suggestion
                process.communicate(timeout=GLOBAL_TIMEOUT)
```
?

Why are the `stdout, stderr` not needed here?

I understand that this is a timeout just for trying to launch Docker. Maybe a comment here would help distinguish from the rest.

tools/tests/systemtests/Systemtest.py
```diff
@@ -397,16 +396,16 @@ def _build_docker(self):
...
-                stdout, stderr = process.communicate()
+                stdout, stderr = process.communicate(timeout=GLOBAL_TIMEOUT)
             except KeyboardInterrupt as k:
                 process.kill()
                 # process.send_signal(9)
                 raise KeyboardInterrupt from k
             except Exception as e:
                 logging.critical(
                     f"systemtest {self} had serious issues building the docker images via the `docker compose build` command. About to kill the docker compose command. Please check the logs! {e}")
+                process.communicate(timeout=10)
```
MakisH, line 407: Since this 10 appears again and again, I am wondering if a common parameter would help.

tools/tests/component-templates/openfoam-adapter.yaml
```diff
@@ -5,10 +5,6 @@ build:
       - {{key}}={{value}}
     {% endfor %}
   target: openfoam_adapter
-  cache_from:
-  - type=gha
-  cache_to:
-  - type=gha,mode=min,scope=openfoam_adapter
```
MakisH, line 11: Why are these not needed anymore? What is the equivalent we currently use when executing inside GitHub Actions?

.github/workflows/generate_reference_results_workflow.yml
```diff
@@ -0,0 +1,69 @@
...
+        echo "Using Ref: ${{ inputs.from_ref }}"
+        echo "Commit message on success:  ${{ inputs.commit_msg }}"
+    - name: Move LFS URL to local LFS server
+      run: |
+        /home/precice/runners_root/scripts/make_lfs_local.sh
+    - name: 'Cleanup the folder'
+      run: |
+        ls -la ./
+        rm -rf ./*
+        rm -rf ./.??*
```
MakisH, line 35: What is this intended for?

.github/workflows/generate_reference_results_workflow.yml
```diff
@@ -0,0 +1,69 @@
...
+        python $GENERATE_REF_RESULTS --log-level=${{inputs.loglevel}}
+        cd ../../
+    - name: Create commit
+      if: success()
+      run: |
+        git checkout ${{ inputs.from_ref }}
+        git add ./*/*/*.tar.gz
+        git add ./*/*.tar.gz
+        git add ./*/*.metadata
+        git commit -m "${{inputs.commit_msg}}"
```
MakisH, line 61: I see that the commit author is `preCICE Tests VM <tests@precice.org>`. Is that configured in the system? It should probably be documented in the infrastructure.

MakisH (approved):
The implementation looks very nice!
Some comments to consider, feel free to merge afterwards.

MakisH:
Closes #405.

This is related to #401 (git hash reporting) by requiring (only via documentation) specific Git commit or tag (not branches, which can change over time) when generating reference results. I would still feel safer if I had a dump of all hashes in one log file, but I remember from our last meeting that this is complicated due to the current architecture. I assume it still creates the same report for the current/tested results. So, yes, it closes #401 as well.

Regarding #400 (contents of the archive):
- `reference_data` is renamed to `reference-results`
- `TUTORIALS_REF` is included in the report
- I have not checked what the actual contents are, because I don't have an archive readily available, but I still believe it closes #400.

tools/tests/systemtests/Systemtest.py
valentin-seitz: You have a good eye ;) 
It was a hacky solution i threw in to check if it works.

tools/tests/component-templates/openfoam-adapter.yaml
valentin-seitz: The default caching behavior of the local builder is suffiencient and does the right thing. 
This is only needed to enable GHA caching (which is still broken)

.github/workflows/generate_reference_results_workflow.yml
valentin-seitz: As we reuse our VM for all systemtests we cannot safely assume that the directory is emtpy (in fact the last systemtests execution will be still present), so we need to clean up a bit. 
The `./.??*` is there to make sure no hidden files like `./.git` survive.

flow-over-heated-plate/reference_results.metadata
```diff
@@ -27,3 +28,36 @@ This file contains an overview of the results over the reference results as well
...
 | CALULIX_VERSION | 2.20 |
 | CALULIX_ADAPTER_REF | v2.20.0 |
+## Information about the machine
+
+### uname -a
+
+Linux precice-tests 5.15.0-89-generic #99-Ubuntu SMP Mon Oct 30 20:42:41 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
+
+
+### lscpu
```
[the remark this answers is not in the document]
valentin-seitz: My idea is that the set of arguments used to generate the test results should be everything you need to reproduce the results (and the machine you executed on, especially the CPU ISA extension). 
So a dump of the hashes can be added, but it should not add any additional information :)

tools/tests/README.md
```diff
@@ -38,17 +38,17 @@ gh workflow run run_testsuite_manual.yml -f suites=fenics_test --ref=develop
 Another example, to use the latest develop branches and enable debug information of the tests:
 
 ```shell
-gh workflow run run_testsuite_manual.yml -f suites=fenics_test -f build_args="PRECICE_REF:develop,OPENFOAM_ADAPTER_REF:develop,PYTHON_BINDINGS_REF:develop,FENICS_ADAPTER_REF:develop" -f loglevel=DEBUG --ref=develop
+gh workflow run run_testsuite_manual.yml -f suites=fenics_test -f build_args="PRECICE_REF:v2.5.0,OPENFOAM_ADAPTER_REF:v1.2.3,PYTHON_BINDINGS_REF:v2.5.0.4,FENICS_ADAPTER_REF:v1.4.0" -f loglevel=DEBUG --ref=develop
 ```
 
-where the `*_REF` can also be specific Git commits.
+where the `*_REF` should be a specific [commit-ish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish).
```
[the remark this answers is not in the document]
valentin-seitz: I mean there are easy ways to do this: 
- Manipulate the Build_argument and append the current day and be okay if develop might be a day old? 
- Have a cronjob invalidate the build cache in a certain interval.
etc. 

I think using commit-ish objects is the way to go here in the long run. 
Only if we see, that we rebuild a lof of the layers often or spend more time building the container than running the simuations, we might need to think about it. We then can think of trading off some consistency for performance.

flow-over-heated-plate/reference_results.metadata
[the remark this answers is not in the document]
MakisH: I consider this just a psychological issue. Let's just live with it.

tools/tests/README.md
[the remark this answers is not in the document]
MakisH: Continuing in #423

tools/tests/component-templates/openfoam-adapter.yaml
MakisH: Since we don't use the GitHub-hosted runners anymore, we don't need the GitHub Actions caching mechanism, which was anyway broken.

tools/tests/systemtests/Systemtest.py
valentin-seitz: 561533e2c6960c44ac29890c76433448bc4d97af should solve it

.github/workflows/generate_reference_results_workflow.yml
valentin-seitz: 248fff14ebbfe3078eab2669d554d02a39ff371f solves this

PR #411, opened 2023-11-28
Clean precice-profiling/

MakisH:
Removes the `precice-profiling/` directory from every case directory.

@davidscn any other new files we should clean up?

### DIFF
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -32,7 +32,7 @@ clean_precice_logs() {
             ./precice-*-watchintegral-*.log \
             ./events.json \
             ./core
-        rm -rfv ./precice-events/
+        rm -rfv ./precice-events/ ./precice-profiling/
     )
 }
 


tools/cleaning-tools.sh
```diff
@@ -32,7 +32,7 @@ clean_precice_logs() {
             ./precice-*-watchintegral-*.log \
             ./events.json \
             ./core
-        rm -rfv ./precice-events/
+        rm -rfv ./precice-events/ ./precice-profiling/
```
davidscn, line 35: Our tutorials are related to a specific preCICE version and the new state of the tutorials won't work with any older preCICE version, so I would consider removing (or rather replacing) `precice-events` here.

`precice-profiling` is now what preCICE auto-generates. If people really work with the events, one would typically (or by default) generate  `profiling.json` (previously it was `events.json`). It's somewhat of a political question whether we want to delete it in the cleaning script or not. Generating a trace file (more rare) would generate a `trace.json`.

uekerman: In the tutorials, I would expect a full clean, but no strong opinion.

tools/cleaning-tools.sh
```diff
@@ -30,9 +30,9 @@ clean_precice_logs() {
             ./precice-*-convergence.log \
             ./precice-*-watchpoint-*.log \
             ./precice-*-watchintegral-*.log \
-            ./events.json \
             ./core
```
MakisH, line 33: @davidscn the `./core` was added by @uekerman in https://github.com/precice/tutorials/commit/b797642ce90c66fe28d8edcbaca5e0ebe805564e, as part of https://github.com/precice/tutorials/pull/160.

Why would you remove it if we don't know where it comes from?

> I would also remove the `core`. I will leave it up to you, so ready to merge

Btw, it is also still in the `.gitignore` of preCICE:

https://github.com/precice/precice/blob/39accb6fd3fd9ac35ae06a13270e3fca7127db45/.gitignore#L14

tools/cleaning-tools.sh
```diff
@@ -30,9 +30,9 @@ clean_precice_logs() {
...
-        rm -rfv ./precice-events/
+        rm -rfv ./precice-profiling/ profiling.json trace.json
+        rm -rfv ./preCICE-output/
```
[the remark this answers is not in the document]
MakisH: I am merely refactoring the current script. Let's not resolve the name inconsistency here, I am afraid it may have side-effects. Opened a separate issue: https://github.com/precice/tutorials/issues/417

tools/cleaning-tools.sh
uekerman: IMO cleaning the `core` file after crashes can be helpful. But it's not sth that preCICE creates, of course.

PR #413, opened 2023-11-29
Remove unnecessary `clean_aste` call from base cleaning

IshaanDesai:
As mentioned in the title. I noticed this in the cleaning logs while creating a new tutorial.

### DIFF
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -121,7 +121,7 @@ clean_aste() {
     (
         set -e -u
         cd "$1"
-        echo "--- Cleaning up ASTE results in $(pwd)"
+        echo "--- Looking for and cleaning up any potential ASTE results in $(pwd)"
         rm -fv result.vtk result.stats.json
         rm -fvr fine_mesh coarse_mesh mapped
     )
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -120,7 +120,8 @@ clean_su2() {
 clean_aste() {
     (
         set -e -u
-        echo "--- Cleaning up ASTE results"
+        cd "$1"
+        echo "--- Cleaning up ASTE results in $(pwd)"
         rm -fv result.vtk result.stats.json
         rm -fvr fine_mesh coarse_mesh mapped
     )
--- a/tools/cleaning-tools.sh
+++ b/tools/cleaning-tools.sh
@@ -121,7 +121,7 @@ clean_aste() {
     (
         set -e -u
         cd "$1"
-        echo "--- Looking for and cleaning up any potential ASTE results in $(pwd)"
+        echo "--- Cleaning up ASTE results in $(pwd)"
         rm -fv result.vtk result.stats.json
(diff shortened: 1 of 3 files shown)


MakisH:
Why is this not necessary? I don't remember the motivation for putting it there, but I assume it was because we have at least one tutorial with ASTE.

Does it create any issues?

IshaanDesai:
We call `tools/clean-tutorial-base.sh` to clean any tutorial, and in doing this, one sees

```bash
--- Cleaning up ASTE results
```

even if no ASTE results have been cleaned. It is misleading and confusing. For example, this occurs if we run the `clean-tutorial.sh` script in the Quickstart.

MakisH:
Every tutorial contains a `clean-tutorial.sh`, which is a symlink to `../tools/clean-tutorial-base.sh`. This allows us to reduce duplication.

**Note:** The only tutorial that currently does not follow this approach is the `two-scale-heat-conduction`. Please adapt that to follow the same convention.

Now, in case we really want to not show this message, we have two possible approaches:

1. Modify each cleaning function to check if results exist before displaying any message (a bit tricky).
2. Specifically for the `aste-turbine` tutorial, replace the symlink to a script that duplicates the `clean-tutorial-base.sh`. But this would be a special case. And duplicate. And I am not a big fan of special cases where we have already set up some standards. :sweat_smile: 

This seems to be a high effort / tiny benefit to me. As a workaround, I would suggest rephrasing the text of each function to include the word "potential".

MakisH:
> Adding a `clean.sh` script in the ASTE tutorial which has the `clean_aste()` call would be a better solution, and also what would stick to our standards.

I see now. Yes, this should work, please do it!

And if you can test with ASTE, that would be great (I currently have some issues with dependencies that I don't want to look into).

IshaanDesai:
I did not run the ASTE tutorial, but just running the script `aste-turbine/clean-tutorial.sh` shows the correct print statements on the terminal, which I think is a good enough check.

tools/cleaning-tools.sh
```diff
@@ -12,6 +12,10 @@ clean_tutorial() {
         echo "-- Cleaning up all cases in $(pwd)..."
         rm -rfv ./precice-run/
 
+        if test "clean.sh"; then
```
MakisH, line 15: Lines 15-17 and 19-24 are aiming to do the same thing, no?

IshaanDesai: Initially I thought so too, but lines 19-24 iterate through folders in a tutorial, and for the aste-turbine tutorial, we do not have a participant folder, but just the base folder. Hence I added lined 15-17 which check if there is a `clean.sh` in the tutorial folder itself, and if yes, runs it.

MakisH: Then at least a comment here on the intention (where the script is called from) would be helpful for the future reader :sweat_smile:

IshaanDesai: Done :)

davidscn (approved):

PR #414, opened 2023-11-29
Port quickstart to precice v3

davidscn:


quickstart/solid-cpp/rigid_body_solver.cpp
```diff
@@ -202,21 +201,18 @@ int main()
...
-      precice.readBlockVectorData(read_id,
-                                  n_nodes,
-                                  vertex_ids.data(),
-                                  forces.data());
+    // We use an explicit time integration scheme, thus we always read data
+    // at the beginning of a time window (0)
+    precice.readData(mesh_name,
+                     data_read_name,
+                     vertex_ids,
+                     0,
```
davidscn, line 209: @BenjaminRodenberg what am I supposed to put here? OpenFOAM uses an implicit time integrator, the solid an explicit one. In principle, I would expect `0` to be the correct value, and it also works, but I get the warning 
```
The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
```
in the last time window after converging. With an argument of `dt` I don't get this warning.

BenjaminRodenberg: Since `readData` basically provides your right-hand side, I also think the combination of window end for the implicit integrator and window beginning for the explicit one sounds reasonable. Do you have some kind of reference solution to compare against? If you evaluate the right-hand side at the wrong point in time I would expect an error of $\mathcal{O}(\Delta t)$. This is maybe still good enough to work, but you might get a difference in the quality of the result.

Maybe this is also a hint that you do not really need an implicit coupling scheme here? To some degree it also makes sense that the coupling residual becomes zero, because the explicit integrator just solves the same problem with the same input over and over again, if we use an implicit coupling scheme. Why not use an explicit coupling scheme, where the explicit integrator goes first, then the implicit integrator second.

davidscn: >  because the explicit integrator just solves the same problem with the same input over and over again, if we use an implicit coupling scheme.

The readTime at t = 0 returns the coupling data from the previous time window, right? As this is our very basic tutorial, I would say having an implicit scheme makes sense to highlight the capabilities of preCICE (but no strong opinion). Setting the Solid as first participant and the Fluid as second participant in an implicit coupling scheme, i.e., applying acceleration and convergence measures to `Force` gets rid of the warning and runs smoothly as well (and it also makes more sense as you already mentioned, but in preCICE v2 it was not possible to distinguish here the read time)

I think the quickstart should state an example in terms of the configuration. Using an explicit coupling scheme works here as well (with Fluid being the second participant). Opinions? Maybe @MakisH ?

BenjaminRodenberg: > The readTime at t = 0 returns the coupling data from the previous time window, right?

No so happy with the wording :smirk: $t = 0$ points to the beginning of the current time step. If you do not use subcycling this is the same time as the end of the previous window.

Alternative suggestion: How about using a different time stepping scheme? With generalized $\alpha$ or Newmark $\beta$ you, for example, rely on a right-hand side evaluation that is not at the beginning of the window. So here iterating (and even interpolating) makes much more sense. If you do not want to worry about interpolation (it should be a simple quickstart) Newmark $\beta$ should be fine, because it requires a right-hand side evaluation at the end of the time step (also consistent with what we always did in preCICE v2)

MakisH: > what am I supposed to put here?

From the user perspective, I understand that one should always put the `dt` there. If there are reasons to distinguish cases, then this needs to be clearly documented and communicated at different levels. My gut feeling tells me that this is wrong, or that our new API is not high-level enough.

> but I get the warning

Sounds related to https://github.com/precice/precice/issues/1558, where this warning only appears for a serial-implicit, but not for a parallel-implicit.

> I think the quickstart should state an example in terms of the configuration. Using an explicit coupling scheme works here as well (with Fluid being the second participant). Opinions? Maybe @MakisH ?

Well, all tutorials should be examples of good configuration practices. What is special about the Quickstart is that:

1. it should just work
2. the user can see and understand the (short) code
3. is should be very simple, but still resemble some realistic geometrical and configuration setup.

Of course, if we get the aforementioned warning, we break the "it just works". Users will be asking "why does this highlighted tutorial show warnings?".

Also, if we have to hard-code a value, this would generate questions, or lead people to hard-code the same value by default.

I would aim to maintain the serial-implicit scheme, or at least a parallel-implicit, since this is the configuration that one would use in most applications. Showing an explicit scheme in our highlighted tutorial may give wrong impressions.

BenjaminRodenberg: > > what am I supposed to put here?
> 
> From the user perspective, I understand that one should always put the `dt` there. If there are reasons to distinguish cases, then this needs to be clearly documented and communicated at different levels. My gut feeling tells me that this is wrong, or that our new API is not high-level enough.

I think the kind of discussion we are having here is exactly what we want to trigger with our users in the future and that's also the reason why we are having the `dt` as a mandatory parameter now. The simple high-level approach is to always use `dt = getMaxTimeStepSize()`. The the user does not really have to worry about anything, but this, of course comes at a cost and might reduce the order of your time stepping scheme or the performance.

In this example: If you use `dt = getMaxTimeStepSize()` you should get the same result like in v2 and probably also not the `The coupling residual ...` error. But you pay with an implicit coupling scheme that you do not really need. Like already mentioned above: Explicit coupling with solid first using `dt=0` should give you the same or even better quality of results and better performance.

If you do not want to open this box in this simple tutorial and still want to do things right w.r.t time stepping I would suggest to use a different time stepping scheme where a right-hand side evaluation at `dt=getMaxTimeStepSize()` is correct. (Remember: Our current strategy in the documentation is "just write `dt=getMaxTimeStepSize()` and don't think about it")

quickstart/solid-cpp/rigid_body_solver.cpp
```diff
@@ -177,13 +176,13 @@ int main()
...
-                          vertex_ids.data());
+  precice.setMeshVertices(mesh_name,
+                          vertices,
+                          vertex_ids);
 
-  // initialize the Solverinterface
-  double dt = precice.initialize();
+  // initialize precice
+  precice.initialize();
+  double dt = precice.getMaxTimeStepSize();
```
BenjaminRodenberg, line 185: Is there any possibility that you might want to use subcycling here? Generally we suggest to do the following in our documentation (see https://precice.org/couple-your-code-steering-methods.html):

```
double solverDt = ???;  // given by the user. precice.getMaxTimeStepSize() would be the easiest option.
double preciceDt = precice.getMaxTimeStepSize();
double dt = min(preciceDt, solverDt);
```

Please also shift this part inside the `while(precice.isCouplingOngoing()` loop before the `readData`.

quickstart/solid-cpp/rigid_body_solver.cpp
```diff
@@ -202,21 +201,18 @@ int main()
...
+                     data_read_name,
+                     vertex_ids,
+                     0,
+                     forces);
 
     // Store time dependent values
-    if (precice.isActionRequired(
-            precice::constants::actionWriteIterationCheckpoint())) {
+    if (precice.requiresWritingCheckpoint())
       data_container.save_old_state(vertices, theta, theta_dot);
```
BenjaminRodenberg, line 214: Please write the checkpoint at the very beginning of the simulation loop, not somewhere in the middle.

quickstart/solid-cpp/rigid_body_solver.cpp
```diff
@@ -227,24 +223,19 @@ int main()
...
-                                   displacement.data());
+    precice.writeData(mesh_name,
+                      data_write_name,
+                      vertex_ids,
+                      displacement);
 
     std::cout << "Rigid body: advancing in time\n";
-    dt = precice.advance(dt);
+    precice.advance(dt);
+    dt = precice.getMaxTimeStepSize();
```
BenjaminRodenberg, line 233: This is the wrong place for this API call. You are getting the `dt` from the next window already here, but you are increasing the `time += dt` later with that `dt`.

I would increase the `time` as soon as the time step is completed. This is even before advance. Then reset the time to the checkpointed value, if necessary. Otherwise you might also get problems with subcycling, if you only increase the time when `precice.isTimeWindowComplete()`.

BenjaminRodenberg (changes requested):
Generally, I think we should follow the order of API calls given in the [step-by-step guide](https://precice.org/couple-your-code-steering-methods.html) inside of the simulation loop. Not `read->solve->write->advance->determine dt` but `determine dt->read->solve->write->advance`.

Especially with subcycling and first participant time stepping it is easy to do things wrong here, if you update the time step size at the wrong point in time.

davidscn:
Purpose of the changes here are updating the API calls. If you think we should rewrite the numerics, please open an issue. I made now the solid participant `first` and the fluid participant `second` and the tutorial runs smooth.

davidscn:
Despite some discussion, I moved now to an explicit coupling scheme, as it is numerically much more reasonable and the results are the same. Nice side effect: the tutorial runs twice as fast. For implicit coupling and QN settings, the perpendicular flap example should still be basic enough.

PR #415, opened 2023-11-29
Higher-order implicit Runge-Kutta methods for the partitioned heat equation

NiklasVin:
With this implementation, you could compute the partitioned heat equation with any implicit RK method, which is provided
by ButcherTableaux.py file.
You just need to change the time stepping scheme `tsm` to the IRK you want to use. 
The weak form is obtained algorithmically and does not need to be changed by the user.


partitioned-heat-conduction/fenics/heat.py
```diff
@@ -216,16 +257,50 @@ def determine_gradient(V_g, u, flux):
...
     # Dirichlet BC and RHS need to point to end of current timestep
     u_D.t = t + float(dt)
-    f.t = t + float(dt)
-
-    # Coupling BC needs to point to end of current timestep
-    read_data = precice.read_data(dt)
+    # update boundary conditions
+    for i in range(tsm.num_stages):
+        f[i].t = t + tsm.c[i] * float(dt)
+        du_dt[i].t = t + tsm.c[i] * float(dt)
```
BenjaminRodenberg, line 263: Instead of `tsm.c[i] * float(dt)` you could also introduce a variable `stage_time`. This would be relative to `t`, so you need `t + stage_time` for `f` and `du_dt` and only `stage_time` for `read_data` further below.

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -216,16 +257,50 @@ def determine_gradient(V_g, u, flux):
...
+
+        # get first derivative
+        bsplns_der = {}
+        for ki in bsplns.keys():
+            bsplns_der[ki] = bsplns[ki].derivative(1)
+        for i in range(tsm.num_stages):
+            # values of derivative at current time
+            val = {}
+            for ki in bsplns_der.keys():
+                val[ki] = bsplns_der[ki](tsm.c[i]*float(dt))
```
BenjaminRodenberg, line 278:
```suggestion
                val[ki] = bsplns_der[ki](stage_time)
```
See above

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -216,16 +257,50 @@ def determine_gradient(V_g, u, flux):
 
     # Dirichlet BC and RHS need to point to end of current timestep
     u_D.t = t + float(dt)
-    f.t = t + float(dt)
-
-    # Coupling BC needs to point to end of current timestep
-    read_data = precice.read_data(dt)
+    # update boundary conditions
+    for i in range(tsm.num_stages):
```
BenjaminRodenberg, line 261: I'm not so happy with the order of the loops, because the place where you are setting `f[i]` and `du_dt[i]` are quite far apart. Why not the following?

```
for i in range(tsm.num_stages):
    f[i].t = ...
    du_dt[i].t = ...
    
        # DirichletBC for stage i
    else:
        # NeumannBC for stage i 
```
I think this would save the second `for i in range(tsm.num_stages)`-loop inside the individual Dirichlet/Neumann branches.

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -271,9 +346,6 @@ def determine_gradient(V_g, u, flux):
         error, error_pointwise = compute_errors(u_n, u_ref, V, total_error_tol=error_tol)
         print("n = %d, t = %.2f: L2 error on domain = %.3g" % (n, t, error))
 
-    # Update Dirichlet BC
-    u_D.t = t + float(dt)
-    f.t = t + float(dt)
```
BenjaminRodenberg, line 276: If I remember correctly, this was needed to make sure that `u_D` and `f` are also correct at the very end of the simulation loop. Is this still the case? But generally it makes sense to perform the update of the boundary conditions etc. where you are doing it now. Maybe you just need to duplicate this part after leaving the `while precice.is_coupling_ongoing()` to make sure the output at `t_end` is correct. See below at
```
...
for sample in ref_write:
    ref_out << sample
...
```

BenjaminRodenberg: After looking at this closer: I think the output in `ref_out` is probably wrong, because the update `u_ref = interpolate(u_D, V)` happens *after* `uu_ref = u_ref.copy()`. This is something we should directly fix on `develop`.

partitioned-heat-conduction/fenics/utils/ButcherTableaux.py
BenjaminRodenberg, line 1: Is there a need to review this file? I think we should keep it consistent with https://github.com/firedrakeproject/Irksome/blob/master/irksome/ButcherTableaux.py, if no changes are necessary.

partitioned-heat-conduction/fenics/utils/utils.py
```diff
@@ -0,0 +1,102 @@
...
+def getVariationalProblem(v, k, tsm, f, dt, initialCondition):
+    """
+    :param v: test function
+    :param k: trial function
+    :param tsm: time stepping method which should be used
+    :param f: f of the problem
+    :param dt: time step (must not be a float but a fenics expression! e.g. Constant(0))
+    :param initialCondition: function which defined the inital value/initial condition
+    :return: returns variational problem (F=0) of the given params from the constructor
+    """
```
BenjaminRodenberg, line 66: Can you add some more documentation here? This is basically the "heart" of the RK version of `heat.py`, isn't it?

partitioned-heat-conduction/fenics/utils/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    Defines the weak left hand side of a given problem
+    :param u: trial function
+    :param v: test function
+    :param k: function of function space for stages of RK methods
+    """
+    return inner(k, v) * dx + inner(grad(u), grad(v)) * dx
+
+
+def getVariationalProblem(v, k, tsm, f, dt, initialCondition):
```
BenjaminRodenberg, line 57:
```suggestion
def getVariationalProblem(v, k, tsm, f, dt, initial_condition):
```

partitioned-heat-conduction/fenics/utils/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    :param dt: length of time step
+    :return: expression with the time derivative according to tsm
+    """
+    # get the time derivative of expr
+    du_dt_expr = expr.diff(t)
+    du_dt = tsm.num_stages * [None]
+    for i in range(tsm.num_stages):
+        du_dt[i] = Expression(ccode(du_dt_expr), degree=2, t=0)
+        du_dt[i].t = du_dt[i].t + tsm.c[i]*dt(0)
+    return du_dt
```
BenjaminRodenberg, line 102:
```suggestion
    du_dt = tsm.num_stages * [None]
    for i in range(tsm.num_stages):
        du_dt[i] = Expression(ccode(expr.diff(t)), degree=2, t=0)
        du_dt[i].t = du_dt[i].t + tsm.c[i]*float(dt)
    return du_dt
```

BenjaminRodenberg: Two more comments:
* Why `du_dt[i].t = du_dt[i].t + tsm.c[i]*float(dt)`? Is `du_dt[i].t` not `0`?
* Maybe the following even works: `du_dt[i] = Expression(ccode(expr.diff(t)), degree=2, t=tsm.c[i]*float(dt))`

Then you could even do everything in a list comprehension (maybe you do not even need this function):
`du_dt = [Expression(ccode(expr.diff(t)), degree=2, t=tsm.c[i]*float(dt)) for i in range(tsm.num_stages)]`

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -116,28 +132,58 @@ def determine_gradient(V_g, u, flux):
...
+F = utl.getVariationalProblem(v=v, initialCondition=u_n, dt=dt, f=f, tsm=tsm, k=u)
+
+# boundary conditions
+
+# we define variational form for each RK stage. According to
+# https://doi.org/10.48550/arXiv.2006.16282, All of those stages
+# represent time derivatives and not the actual solution. Thus, we need time derivatives as boundary conditions
+
+# get time derivative of u
+du_dt = utl.time_derivative(u_D_sp, tsm, dt, t_sp)
```
BenjaminRodenberg, line 156: We could remove this function. See below.

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -116,28 +132,58 @@ def determine_gradient(V_g, u, flux):
...
+f = tsm.num_stages * [None]
 # du_dt-Laplace(u) = f
 f_sp = u_D_sp.diff(t_sp) - u_D_sp.diff(x_sp).diff(x_sp) - u_D_sp.diff(y_sp).diff(y_sp)
-f = Expression(sp.ccode(f_sp), degree=2, alpha=alpha, beta=beta, t=0)
-F = u * v / dt * dx + dot(grad(u), grad(v)) * dx - (u_n / dt + f) * v * dx
-
-bcs = [DirichletBC(V, u_D, remaining_boundary)]
+for i in range(tsm.num_stages):
+    f[i] = Expression(sp.ccode(f_sp), degree=2, t=0)
+    f[i].t = tsm.c[i] * float(dt)  # initial time assumed to be 0
```
BenjaminRodenberg, line 145: Again, try to use a list comprehension here like suggested below in `time_derivative` (if this works)

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -116,28 +132,58 @@ def determine_gradient(V_g, u, flux):
...
 
 # Define variational problem
-u = TrialFunction(V)
-v = TestFunction(V)
+# trial and test functions
+u = TrialFunction(Vbig)
+v = TestFunction(Vbig)
+# if dim(Vbig)>1, f needs to be stored in an array with different times,
+# because in each stage, of an RK method it is evaluated at a different time
+f = tsm.num_stages * [None]
```
BenjaminRodenberg, line 140:
```suggestion
```
Not needed, if we use list comprehension here.

BenjaminRodenberg (changes requested):
Couldn't resist taking a look at the PR :smirk: Generally looks good. I did not check the actual results, but this is something you are doing in much more detail than I can do it here. Therefore, I was mainly focusing on code style.

We should also try to directly move some minor changes to `develop`, but this is something I can do.

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -22,20 +22,25 @@
...
+    File, solve, lhs, rhs, grad, inner, dx, interpolate, VectorFunctionSpace, MeshFunction, MPI, MixedElement, split
 from fenicsprecice import Adapter
 from errorcomputation import compute_errors
 from my_enums import ProblemType, DomainPart
 import argparse
 import numpy as np
 from problem_setup import get_geometry
 import dolfin
-from dolfin import FacetNormal, dot
+from dolfin import FacetNormal, dot, project
```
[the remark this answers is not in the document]
NiklasVin: `project` is provided by `dolfin`. And as the imports were already there, I used them.
Actually, also regarding the current implementation of heat.py, all imports from `dolfin` can be deleted, because `dot` is already imported above from `fencis`, and, as you said, `FacetNormal` is not used.

partitioned-heat-conduction/fenics/utils/ButcherTableaux.py
NiklasVin: No, I don't think so. I only added the reference at the top of the file.

partitioned-heat-conduction/fenics/heat.py
NiklasVin: Yes, that would work

partitioned-heat-conduction/fenics/heat.py
NiklasVin: I understand what you mean but I think this is not really practical as I compute the B-Splines/derivatives for all stages.
If I implemented it as you suggested I would either need to compute the derivatives in each iteration or 
before the for loop. The second option would also lead to a similar situation as we have it with the current implementation🤔

partitioned-heat-conduction/fenics/heat.py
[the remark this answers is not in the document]
BenjaminRodenberg: But you can also use `from fenics import project`. I would suggest: Let's remove all `dolfin` imports here. We can then cherry-pick this change to `develop`, because you are of course right that it is not 100% the point of this PR to clean up the imports and therefore we should also not mix this up.

partitioned-heat-conduction/fenics/heat.py
BenjaminRodenberg: If this is technically not possible, I was probably missing something yesterday when suggesting the rearrangement. Then let's keep things as they are.

NiklasVin: Your suggestion is possible to implement, but I was thinking whether your suggestion or the current state makes the code cleaner. As the B-Splines need to be computed once and are used by all stages, you would need to move the `b_splines` call out of the for loop. With that you would avoid one for loop, but then you'd need an additional if statement that the derivatives are only computed for the Dirichlet side. Because the number of stages are not very high, the runtime is also not much affected by the two for loops. Of course I am an advocate for the current implementation, but I still think, the code now makes the separation a bit more clear when the derivatives are required

partitioned-heat-conduction/fenics/utils/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    for i in range(num_stages):
+        rh = rh + f[i] * vs[i] * dx
+
+    # Assemble weak form from lhs and rhs
+    F = 0
+    for i in range(num_stages):
+        # cf. https://doi.org/10.1145/3466168 p.5 equation 14
+        F = F + weak_lhs(v=vs[i], u=u[i], k=ks[i])
+    F = F - rh
+    return F
```
BenjaminRodenberg, line 94:
```suggestion
    # Assemble weak form
    F = 0
    for i in range(num_stages):
        # cf. https://doi.org/10.1145/3466168 p.5 equation 14
        F += inner(ks[i], vs[i]) * dx + inner(grad(u[i]), grad(vs[i])) * dx - f[i] * vs[i] * dx
    return F
```
I don't really see a good reason for introduction `weak_lhs` here. Currently, the whole script does not generalize that well, if you want to solve a different equation anyway, because the `f` comes from "the outside". But this is also ok here.

If you want to get rid of the index, you can also try:
```
for u, v, k, f in zip(us, vs, ks, fs):
    F += inner(k, v) * dx + inner(grad(u), grad(v)) * dx - f * v * dx
```
btw: Renaming the `u` from above to `us` and the `f` to `fs` I think would not harm. Currently, it's a bit odd that you say `vs`, but not `us`.

BenjaminRodenberg (commented):
Minor readability comments. If this means that you have to start with all your testing from the beginning: You can also ignore these suggestions and we just apply this refactoring after this PR is merged.

BenjaminRodenberg:
I think we should definitely try to get this merged into the tutorials, because it is an important test for waveforms in preCICE. But I'm still a bit unsure how we should continue here: For the sake of the review it was very useful to modify `heat.py`, but to me it is not clear whether we should still keep the simple version of the tutorial (without stages, just implicit Euler directly solving the system for u).

So there are the following three options:

* Replace `heat.py` with the advanced high-order version. Drop the old and simple implicit Euler version.
* Provide high-order as `heat.py`, but also keep a `heatImplicitEuler.py` in this folder.
* Leave `heat.py` as it is, but provide `heatHigherOrder.py` in this folder.

Then, there is, of course, also the option to move this case into an entirely new folder `tutorials/partitioned-heat-conduction-high-order`, but I think this would not give this contribution the right level of attention.

Opinions @uekerman and @MakisH?

partitioned-heat-conduction/fenics/heat.py
```diff
@@ -22,18 +22,23 @@
   u = u_0             at t = 0
   u = 1 + x^2 + alpha*y^2 + \beta*t
   f = beta - 2 - 2*alpha
+
+The implementation of the higher-order implicit Runge-Kutta methods was mostly taken from:
+    https://github.com/NikoWul/FenicsIrksome
```
MakisH, line 27: FYI, the original repository has no license. I understand that @BenjaminRodenberg was related to (supervised?) that project, but it would be best to make this clear.

MakisH:
> I think we should definitely try to get this merged into the tutorials, because it is an important test for waveforms in preCICE. 

Definitely! No good work should see slow decay in unmerged PRs.

> But I'm still a bit unsure how we should continue here: For the sake of the review it was very useful to modify `heat.py`, but to me it is not clear whether we should still keep the simple version of the tutorial (without stages, just implicit Euler directly solving the system for u).

My main argument here would be consistency: We have invested a lot of effort in having all solver options of each tutorial behave in similar ways and give similar results. By merging this as-is, it means that FEniCS will be doing something different than the rest of the solver options.

> So there are the following three options:
> 
>     * Replace `heat.py` with the advanced high-order version. Drop the old and simple implicit Euler version.

That would be the breaking case, so for me, this is a clear "no".

>     * Provide high-order as `heat.py`, but also keep a `heatImplicitEuler.py` in this folder.

This changes the default behavior, and hence the consistency with the rest, so also "no" for me.

>     * Leave `heat.py` as it is, but provide `heatHigherOrder.py` in this folder.

This sounds like the easiest, least intrusive option. How easy would it be to merge the two and make the implicit solver a special case, configurable at runtime? That could prevent some duplication.

> Then, there is, of course, also the option to move this case into an entirely new folder `tutorials/partitioned-heat-conduction-high-order`, but I think this would not give this contribution the right level of attention.

We typically do this when we need a different `precice-config.xml`, which is not the case here. It would mainly lead to duplication.

**My recommendation:** Leave `heat.py` as-is, but provide the higher order solver as a special case (either in the same code or, if that is complicated, in a separate file).

partitioned-heat-conduction/fenics/heat.py
BenjaminRodenberg, line 27: Good point. I think we should also refer to the corresponding thesis of Nikola and Niklas (in preparation). This clarifies my role as an advisor as well.

```suggestion
For information on the partitioned heat conduction problem using higher-order implicit Runge-Kutta methods see
* "Wullenweber, Nikola. High-order time stepping schemes for solving partial differential equations with FEniCS. Bachelor's thesis at Technical University of Munich, 2021. URL: https://mediatum.ub.tum.de/1621360"
* "Vinnitchenko, Niklas. Evaluation of higher-order coupling schemes with FEniCS-preCICE. Bachelor's thesis at Technical University of Munich, 2023. (in preparation)"

The implementation of the higher-order implicit Runge-Kutta methods is based on: https://github.com/NikoWul/FenicsIrksome
```

BenjaminRodenberg (approved):
Looks good to me :+1: Will merge this now and apply some final changes directly on develop.

PR #429, opened 2024-01-04
nutils additions for the perpendicular flap example

### LINKED ISSUE #217: Nutils code in prependicular-flap/fluid-nutils does not work for Nutils v7.0
- Tutorials state (last commit / release): https://github.com/precice/tutorials/tree/d2c70c09e6ae50c2a586edfe9008730902ebb980
- Versions of solvers and adapters used: Nutils `v7.0a0` 
- preCICE version: v2.2.1

The Nutils code in `perpendicular-flap/fluid-nutils` does not work and produces the following error:
```
  File "fluid.py", line 13, in <module>
    @function.replace
AttributeError: module 'nutils.function' has no attribute 'replace'
```
Nutils was installed by cloning the repository from GitHub and running `python3 -m pip install --user .`

gertjanvanzwieten:
This PR reworks the Nutils fluid participant to make the code cleaner and compatible with Nutils 8, while remaining compatible with Nutils 7. It also adds a solid participant compatible only with Nutils 8. Only the final commit relates to the solid participant so it can trivially be split off into a separate PR if that is preferred.

The fluid modifications are organized in many small commits with their distinct purpose hopefully clear from the commit message. Up to d89f08d the formulation itself is not modified, though results change slightly because of minor bug fixes (e.g. 4b7b925, 6e05ed0). Beyond that the formulation is progressively simplified to the point of requiring only the previous solution, rather than the past four. The change of d89f08d could optionally be modified to involve the previous `F` for higher accuracy, though at the expense of oscillations, which the current implementation also suffers from -- the proposed formulation is oscillation free. Commit 35b8b34 does away with the hacky `qw` definition by switching to a uniform integration scheme, which anyway seems like reasonable enough given the non-polynomial nature of the integrand. All in all, the modified script is of a more standard nature and should be easier to follow for most Nutils users.

The failing Lint docs test seems unrelated to the changes in this PR.

closes #217

### DIFF
--- a/perpendicular-flap/fluid-nutils/fluid.py
+++ b/perpendicular-flap/fluid-nutils/fluid.py
@@ -15,11 +15,11 @@ def subs0(f):
     if isinstance(f, function.Argument) and f._name == 'lhs':
         return function.Argument(name='lhs0', shape=f.shape, nderiv=f._nderiv)
     if isinstance(f, function.Argument) and f._name == 'meshdofs':
-        return function.Argument(name='oldmeshdofs', shape=f.shape, nderiv=f._nderiv)
-    if isinstance(f, function.Argument) and f._name == 'oldmeshdofs':
-        return function.Argument(name='oldoldmeshdofs', shape=f.shape, nderiv=f._nderiv)
-    if isinstance(f, function.Argument) and f._name == 'oldoldmeshdofs':
-        return function.Argument(name='oldoldoldmeshdofs', shape=f.shape, nderiv=f._nderiv)
+        return function.Argument(name='meshdofs0', shape=f.shape, nderiv=f._nderiv)
+    if isinstance(f, function.Argument) and f._name == 'meshdofs0':
+        return function.Argument(name='meshdofs00', shape=f.shape, nderiv=f._nderiv)
+    if isinstance(f, function.Argument) and f._name == 'meshdofs00':
+        return function.Argument(name='meshdofs000', shape=f.shape, nderiv=f._nderiv)
 
 # some helper function to shift variables by two timesteps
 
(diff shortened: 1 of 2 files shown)


uekerman:
Thanks for the nice modification and documentation :pray: 

I would indeed suggest to split the PR into two. The fluid modifications and the new solid participant. The latter should be quick to merge. The fluid modifications are a bit tricky. Nice job organizing the changes in simple commits. This way it should be easy to pick.

> Beyond that the formulation is progressively simplified to the point of requiring only the previous solution, rather than the past four.

The old formulation was complicated indeed. We created it when we wanted to test the waveform relaxation in preCICE. For this, we needed a clean second-order fluid participant, including the mesh movement and the force computation. Back then, I experimented a lot and could not find a simpler solution. I did not yet look at all details, but I am afraid that the new formulation is no longer second order in time.

@gertjanvanzwieten Could you please remove the last commit here and move it to a new PR?

gertjanvanzwieten:
I moved the last commit to #431, but made it a draft for reasons explained there.

As for the formulation, I wasn't sure whether the formulation was set in stone, hence the commit layout as it is. Naturally these are all just suggestions.

That said, the solution doesn't seem change much looking at plot-displacement.sh, and if it does then there is still this idea with involving the former `F` that might restore 2nd order for the traction evaluation to experiment with. This would involve changing line 72 to:
```python
resF = res + tθ(couplinginterface.sample('gauss', 4).integral('ubasis_ni F_i d:x' @ ns))
```
Note the added `tθ`. It does induce oscillations which is why I left it out, but code wise it's trivial. For the formulation itself I'm not certain that this is not still 2nd order.

gertjanvanzwieten:
Hi, I removed my earlier fork of this repo to work around github's incessant refusal to push #648 to it, but didn't realize that this PR was still pointing from it. I don't seem to have my own copy of it anymore and I don't know how to restore it from github, even though the commits are still viewable, so if somebody else kept a copy let me know and we can restore the commits from there. Assuming there is still interest in them.

Edit: I managed to pull the final state of fluid.py and solid.py so we could work with that if it comes to it.

MakisH:
@IshaanDesai seems to have a copy of the branch around. I guess just push the branch here and then maybe @gertjanvanzwieten can then edit the PR and change the source branch, or create a new one.

IshaanDesai:
I pushed the branch directly to the organizational repository: https://github.com/precice/tutorials/tree/perp-flap-nutils as pushing to this pull request HEAD reference was not working. @gertjanvanzwieten you can sync your fork with the upstream and then you should have the branch again.

gertjanvanzwieten:
On it, thanks!

gertjanvanzwieten:
I don't think I can reopen this PR so I'll make a new one. Is it useful if I rebase to develop?

IshaanDesai:
> Is it useful if I rebase to develop?

I would say yes.

PR #430, opened 2024-01-08
Specify right repository

valentin-seitz:
Since this workflow is reusable and can be called my multiple repos and default value is ${{ github.repository }}. This poses a problem as the action will check out the calling repository, which is not what we want if the calling repo is e.g. the openfoam-adapter.


MakisH (approved):
Makes sense!

PR #431, opened 2024-01-09
add nutils solid participant

gertjanvanzwieten:
This PR adds a solid participant compatible only with Nutils 8, formerly part of #429.

It's a draft for now because it was discovered that the system has more damping than reference results, potentially resulting from the fully implicit first order formulation. I will do some more testing in the coming days to try and sort this out.

### DIFF
--- a/perpendicular-flap/README.md
+++ b/perpendicular-flap/README.md
@@ -37,6 +37,8 @@ Solid participant:
 
 * DUNE. For more information, have a look at the [experimental DUNE adapter](https://github.com/precice/dune-adapter) and send us your feedback.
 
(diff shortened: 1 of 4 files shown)


uekerman:
This addition goes indeed to `master` as it still uses preCICE v2.

MakisH (approved):
I have not tested this, but the changes make sense.
Merging to master is fine, but please remember backporting to develop.

MakisH:
> Linter problem seems unrelated:
> 
> ```
> MD034/no-bare-urls Bare URL used [Context: "www.openfoam.com"]
> ```

FYI, the solution to this is disabling the `MD034` linter rule in this file:

https://github.com/precice/tutorials/blob/14197c235d4a72d841b04d01446a21a227cf0639/.markdownlint.json

We already do that in develop:

https://github.com/precice/tutorials/blob/develop/.markdownlint.json

Since the `master` is not going to stay on for long, I guess it is not worth fixing and all the merging around.

Issue #340, opened 2023-05-22
Elastic-tube-1d solid cpp x coordinate bug


fsimonis:
The elastic tube 1d solid solver uses the chunksize directly as x coordinates instead of splitting the actual pipe:
```
0 1 2 3 4 ... 100 
```
instead of 
```
0 0.1 0.2 ... 10
```

https://github.com/precice/tutorials/blob/0482e5b4d3e1604dd031d2b113a2aaa63733c9a8/elastic-tube-1d/solid-cpp/src/SolidSolver.cpp#L40-L44

fsimonis:
Fixed by https://github.com/precice/tutorials/pull/434

Issue #395, opened 2023-11-18
Compatibility of elastic-tube-1d with preCICE v3


MakisH:
This case is still in v2: https://github.com/precice/tutorials/blob/develop/elastic-tube-1d/fluid-cpp/src/FluidSolver.cpp

Output from #396:

```
 Network elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138_default  Creating
 Network elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138_default  Created
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-prepare-1  Creating
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-prepare-1  Created
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1  Creating
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-solid-python-1  Creating
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-solid-python-1  Created
 Container elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1  Created
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     | /runs/elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138/fluid-cpp/src/FluidSolver.cpp: In function 'int main(int, char**)':
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     | /runs/elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138/fluid-cpp/src/FluidSolver.cpp:72:29: error: invalid conversion from 'const char*' to 'int' [-fpermissive]
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     |    72 |   interface.setMeshVertices(meshName, chunkLength, grid.data(), vertexIDs.data());
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     |       |                             ^~~~~~~~
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     |       |                             |
elastic-tube-1d_fluid-cpp-solid-python_2023-11-18-125138-fluid-cpp-1     |       |                             const char*
[...]

PR #434, opened 2024-01-10
Port elastic-tube-1d solvers to version 3

fsimonis:
This PR ports the elastic-tube-1d cpp solvers to version 3

This also rearranges the solvers to follow the pattern read data -> solve -> write data -> advance.

Also fixes the incorrect coordinates in solid-cpp #340

### DIFF
--- a/elastic-tube-1d/fluid-cpp/CMakeLists.txt
+++ b/elastic-tube-1d/fluid-cpp/CMakeLists.txt
@@ -15,7 +15,7 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STRE
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wno-unused-parameter")
 endif()
 
-find_package(precice REQUIRED CONFIG)
+find_package(precice 3 REQUIRED CONFIG)
 find_package(LAPACK REQUIRED)
 
 add_executable(FluidSolver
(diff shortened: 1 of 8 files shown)


fsimonis:
Related to https://github.com/precice/tutorials/issues/395

valentin-seitz:
I just started a small system test job: https://github.com/precice/tutorials/actions/runs/7492877499/job/20397361340
And apparently, field-compare reports some major differences between the v2.5 reference data (we use the exchanged data between the participants) and the current status. (see stdout.log in the respective folders)
I just wanted to share this. It could also be a bug in the tests or some v2 vs. v3 hiccup.

IshaanDesai (approved):
I ran several combinations of the C++ and Python participants and everything seems to be working.

fsimonis:
@valentin-seitz The solid-cpp solver was bugged and produced very bad results before 7de3813d2931c9fa9ad75db38f25faa9466ec5b4.
Could that be it.

Solid Cpp | Solid Python
--- | ---
1d-sc-fc | 1d-sp-fc

fsimonis:
@valentin-seitz I checked the uploaded run of the workflow you linked to.

For the solid-python fluid-python case, fieldcompare fails, but the exports are close to identical apart from some minor differences. I checked provided meshes only. Is there a way to get more detailed information on what differs?

fsimonis:
I ran the version 2 of the elastic-tube-1d with the watchpoint at x=5 on the Fluid Solver and then plotted them using the state of #438 (this PR + #435 + watchpoints and plotting).

The result is this:

tutorials-elastic-tube-1d-all

The results of the v2 python solid match the results of this PR.
The buggy (too coarse) solid solver of v2 leads to different results (larger max diameter in oscilations)

PR #436, opened 2024-01-11
Fix: Participants could not find each other during the systemtests

valentin-seitz:
This PR introduces a fix that enables to rewrite the `precice-config.xml `for both v2 and v3 versions. 
Since the word “from” does not appear, sed should now be able to set the network used for systemtests accordingly.


MakisH (approved):
Since you are editing this file, you could also add a line break after the added `export:vtk` tag (should actually better be `export:vtu`, as that behaves the same way in v2.5.0 and v3) and the closing `</participant>`.

Issue #402, opened 2023-11-19
System tests: Restrict preCICE time for selected tests


MakisH:
Some tests are expected to be taking too long to run. For such cases, it would be nice to have an option to overwrite the `max-time` in the `precice-config.xml`, both for running the test and for the reference results. Something like this in the `tests.yaml`:

```yaml
  openfoam_adapter_pr:
    tutorials:
      - path: flow-over-heated-plate
        case_combination:
          - fluid-openfoam
          - solid-openfoam
        reference_result: ./flow-over-heated-plate/reference-data/fluid-openfoam_solid-openfoam.tar.gz
        max_time: 0.5
```

which would then perform the following change:

```diff
- <max-time value="1" />
+ <max-time value="0.5" />
```

valentin-seitz:
I see the use case, also considering filesize of the generated artifacts. 
My proposal would look be to do the following: 
```yaml
openfoam_adapter_pr:
    tutorials:
      - path: flow-over-heated-plate
        case_combination:
          - fluid-openfoam
          - solid-openfoam
        reference_result: ./flow-over-heated-plate/reference-data/fluid-openfoam_solid-openfoam.tar.gz
        precice_config_override:
          coupling-scheme:
            max-time: value="1"

```
I think this would provide a bit more semantical support in terms of what `max-time` means in that context (namely overriding the precice_config.xml). Another benefit would be the added flexibility (Maybe we want to change the max-iterations too?)

MakisH:
The `precice_config_override` sounds even better! But probably harder to implement, and difficult to argue why some options are available and others not.

If you know how to do this, then it sounds to me like the way to go. :+1:

PR #437, opened 2024-01-11
Add a release test suite

valentin-seitz:
Adds a test suite covering, what we currently can cover from the table in https://github.com/precice/precice/pull/1914


valentin-seitz:
Perpendicular flap (fluid-nutils, solid-calculix) could not be executed because of: https://github.com/precice/tutorials/issues/217

For now removing that test case

valentin-seitz:
Removing the Elastic tube 3d until https://github.com/precice/tutorials/issues/402 is implemented, due to very long runtime > 10min

tools/tests/tests.yaml
MakisH, line 1: As far as I understand, this is the only relevant file. Do we need to update the reference results here, since we are not changing the tests themselves?

I see a lot of duplication in the configuration, but I am not sure yet how to prevent that. Let's continue with the duplicate mode until we get a more complete picture of tests.

tools/tests/tests.yaml
```diff
@@ -51,3 +51,25 @@ test_suites:
           - fluid-cpp
           - solid-python
         reference_result: ./elastic-tube-1d/reference-data/fluid-cpp_solid-python.tar.gz
+  release_test:
```
MakisH, line 54: Let's document how to trigger this specific one (from the perspective of Lisa) as an example in https://precice.org/website-testing/dev-docs-system-tests.html#running-the-system-tests

MakisH (approved):
We definitely need such a test suite, even if currently it only tests two tutorials. Let's merge it now, so that we can collectively add more tutorial cases here.

MakisH:
> Perpendicular flap (fluid-nutils, solid-calculix) could not be executed because of: #217
> 
> For now removing that test case

Well, a failing test case is also a useful test case!

This is kind-of related to #431

tools/tests/tests.yaml
MakisH: Added.

tools/tests/tests.yaml
MakisH: Let's not worry about this in this PR.

MakisH (approved):
Merging to extend. Thanks for providing the foundation!

PR #435, opened 2024-01-11
Add elastic-tube-1d rust solvers

fsimonis:
This PR is based on #434 and adds elastic-tube-1d rust solvers.

As long as cargo is installed, they can be run solely by running `./run.sh`.
VTK output and plotting is supported.

Issue #426, opened 2023-12-17
Would be nice to have visualize.sh scripts


MakisH:
In cases where it is complicated to visualize results, we currently have some `plot.sh` or similar scripts that rely on Gnuplot, matplotlib, etc.

Since most visualization tools we use have some kind of CLI / API, it would be possible to provide `visualize.sh` scripts, at least per case.

My motivation is that I find that I click too much on ParaView every time I want to check something simple and this could save some of my attention/time.

valentin-seitz:
I think this could also add a lot of value to the systemtests.
Especially if we keep the scripts flexible enough to also work on fields generated by the `--diff` option of fieldcompare.

PR #438, opened 2024-01-12
Add elastic-tube-1d watchpoint plot

fsimonis:
This PR is based on #434 and #435.
It adds a watchpoint at x=5 of the fluid solver as well as a plot of for pressure and diameter over all time steps.
Scripts run-all.sh and plot-all.sh generate this plot.

image


elastic-tube-1d/run-all.sh
MakisH, line 1: What is the purpose of this script?
We currently don't use something like this anywhere else. We previously had such `Allrun` scripts, which we decided to remove, as they typically led to more confusion (what is happening? what step failed? can I still get the results I need?) than helping.

elastic-tube-1d/plot-all.sh
MakisH, line 1: This is nice! I did not test it, but could we ensure that it just prints whatever is available, without having to first run aaaaaaalll the cases?

MakisH (commented):
Related to (or in the same context as) #426.

Still needs some rebasing, as it includes changes related to porting to v3.

elastic-tube-1d/run-all.sh
fsimonis: We need some way of generating data of the cross-product of solid and fluid participants.

I am not sure how else to generate the data for plot-all.sh.

Maybe `plot-all.py --generate` to setup and run all cases?
Suggestions welcome of course.

elastic-tube-1d/plot-all.sh
fsimonis: Sure, but that would require additional logic in the script.
Currently, it relies on `run-all.sh`.

elastic-tube-1d/run-all.sh
MakisH: > Maybe `plot-all.py --generate` to setup and run all cases? Suggestions welcome of course.

Sounds like we are reversing the flow (setup -> run -> visualize becomes setup -> visualize(run)).

I would currently run these (or only the ones I want to compare) one by one. Do we need to manually execute all of them all the time? If it is just for occasional testing, then a `--generate` option here would be appropriate.

If we wanted an automatic workflow, then I would define the respective combinations in that workflow. But then is starts being relevant for the system tests.

elastic-tube-1d/plot-all.sh
MakisH: I think that gnuplot will just skip these, no? I remember getting such warnings from some tutorial.

fsimonis:
I removed the run-all script and changed `plot-all.sh` to plot the watchpoint of all fluid solvers.

This omits the solver participant from the plots, but should be practical enough to detect problems.

elastic-tube-1d/plot-all.sh
fsimonis: I reduced it to only plotting the fluid watchpoints. So there is no more cross-product going on.

elastic-tube-1d/README.md
```diff
@@ -122,6 +122,13 @@ python3 plot-vtk.py diameter fluid-python/output/out_fluid_
 
 ![FSI3 setup](images/tutorials-elastic-tube-1d-diameter.png)
 
+## Comparing different Fluid results
+
+The Fluid participant defines a watchpoint at `x=5`, which records `pressure` and `diameter` for every timestep.
+To compare the results of the various Fluid participants, you can run them all and plot the watchpoints using `plot-all.sh`.
+
+![Plot comparing pressure and diameter of all Fluid participants](images/tutorials-elastic-tube-1d-all.png)
```
MakisH, line 130: Should not matter (much) here, but we should also document with which Solid participant these results were acquired.

MakisH (approved):
Makes sense, but I have not tried if it works. Please also document the Solid participant used in the plot.

fsimonis:
It's not possible to get the solid participant from the run files. So hard coding a name can easily become incorrect.

Not sure how to solve this.

MakisH:
> It's not possible to get the solid participant from the run files. So hard coding a name can easily become incorrect.
> 
> Not sure how to solve this.

You checked in a plot, which you generated after running some combinations, right? I am only asking that you document which Solid participant you used, nothing more.

elastic-tube-1d/README.md
fsimonis: Done

fsimonis:
FYI the new image looks like this:

image

PR #442, opened 2024-01-16
Optimize Upload times for systemtests

### LINKED ISSUE #425: Pack artifacts before uploading to GitHub
While our caching and runs are now working rather efficiently, we still have very long archiving time. For [example](https://github.com/precice/tutorials/actions/runs/7214868869/job/19657898385), this job used cached Docker images, ran the a test suite for 2m24s, and then it took more than 10m to compress and upload the results.

Since we have a more capable VM, we could first bundle together all files into one, and then upload them.

valentin-seitz:
Apparently we have been using a wrong version here: 
https://github.com/actions/upload-artifact/issues/199#issuecomment-1862996650

This should fix performance problems for now without complicating the workflow too much. 

I would consider it a fix for https://github.com/precice/tutorials/issues/425

### DIFF
--- a/.github/workflows/run_testsuite_workflow.yml
+++ b/.github/workflows/run_testsuite_workflow.yml
@@ -63,7 +63,7 @@ jobs:
         cd ../../
     - name: Archive run files
       if: ${{  failure() || inputs.upload_artifacts == 'TRUE' }}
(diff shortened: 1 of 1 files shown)


MakisH (approved):
Excellent! Let's merge and observe if we still have such issues.

PR #443, opened 2024-01-17
Fix subcycling

BenjaminRodenberg:
I updated the tutorials to follow the pattern suggested in our adapterexample. In some cases, the implementation of subcylcling was also buggy.

@uekerman, @IshaanDesai : Especially a `dt = min(dt, precice_dt)` is very dangerous in the context of subcycling and can lead to errors that are very hard to find. I did not run the cases that I touched. Please do a quick check for the cases that you feel responsible for.

### DIFF
--- a/perpendicular-flap/fluid-nutils/fluid.py
+++ b/perpendicular-flap/fluid-nutils/fluid.py
@@ -120,6 +120,8 @@ def main(inflow: 'inflow velocity' = 10,
 
     # initialize preCICE
     participant.initialize()
+    precice_dt = participant.get_max_time_step_size()
+    dt = min(precice_dt, timestepsize)
 
     # boundary conditions for fluid equations
     sqr = domain.boundary['wall,flap'].integral('urel_k urel_k d:x0' @ ns, degree=4)
--- a/perpendicular-flap/fluid-nutils/fluid.py
+++ b/perpendicular-flap/fluid-nutils/fluid.py
@@ -120,8 +120,6 @@ def main(inflow: 'inflow velocity' = 10,
 
     # initialize preCICE
     participant.initialize()
-    precice_dt = participant.get_max_time_step_size()
-    dt = min(precice_dt, timestepsize)
 
     # boundary conditions for fluid equations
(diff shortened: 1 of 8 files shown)


uekerman (commented):
Looks meaningful to me. I did not, however, not run any cases.

perpendicular-flap/fluid-nutils/fluid.py
```diff
@@ -120,8 +120,6 @@ def main(inflow: 'inflow velocity' = 10,
 
     # initialize preCICE
     participant.initialize()
-    precice_dt = participant.get_max_time_step_size()
-    dt = min(precice_dt, timestepsize)
```
fsimonis, line 124: Without this I get:

NameError: cannot access free variable 'dt' where it is not associated with a value in enclosing scope

BenjaminRodenberg: Thanks! I assume this is related to some of the lambda functions further above being evaluated for some initialization steps executed before entering the main time loop. I will revert this change. Then we obviously need to do this step here.

BenjaminRodenberg: I compute the `dt` directly after `initialize` to be on the safe side. This is generally allowed and does not harm. Maybe it is unnecessary in some places, but it is hard for me to uncover the possible indirections here and this should be the safe option. See [67093b9](https://github.com/precice/tutorials/pull/443/commits/67093b97edba9d02c94649fab86d1ce1e118ce02).

MakisH (approved):
The current state definitely looks buggy, good catch! The changes make sense to me (separating the `dt` from the `solver_dt` and moving the `read_data` after the writing the checkpoint).

I do not feel responsible for any particular case, but the channel-transport seems to ~~at least start and run~~ work and produce the expected results.

IshaanDesai (approved):
I ran the channel-transport and flow-over-heated-plate cases with the Nutils participants which have the changes, and the results in ParaView look the way they are supposed to look. I also ran the first few time steps of the two-scale-heat-conduction and things seem to be in order.

PR #451, opened 2024-01-29
Use endTime = 5 like max-time in precice-config.xml

BenjaminRodenberg:
I'm not sure if it really matters, because preCICE does the steering, but having a different `endTime` here looks strange to me.

### DIFF
--- a/perpendicular-flap/fluid-openfoam/system/controlDict
+++ b/perpendicular-flap/fluid-openfoam/system/controlDict
@@ -15,7 +15,7 @@ startTime       0;
 
 stopAt          endTime;
 
-endTime         10;
+endTime         5;
 
 deltaT          0.01;
 


MakisH (approved):
It does not matter for the OpenFOAM cases (the adapter overwrites it to infinity and preCICE takes over), but it is indeed nicer to be consistent.

Could you maybe take a look at all other OpenFOAM cases as well?

BenjaminRodenberg:
> Could you maybe take a look at all other OpenFOAM cases as well?

I don't really have the time to go through all the cases and figure out the time stepping setup. The perpendicular flap case I know pretty well. The others would need some more time from my side. Feel free to do it. I agree that it's useful. Just wanted to quickly contribute the straight-forward fix and not start a bigger work package here.

MakisH:
I looked at all other tutorials and updated the OpenFOAM cases.
Again: this is just to look consistent. The `endTime` does not matter anywhere in the OpenFOAM adapter, as the adapter overwrites that to infinity and uses the `isCouplingOngoing()` to trigger an end.

I only left the elastic-tube-3d, as I was not immediately sure what the correct value should be.

BenjaminRodenberg:
> I looked at all other tutorials and updated the OpenFOAM cases. Again: this is just to look consistent. The `endTime` does not matter anywhere in the OpenFOAM adapter, as the adapter overwrites that to infinity and uses the `isCouplingOngoing()` to trigger an end.
> 
> I only left the elastic-tube-3d, as I was not immediately sure what the correct value should be.

Thank you for taking care!

PR #452, opened 2024-01-29
Solve compatibility issue by using pandas newer version and matplotlib for plotting

Fujikawas:
For the newer `pandas` version(2.2.0 for my instance), it requires the data to be converted to a numpy array before indexing for plotting with `matplotlib.pyplot.plot()`. Otherwise it gives error like:
```
ValueError: Multi-dimensional indexing (e.g. `obj[:, None]`) is no longer supported. Convert to a numpy array before indexing instead.
```
The `plot-trajectory.py` in the tutorial case `oscillator` is this case.

### DIFF
--- a/oscillator/plot-trajectory.py
+++ b/oscillator/plot-trajectory.py
@@ -19,13 +19,13 @@ filename = args.csvFile
 df = pd.read_csv(filename, delimiter=';')
 
 if args.plotType == PlotType.U_OVER_T.name:
-    plt.plot(df['time'], df['position'])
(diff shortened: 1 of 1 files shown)


Fujikawas:
Haven't tested if this works for older version yet. Maybe it's better to add a conditional clause here to decide if the conversion is needed.

BenjaminRodenberg (approved):
No need for backwards compatibility. This is a helper script and I think we can expect the (few) users of this script to have a recent version.

Issue #457, opened 2024-01-31
Perpendicular-flap solid-nutils case cannot start first (interruption signal)


MakisH:
Discovered in the context of https://github.com/precice/precice/pull/1914#issuecomment-1919099949 Copying here:

{fluid-openfoam:1, solid-nutils:1} fails consistently with some network error if I first start Nutils and then OpenFOAM.

```
Successfully installed appdirs-1.4.4 bottombar-2.0.2 mpi4py-3.1.5 numpy-1.26.3 nutils-8.0 nutils-poly-1.0.0 psutil-5.9.8 pyprecice-3.0.0.0.dev2 stringly-1.0b3 treelog-1.0 typing-extensions-4.9.0
arguments > young=4000000.0
arguments > density=3000.0
arguments > poisson=0.3
arguments > nelems=2
arguments > solver_dt=0.01
arguments > npoints_per_elem=3
start 2024-01-31 17:20:45
optimize > constrained 6/126 dofs
optimize > optimum value 0.00e+00
optimize > optimum value 0.00e+00
---[precice]  This is preCICE version 3.0.0
---[precice]  Revision info: v2.5.0-604-g150d4ee8
---[precice]  Build type: Debug
---[precice]  Configuring preCICE with configuration "../precice-config.xml"
---[precice]  I am participant "Solid"
---[precice]  Setting up primary communication to coupling partner/s
---[precice] ERROR:  Accepting a socket connection at 127.0.0.1:41725 failed with the system error: accept: Interrupted system call
```

:warning: **UPDATE:** It looks like this only works if I first start OpenFOAM (first participant) and then Nutils (second participant). Previously, I was starting Nutils first.

- I consistently reproduce this on my system: starting with Nutils breaks, starting with OpenFOAM works
- I cannot replicate this issue with OpenFOAM-CalculiX of the same tutorial
- I cannot replicate this issue with OpenFOAM-Nutils in the channel-transport
- I cannot replicate this issue with the Python solverdummies.
- I do get the same error if I try to (indeed) interrupt Nutils during initialization. But why does this always happen for me in the perpendicular-flap?
- @IshaanDesai reproduced this.

MakisH:
Related to `bottombar`, a dependency of Nutils. We need to adapt the `requirements.txt`.

https://github.com/precice/tutorials/pull/431#pullrequestreview-1812711006

uekerman:
[...]

PR #458, opened 2024-01-31
Add bottombar to nutils cases

fsimonis:
This PR adds bottombar to nutils cases to prevent a signal interrupt in preCICE.
Only exception is the nutils 6 case.

Original solution from here https://github.com/precice/tutorials/pull/431#pullrequestreview-1812711006


fsimonis:
Closed after reading https://github.com/precice/tutorials/issues/457#issuecomment-1919767101

PR #460, opened 2024-02-07
Several compatibility updates to the two-scale-heat-conduction tutorial

IshaanDesai:
This PR includes a bunch of updates for the two-scale-heat-conduction tutorial, to make it compatible with the latest state of the Micro Manager. The Micro Manager will be released soon, which will finalize the changes here.


two-scale-heat-conduction/micro-dumux/appl/micro_sim.cpp
mathiskelm, line 1: Need to adapt the pybind wrapper to the constructor taking an argument `simulationID`, i.e. in the wrapper's constructor and `setstate` functions.

mathiskelm (approved):

PR #462, opened 2024-02-07
Update Nutils requirements.txt

MakisH:
Most of the cases already specify `nutils==7` or similar. This means that they will take the latest version of Nutils 7.x. If needed, I can change that to `nutils>=7`, but I would not suggest that, because:

a) with an `==`, we are relatively sure that the case will work in the future, as long as Nutils follows semantic versioning.
b) this is anyway in a virtual environment. If we install Nutils 7 here, we don't exclude other cases from using Nutils 8.

At the end of the day, we need to accurately convey what we need. If we need "Nutils 7 or newer", and we have reasons to expect that this case will also work with Nutils 10 in a few years, then we should change it to `>=`.

Other than that, this PR:

- specifies the nutils version in `perpendicular-flap/solid-nutils` to be `>=8.5`. This will accept 8.5, 8.6, 9.0, and later. We make the assumption here that versions are **forward compatible**. The alternative, and consistent with the rest, would be: `>=8.5,9`.
- relaxes the restriction for pyprecice from `==3.0.0.0` to `==3`, as I would expect is intended, since we do follow semantic versioning.
- adds a `requirements.txt` for the `channel-transport/transport-nutils` case.

I have verified that all cases work with the provided versions.

@uekerman I mainly need a decision on the `==` vs `>=`. I would vote for the current state and see what happens with the `>=8.5` later on, even though it is inconsistent with the rest.


uekerman (approved):
Looks good.
We will need to update these in the future, but for the moment everything should be fine.

Note: For most cases (or even all), Nutils 8 (8.5) should work as well. But we cannot write `>=7` as for `8<=x<=8.4`, we run into the bottombar issue.

MakisH:
> We will need to update these in the future

Here is the beauty of it: we don't need to update these ever, as long as the respective version exists in PyPI. Since Nutils is installed in a virtual environment, everything is local to each tutorial and easy to adapt.

We can, of course, update to use more recent Nutils features.

Do I misunderstand something?

PR #463, opened 2024-02-12
Fix images on two-scale-heat-conduction case

uekerman:
Otherwise, not shown when rendered on website.


uekerman:
Did not help :/

PR #465, opened 2024-02-21
Add FMU participants to oscillator tutorial

uekerman:
v3-compatible replacement of #322

Issue #461, opened 2024-02-07
Inconsistencies in structure of tutorials


MakisH:
Follow-up of https://github.com/precice/tutorials/issues/228. The structure is documented in [the website](https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial) and clarified in https://github.com/precice/precice.github.io/pull/331.

The following tutorials need updates:

- oscillator @uekerman 
  - [x] python
  - [x] fmi
- partitioned-heat-conduction
  - [x] fenics: https://github.com/precice/tutorials/pull/459
  - [x] fenicsx (move from develop to branch) @IshaanDesai 
  - [x] nutils ~~@uekerman~~ @IshaanDesai https://github.com/precice/tutorials/pull/496
- partitioned-heat-conduction-complex
  - [x] fenics
- partitioned-heat-conduction-direct ~~@uekerman~~ @Fujikawas
  - [x] nutils
- volume-coupled-diffusion @BenjaminRodenberg 
  - [x] fenics https://github.com/precice/tutorials/pull/489

I would still consider the `aste-turbine` to be a very special case that is not meant to be extended or repeated in other tutorials.

Please comment regarding which cases you would like to port. No need to port every tutorial at once, but let's try to have a consistent structure until the next release of the tutorials.

BenjaminRodenberg:
I don't want to prioritize this before my thesis is submitted or at least close to submission. I think this is also something that could be easily done by a research assistant and not urgent.

MakisH:
You mean that you don't want to work on it, or that you don't want that the structure changes?

BenjaminRodenberg:
I'm fine with the structural changes and we can also implement them rather sooner than later, but I don't have time to do it myself now.

If I stumble across a point where I can briefly do it anyway, I will do it occasionally (e.g. https://github.com/precice/tutorials/commit/459ab90fd803de85a5c370430a11693277d70c83).

MakisH:
Thank you, I can also take over work packages on this.

PR #466, opened 2024-02-21
Add FMU participants to oscillator tutorial

uekerman:
v3-compatible replacement of #322
Using my own branch to have write access, instead of #465 

Ready for review. Should work with the `main` branch of the FMI Runner (v0.2.0 will be released before merging here).

Does still ignore #461. I would suggest to do the `oscillator` after merging here. Otherwise, things get complicated here.

I ran the cases with the precompiled FMU, I did not try to recompile on any other system. @LeonardWilleke compiled the FMU.

The FMU cases give a higher error than the Python implementation (but converge for smaller time step sizes). We still don't understand why exactly IIRC, but this is consistent with the thesis and paper from @LeonardWilleke.


.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
 repos:
 # Official repo for default hooks
 - repo: https://github.com/precice/precice-pre-commit-hooks
-  rev: 'v3.1'
+  rev: 'v3.3'
   hooks:
   - id: format-precice-config
 - repo: https://github.com/igorshubovych/markdownlint-cli
-  rev: v0.39.0
+  rev: v0.30.0
```
uekerman, line 8: This diff will disappear once #468 is merged.

oscillator/thirdparty/LICENSE.txt
uekerman, line 1: We link to this file from `fmi/fmu/README.md`.
OK like this?

uekerman: We could probably move the subfolder `thirdparty` into `fmi/fmu`.

oscillator/README.md
```diff
@@ -17,19 +17,16 @@ This tutorial solves a simple mass-spring oscillator with two masses and three s
...
+- *FMI*: A solver using the [preCICE-FMI runner](https://github.com/precice/fmi-runner) (requires at least v0.2). The Runner executes the FMU model `Oscillator.fmu` for computation. The compiled FMU model for Linux is part of this repository. For other systems, please recompile the model from the provided [C-files](https://github.com/precice/tutorials/tree/master/oscillator/fmi/fmu). If you want to change the model parameters or the initial conditions of the simulation, have a look inside the setting files for [MassLeft](https://github.com/precice/tutorials/tree/master/oscillator/fmi/MassLeft) and [MassRight](https://github.com/precice/tutorials/tree/master/oscillator/fmi/MassRight). For more information, please refer to [2].
 
 ## Running the Simulation
```
uekerman, line 27:
```suggestion
## Running the simulation
```

.pre-commit-config.yaml
MakisH: We can still keep the upgrade to v3.3 here.

oscillator/thirdparty/LICENSE.txt
MakisH: Yes, let's move it (I can do it).
It currently sounds like the `thirdparty/` and the licenses also apply to the `python` case.

oscillator/README.md
```diff
@@ -39,19 +36,21 @@ cd python
 and
 
 ```bash
-cd python
+cd fmi
 ./run.sh -r
```
MakisH, line 40: I cannot run this case at the moment, it looks like some kind of configuration issue:

```
~/repos/precice/tutorials/oscillator/fmi [create-fmu-oscillator-v3]$ ./run.sh -r
  File "/home/gc/.local/bin/fmiprecice", line 33, in <module>
    sys.exit(load_entry_point('fmiprecice', 'console_scripts', 'fmiprecice')())
  File "/home/gc/repos/precice/fmi-runner/fmiprecice/runner.py", line 54, in main
    model_description = read_model_description(path)
  File "/home/gc/.local/lib/python3.10/site-packages/fmpy/model_description.py", line 889, in read_model_description
    raise ValidationError(problems)
fmpy.model_description.ValidationError: Failed to validate modelDescription.xml:
- The combination causality="parameter" and variability="tunable" in variable "mass.a" (line 69) is not allowed.
```

I am using the fmi-runner from `main`/`2e3ab1cd2` (March 7).
I get the same both with the default `Oscillator.fmu`, as well as if I copy my self-build FMU to its place.

@LeonardWilleke any idea?

oscillator/fmi/Oscillator.fmu
MakisH, line 1: I could build the FMU very easily on my system. It does not seem to specify any dependencies. What is the motivation behind providing pre-built binaries for the FMU files?

I am just worried of setting unneeded precedences and potential inconsistencies.

oscillator/fmi/fmu/Oscillator/FMI1CS.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<fmiModelDescription
+  fmiVersion="1.0"
+  modelName="Oscillator"
+  modelIdentifier="Oscillator"
+  description="This model is used for the partitioned simulation of an oscillator with three springs and two masses. The oscillator is cut in the middle spring for partitioning. Here, two springs with one mass between them are calculated."
+  generationTool="Reference FMUs (development build)"
+  guid="{1AE5E10D-9521-4DE3-80B9-D0EAAA7D5AF1}"
```
MakisH, line 8: Same comment as in https://github.com/precice/tutorials/pull/474: What is this? Looks user-specific.

oscillator/fmi/fmu/Oscillator/readme.html
MakisH, line 1: Same as in https://github.com/precice/tutorials/pull/474: Do we need both the HTML and the markdown variants?

oscillator/fmi/fmu/README.md
```diff
@@ -0,0 +1,19 @@
+# Build instruction
+
+To build the Oscillator FMU you need [CMake](https://cmake.org/) &GreaterEqual; 3.16 and a supported [build tool](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html) e.g. Visual Studio &GreaterEqual; 2013 , Xcode or make.
+
+Generate the build files with the following commands:
+
+```bash
+mkdir build
+cd build
+cmake -DFMI_TYPE=CS -DFMI_VERSION=3 ..
```
MakisH, line 10: Maybe explain what these options mean, what restrictions are here, and what alternatives?

oscillator/fmi/run.sh
MakisH, line 1: Based on https://github.com/precice/tutorials/issues/461, I would just split this `run.sh` into one smaller and simpler `run.sh` for each of the `MassLeft` and `MassRight` (and generally restructure this directory, which can come in a follow-up PR).

But this case is already in a very good state to be very easily restructured into the expected form.

MakisH (commented):
I tried running this tutorial, but I faced some configuration issue, which should be easy to fix.

In terms of structure, it needs some changes to address https://github.com/precice/tutorials/issues/461, but this can happen in a follow-up.

oscillator/fmi/fmu/README.md
uekerman: These variables are standard in FMI. `FMI_TYPE` is the type of interface, we need "co-simulation" here. `FMI_VERSION` is the version of the standard, "3" is the newest version. There are no real alternatives, should always be built like this.

MakisH: If these are very standard in FMI and I would know if I knew anything about FMI, then we don't need to write anything in the `README.md`. If these are something we define, then they would need some documentation.

I am just inclined to always have more information than less in anything that might help us with future maintenance.

oscillator/fmi/Oscillator.fmu
uekerman: Removed binary and replaced with automatic building in run script.

oscillator/fmi/fmu/README.md
uekerman: I think we are good here. Both are among the first things somebody learns when studying FMI. In the end, I knew them :D

oscillator/fmi/run.sh
MakisH: No worries, I can handle this in a separate PR.

oscillator/README.md
uekerman: In FMI3.xml, there is

```xml
    <Float64
      name="mass.a"
      valueReference="4"
      causality="parameter"
      variability="tunable"
      initial="calculated"
      unit="m/s2"
      description="Acceleration of the body" />
```
so, the error message make sense. Seems to be a problem of the FMU.

I don't get the error, however. Which version of FMpy are you running? I have `FMPy==0.3.13`.

MakisH: I have 0.3.20.

uekerman: With that version, I can reproduce the problem.

BenjaminRodenberg: This looks to me like a possible solution:

https://github.com/CATIA-Systems/FMPy/issues/189#issuecomment-723925332

Did you try this out?

uekerman: When I switch to

```xml
<Float64
      name="mass.a"
      valueReference="4"
      causality="parameter"
      variability="tunable"
      initial="exact"
      start="0"
      unit="m/s2"
      description="Acceleration of the body" />
```

I run into the exception

```
Implicit coupling not possible because FMU model can't reset state. ...
```

even though in `modelDescription.xml`, I get

```xml
<CoSimulation
    modelIdentifier="Oscillator"
    canGetAndSetFMUstate="true"
...
```

Commenting out the exception, everything runs through and gives the correct results. I am puzzled :/

uekerman: > This looks to me like a possible solution:
> 
> [CATIA-Systems/FMPy#189 (comment)](https://github.com/CATIA-Systems/FMPy/issues/189#issuecomment-723925332)
> 
> Did you try this out?

Seems to solve the original problem :+1: 

But the `canGetAndSetFMUstate` problems remains :exploding_head:

oscillator/fmi/fmu/Oscillator/FMI2.xml
```diff
@@ -0,0 +1,167 @@
...
+      variability="fixed"
+      initial="exact"
+      description="Mass of the body">
+      <Real start="1" unit="kg" />
+    </ScalarVariable>
+    <ScalarVariable
+      name="mass.u"
+      valueReference="2"
+      causality="parameter"
+      variability="tunable"
```
BenjaminRodenberg, line 80:
```suggestion
      causality="local"
      variability="tunable"
```
See https://github.com/CATIA-Systems/FMPy/issues/189#issuecomment-723925332. Some more cases below.

However, maybe `causality="parameter", variability="fixed"` makes more sense? Without having spent a lot of thinking on this issue, I think this at least "looks" much clearer to me with my shallow understanding of `causality` and `variability`. `tunable` and `local` sounds to me very cryptic. With `parameter` and `fixed` I at least have a feeling about what's happening - I know, not a very strong argument ;)

oscillator/README.md
BenjaminRodenberg: Where is this exception?

BenjaminRodenberg: Found it: https://github.com/precice/fmi-runner/blob/b45161052cd6547782c7eb3c8d56f84b49f9f6b8/fmiprecice/runner.py#L188C20-L192

Maybe `can_get_and_set_fmu_state` is just not set propely? To me this looks like an issue with the fmi-runner.

oscillator/fmi/fmu/Oscillator/FMI1CS.xml
LeonardWilleke: Token was taken from FMU templates. This is actually the same token as in #474

oscillator/fmi/fmu/Oscillator/readme.html
LeonardWilleke: As in #474: No, I would pick one and delete the other.

oscillator/fmi/fmu/Oscillator/FMI2.xml
LeonardWilleke: Can the FMI runner access `mass.u` from outside when you set `causality=local`? I remember that only some variable types are accessible and others are shielded from being read or set during the simulation. We want to read and store `mass.u` to plot it later. This is possible with `causality=parameter`

`variability=fixed` should allow you to set `mass.u` at the beginning of the simulation but then don't change it anymore from outside during simulation, which sounds right.

Side note: You are proposing changes in `FMI2.xml` but I think the tutorial uses `FMI3.xml` because we build the FMU with `-DFMI_VERSION=3`. In the end, any changes should be transferred into all the XML files. Unfortunately this can be very confusing when working with the case files.

oscillator/README.md
MakisH, line 1: General issue: after patching the `FMI3.xml` to start the simulation (see separate discussion), I eventually get this error:

```
Exception: Implicit coupling not possible because FMU model can't reset state. Please implement getFMUstate() and setFMUstate() in FMU and set the according flag in ModelDescription.xml. Alternatively, choose an explicit coupling scheme.
```

I am using the fmi-runner commit b8b6e66cdca78c2336136ae7c3bc816e46f99602 (Mar 11).

oscillator/fmi/fmu/Oscillator/FMI2.xml
MakisH: Changing the `mass.a` to `causality="local"` in the `FMI3.xml` resolves this error. But I guess we need to port this change to all variables (as well as all the FMI standards).

@BenjaminRodenberg Note that you need to rebuild the `.fmu` after changing these.

oscillator/README.md
uekerman: I get the same, see above. Will do some more debugging.

oscillator/fmi/fmu/Oscillator/FMI2.xml
BenjaminRodenberg: > Side note: You are proposing changes in `FMI2.xml` but I think the tutorial uses `FMI3.xml` because we build the FMU with `-DFMI_VERSION=3`. In the end, any changes should be transferred into all the XML files. Unfortunately this can be very confusing when working with the case files.

This is mainly a policy question: Do we really want to support FMI2 and FMI3? FMI3 is still quite recent and FMI2 is still broadly used (as far as I know). So supporting both standards sounds reasonable. However, this also will result in a higher maintainment cost.

oscillator/README.md
LeonardWilleke: Maybe it is connected to this issue in FMPy: https://github.com/CATIA-Systems/FMPy/issues/560

uekerman: Quick test seems promising :+1: 
I'll have a closer look.

uekerman: indeed solved the problem

oscillator/README.md
```diff
@@ -17,19 +17,16 @@ This tutorial solves a simple mass-spring oscillator with two masses and three s
...
+- *FMI*: A solver using the [preCICE-FMI runner](https://github.com/precice/fmi-runner) (requires at least v0.2). The Runner executes the FMU model `Oscillator.fmu` for computation. The compiled FMU model for Linux is part of this repository. For other systems, please recompile the model from the provided [C-files](https://github.com/precice/tutorials/tree/master/oscillator/fmi/fmu). If you want to change the model parameters or the initial conditions of the simulation, have a look inside the setting files for [MassLeft](https://github.com/precice/tutorials/tree/master/oscillator/fmi/MassLeft) and [MassRight](https://github.com/precice/tutorials/tree/master/oscillator/fmi/MassRight). For more information, please refer to [2].
```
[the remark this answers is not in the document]
uekerman: done

oscillator/fmi/fmu/Oscillator/FMI2.xml
LeonardWilleke: I agree to remove FMI1.

MakisH: Do we want to remove it from the flow-around-a-controlled-cylinder as well?

oscillator/fmi/fmu/include/FMI1.h
MakisH, line 1: Do we also want to remove:

- `include/FMI1.h`
- `include/fmi1Function.h`
- `src/FMI1.c`
- (adjust the `CMakeLists.txt` accordingly)

@LeonardWilleke @uekerman Do we have a good reason to maintain both FMI2 and FMI3? I have no idea about the state of things in this community.

oscillator/fmi/fmu/Oscillator/FMI2.xml
```diff
@@ -94,7 +94,7 @@
     <ScalarVariable
       name="mass.a"
       valueReference="4"
-      causality="parameter"
+      causality="local"
```
MakisH, line 97: What about the other variables in here that have the combination `causality="parameter", variability="tunable"`? Are they unused?

MakisH (commented):
I can now also run and plot the results. Almost ready, I just thought we could remove a few potentially unused files and options (see comments).

MakisH:
Note that if I execute the left with FMI and right with Python, visualizing with

```
python3 plot-trajectory.py fmi/output/trajectory-Mass-Left.csv TRAJECTORY
```

gives me this plot:

trajectory-mass-left-fmi-right-python

If I execute the left with Python and right with FMI, visualizing with

```
python3 plot-trajectory.py python/output/trajectory-Mass-Left.csv TRAJECTORY
```

gives me this plot:

trajectory-mass-left-python-right-fmi

fmi-fmi gives me the same as python-python:

fmi-fmi

I have not looked into the details, I am just raising this in case it looks strange.

oscillator/fmi/fmu/Oscillator/FMI2.xml
uekerman: I am not sure if the combination per-se is forbidden. All parameters are used.

oscillator/fmi/fmu/Oscillator/FMI2.xml
uekerman: Yes. I'll do it.

oscillator/fmi/fmu/include/FMI1.h
uekerman: I removed more FMI1 files and code. I did not touch the CMakeLists.txt. I have the feeling that this is a standard FMU variant. Finding out what exactly one could remove looks cumbersome.

Good reasons to keep  the FMI2 variant: The runner can handle FMI2 and this way we have a testcase.

uekerman:
The different trajectories look like numerical features to me. Depends on which timestepping methods are used and combined. Playing around with WI, we should be able to remove the instability, but that's beyond this PR.

MakisH (approved):

oscillator/fmi/fmu/include/FMI1.h
MakisH: Thank you for already updating the flow-around-controlled-moving-cylinder as well.
I tried rebuilding the FMU for both cases, no issues. :+1:

PR #468, opened 2024-02-22
Add formatting to pre commit

fsimonis:
This PR adds clang-format and autopep8 pre-commit hooks and applies the formatting.

Also, the markdownlint hook doesn't work on Ubuntu 20.04 as @uekerman found out. I reverted to v0.30 which is known to work.

Next step would be to replace individual workflows with a pre-commit hook workflow.


MakisH (approved):
Makes sense and is helpful, thanks!
We are using a different `.clang-format` in the OpenFOAM adapter (which we could also use here for the OpenFOAM files), but let's rather keep one central file, same as preCICE. In any case, formatting OpenFOAM source files is an upstream issue.

Let's merge this before we start getting conflicts in other PRs.

PR #472, opened 2024-02-29
Add a `fluid-fake` for easier debugging

BenjaminRodenberg:
See #176 for motivation. Here is a small drawing:

Screenshot from 2024-02-29 11-42-37


MakisH:
I am struggling to decide if we should pursue this direction or not.

On the one hand:

- it is clear that it was useful for you
- this is what we were sometimes suggesting users to try out first
- it does not seem to harm merging it
- it sounds like the right tutorial to add such a fake solver, since most of our users probably start from this tutorial. But let's not add fake solvers to more tutorials.

I also understand that the motivation was mainly to have uni-directional coupling instead of bi-directional, to not have to feed back data while developing the FEniCS adapter.

On the other hand, this is why we developed the [ASTE replay mode](https://precice.org/tooling-aste.html#replay-mode). Instead of creating fake solvers, we could instead (in the long-term) generate replay data and store them next to the `reference-results` we now have.

Clearly, the situation is _"I developed a quick solution before the long-term solution was available, do we now also want to make the quick solution available, or not?"_. Given that ASTE is currently not super easy to install, and this is probably more intuitive for users, I would say "why not". But let's add a note in the README that ASTE replay mode would be the more powerful approach.

BenjaminRodenberg:
Since I stumbled across this PR and I think we should merge it rather sooner than later:

> But let's add a note in the README that ASTE replay mode would be the more powerful approach.

I can link to https://precice.org/tooling-aste.html.

However, I do not fully agree that both are addressing the same thing: The replay mode has a different use case.

With the fake fluid you replace the real fluid with a simplified fake version. This eliminates possible sources of error (other solver and other adapter) and allows us to focus on only one party in the overall simulation (this solver and this adapter). I think this is useful, if you do not want to bother with the other solver or you do not trust or cannot assess the output. This is particularly useful for development when the other side is not yet available.

The replay mode only bypasses the computation of one solver and simplifies the feedback loop somehow. This reduces computation time and provides an environment that is close to reality. But you should be quite sure that the original setup that you used for creating the recording is correct. This is useful for detecting regressions.

uekerman:
> The replay mode only bypasses the computation of one solver and simplifies the feedback loop somehow. This reduces computation time and provides an environment that is close to reality. But you should be quite sure that the original setup that you used for creating the recording is correct. This is useful for detecting regressions.

With ASTE, you could also write artificial data. Then, you have exactly the same functionality as the `fake-fluid` here.

perpendicular-flap/README.md
```diff
@@ -27,6 +27,8 @@ Fluid participant:
 
 * SU2. As opposed to the other two fluid codes, SU2 is in particular specialized for compressible flow. Therefore the default simulation parameters haven been adjusted in order to pull the setup into the compressible flow regime. For more information, have a look at the [SU2 adapter documentation](https://www.precice.org/adapter-su2-overview.html).
 
+* Fake. A simple python script that acts as a fake solver and provides a time-dependent force. This solver can be used for debugging of the solid participant and its adapter. Note that [ASTE's replay mode](https://precice.org/tooling-aste.html#replay-mode) has a similar use case.
```
uekerman, line 30:
```suggestion
* Fake. A simple python script that acts as a fake solver and provides a time-dependent force. This solver can be used for debugging of the solid participant and its adapter. Is also works with implicit coupling, thus no changes to the preCICE configuration are necessary. Note that [ASTE's replay mode](https://precice.org/tooling-aste.html#replay-mode) has a similar use case.
```

perpendicular-flap/fluid-fake/fake.py
```diff
@@ -0,0 +1,53 @@
...
+configuration_file_name = "../precice-config.xml"
+participant_name = "Fluid"
+mesh_name = "Fluid-Mesh"
+write_data_name = 'Force'
+
+solver_process_index = 0
+solver_process_size = 1
+
+interface = precice.Participant(participant_name, configuration_file_name, solver_process_index, solver_process_size)
+dimensions = interface.get_mesh_dimensions(mesh_name)
```
uekerman, line 17: You could add an assertion here that dimensions has to be 2.

uekerman (approved):
Let's add this.

Could be good to also add a `requirements.txt` and a venv here. Similar to the Nutils cases.

perpendicular-flap/fluid-fake/fake.py
BenjaminRodenberg: Done 2d42b538b8e40a7a420eef8c12fc64a34c73a68e

BenjaminRodenberg:
Ready to merge. @MakisH I'll wait for your ok or you can also push the button, if you want. Don't want to interfere with your release with something that is not critical.

perpendicular-flap/README.md
```diff
@@ -27,6 +27,8 @@ Fluid participant:
 
 * SU2. As opposed to the other two fluid codes, SU2 is in particular specialized for compressible flow. Therefore the default simulation parameters haven been adjusted in order to pull the setup into the compressible flow regime. For more information, have a look at the [SU2 adapter documentation](https://www.precice.org/adapter-su2-overview.html).
 
+* Fake. A simple python script that acts as a fake solver and provides a time-dependent force. This solver can be used for debugging of the solid participant and its adapter. It also works with implicit coupling, thus no changes to the preCICE configuration are necessary. Note that [ASTE's replay mode](https://precice.org/tooling-aste.html#replay-mode) has a similar use case.
```
MakisH, line 30: Trying to make it even more prominent that this is not meant for actual simulations and what ASTE is for. This is actually a nice place to advertise both tools as development/debugging approaches, as many users look here early enough.

```suggestion
* Fake. A simple python script that acts as a fake solver and provides an arbitrary, time-dependent force. This solver can be used for debugging of the solid participant and its adapter. It also technically works with implicit coupling, thus no changes to the preCICE configuration are necessary. Note that [ASTE's replay mode](https://precice.org/tooling-aste.html#replay-mode) has a similar use case and could also feed artificial or previously recorded real data, replacing an actual solver.
```

MakisH (approved):
I think that the current state is nice and sustainable. There is a clear description in the README, there is a clear hint to ASTE (while it is also implied that ASTE would be more complicated to use), and the case follows the guidelines.

I also tested the following combinations, and they completed, resulting into a reasonably moving flap:

- fake-deal.II
- fake-OpenFOAM
- fake-Nutils

Let's merge this already, there is no reason to wait, and I can add the watchpoint next to the rest.

perpendicular-flap/README.md
MakisH: I guess you don't disagree with these changes, so I will go ahead and merge everything to move on. Feel free to complain and I will update.

PR #279, opened 2022-06-29
Adds breaking dam test cases

KyleDavisSA:
This PR adds a breaking dam 2D and breaking dam 3D test case for multiphase flow of water/air mixture, where a body of water strikes a solid, deformable wall.

PR #473, opened 2024-03-04
Add 3D breaking dam case

uekerman:
Leftover from #279 
Needs major clean-up and porting to preCICE v3.

Originally contributed by @KyleDavisSA


MakisH:
As soon as #279 works and is merged, porting this one should be easy and quick.

MakisH:
We now have a 2D case in the tutorials: https://precice.org/tutorials-breaking-dam-2d.html

This seems to be already ported to preCICE v3. We still need to run and review it. However, it is not a priority (but would be nice to have).

In principle, this would be a nice first PR to review for a new reviewer.

PR #475, opened 2024-03-10
Fix checkpointing and subcycling for `solid.py`

### LINKED ISSUE #50: Fix checkpointinf for FEniCS FSI tutorials
Only `u_n` is considered by the checkpointint. But we should also properly checkpoint `v_n` and `a_n`. There is already a prototype implementation on [this branch](https://github.com/precice/tutorials/tree/convergenceStudy-WR) (don't merge this branch, it has a different purpose!)

The most important part is

https://github.com/precice/tutorials/blob/b3b498562255b8a4fb60c785cce4c71a9bd36abc/FSI/cylinderFlap/OpenFOAM-FEniCS/Solid/cyl-flap.py#L238-L243

BenjaminRodenberg:
Fixes checkpointing by considering u, v, and a. Closes #50.

Note: This fix requires https://github.com/precice/fenics-adapter/pull/170 which is currently not merged and also not released in v2.0.0 of fenics-adapter. Doing a 2.1.0 release of the adapter for this feature would not be a big problem from my perspective. Adding this fixed version of the tutorial to the distribution release gives us another example using subcycling + time interpolation.


MakisH:
I see several changes here. Which one is the essence of the fix? Why was it broken before?

> Note: This fix requires https://github.com/precice/fenics-adapter/pull/170 which is currently not merged and also not released in v2.0.0 of fenics-adapter. Doing a 2.1.0 release of the adapter for this feature would not be a big problem from my perspective.

In the distribution, we need that all tutorials work with the respective versions of components released in the same distribution. So yes, please do that, if needed.

perpendicular-flap/solid-fenics/solid.py
```diff
@@ -196,13 +197,13 @@ def avg(x_old, x_new, alpha):
...
     if precice.requires_writing_checkpoint():  # write checkpoint
-        precice.store_checkpoint(u_n, t, n)
+        precice.store_checkpoint((u_n, v_n, a_n), t, n)
 
     precice_dt = precice.get_max_time_step_size()
     dt = Constant(np.min([precice_dt, fenics_dt]))
 
     # read data from preCICE and get a new coupling expression
-    read_data = precice.read_data(dt)
+    read_data = precice.read_data((1-float(alpha_f)) * dt)
```
MakisH, line 206: I guess this change is related to the timestepping method used here. A comment on what these parameters `alpha_f` and `alpha_m` mean could be helpful for peasants like me, especially since the `1-alpha_f` seems cryptic.

perpendicular-flap/solid-fenics/solid.py
```diff
@@ -80,15 +79,17 @@ def neumann_boundary(x, on_boundary):
...
 # clamp the beam at the bottom
 bc = DirichletBC(V, Constant((0, 0)), fixed_boundary)
 
 # alpha method parameters
-alpha_m = Constant(0)
-alpha_f = Constant(0)
+alpha_m = Constant(0.2)
+alpha_f = Constant(0.4)
+# alpha_m = Constant(0)
+# alpha_f = Constant(0)
```
BenjaminRodenberg, line 92: `alpha_m` and `alpha_f` are the parameters from the generalized alpha method (reference given below). I can add a comment in the `read_data` line why we sample the data where we sample it.

perpendicular-flap/solid-fenics/solid.py
BenjaminRodenberg, line 206:
```suggestion
    read_data = precice.read_data((1-float(alpha_f)) * dt)  # sample force F at $F(t_{n+1-\alpha_f})$ (see generalized alpha paper)
```

perpendicular-flap/solid-fenics/solid.py
```diff
@@ -80,15 +79,17 @@ def neumann_boundary(x, on_boundary):
 
 precice_dt = precice.get_max_time_step_size()
 fenics_dt = precice_dt  # if fenics_dt == precice_dt, no subcycling is applied
-# fenics_dt = 0.02  # if fenics_dt < precice_dt, subcycling is applied
+# fenics_dt = precice_dt / 5  # if fenics_dt < precice_dt, subcycling is applied
```
IshaanDesai, line 82: A small comment on why we chose a factor of 5 would be helpful here.

perpendicular-flap/solid-fenics/solid.py
```diff
@@ -196,13 +197,13 @@ def avg(x_old, x_new, alpha):
...
     if precice.requires_writing_checkpoint():  # write checkpoint
-        precice.store_checkpoint(u_n, t, n)
+        precice.store_checkpoint((u_n, v_n, a_n), t, n)
 
     precice_dt = precice.get_max_time_step_size()
     dt = Constant(np.min([precice_dt, fenics_dt]))
 
     # read data from preCICE and get a new coupling expression
-    read_data = precice.read_data(dt)
+    read_data = precice.read_data((1-float(alpha_f)) * dt)  # sample force F at $F(t_{n+1-\alpha_f})$ (see generalized alpha paper)
```
IshaanDesai, line 206: Link to the paper?

IshaanDesai (approved):
Looks good :+1: I have only two minor comments.

perpendicular-flap/solid-fenics/solid.py
BenjaminRodenberg: `5` is 100% arbitrary. Maybe this way to make it clearer that this is a configurable parameter?
```suggestion
# n_substeps = 5  # number of substeps per window
# fenics_dt = precice_dt / n_substeps  # if fenics_dt < precice_dt, subcycling is applied
```

PR #459, opened 2024-02-04
Follow naming convention in partitioned heat conduction - FEniCS

BenjaminRodenberg:
Fixed incorrect order for OpenFOAM

Updated structure to the one provided [here](https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial) for FEniCS.

Other cases are still missing.

PR #477, opened 2024-03-10
Add a .gitignore and clean-up

MakisH:
- Adds a central `.gitignore`, taking into account all solvers we currently support, and all rules we have already defined in `cleaning-tools.sh`.
- Adds individual `.gitignore` files in special cases, mainly where we download files from external repositories: `aste-turbine`, `heat-exchanger`.
- Restores some `Make/` directory unintentionally deleted in https://github.com/precice/tutorials/pull/459 and fixes the corresponding `.gitignore`. (cc @BenjaminRodenberg)

Also adds a comments on instructions to do a deep clean, as at the moment all externally downloaded files, python virtual environments, and C++ code builds are not deleted by the cleaning scripts (and one generally should not need to delete them). I already tried extending the cleaning scripts for this purpose, but it was easy to make them unnecessarily more complex.

Also, general cleanup:
- A stray `reference-data` is renamed to `reference-results` (cc @valentin-seitz)
- Improves the readability of the cleaning scripts, so that at least when everything is clean, one can easily see if something is wrong.
- Adds exceptions for the `solver-*` directories, to avoid the expected `skipping` messages.

@IshaanDesai @Fujikawas how do you like this change? Anything I have missed?

I did not really try running anything DUNE-related, but I guess it is covered.

closes https://github.com/precice/tutorials/issues/87


Fujikawas:
>but I guess it is covered.

Thank you doing this, I think all the folders and files to be removed relating to `dunefem` and `dumux` has been covered.

IshaanDesai (approved):
Looks good :+1:

Fujikawas (approved):

PR #480, opened 2024-03-14
Add new tutorial: resonant circuit tutorial

BenjaminRodenberg:
See https://github.com/precice/matlab-bindings/issues/29


MichaelWiesheu:
Meine Änderungen in Solver_I.m
Im wesentlichen wurde
dt = min(solverdt, precicedt);
und das Speichern und Laden von "t" hinzugefügt.


clear; close all; clc;

% Initialize and configure preCICE
interface = precice.Participant("ParticipantI", "precice-config.xml", 0, 1);

% Geometry IDs. As it is a 0-D simulation, only one vertex is necessary.
meshName ="MeshI";
vertex_ID = interface.setMeshVertex(meshName, [0 0]);

% Data IDs
dataNameI = "I";
dataNameU = "U";

% Simulation parameters and initial condition
C = 2;                      % Capacitance
L = 1;                      % Inductance
t0 = 0;                     % Initial simulation time
t_max = 10;                 % End simulation time
Io = 1;                     % Initial current
phi = 0;                    % Phase of the signal

w0 = 1/sqrt(L*C);           % Resonant frequency
I0 = Io*cos(phi);           % Initial condition for I
U0 = -w0*L*Io*sin(phi);     % Initial condition for U

f_I = @(t, I, U) U/L;       % Time derivative of I

% Initialize simulation
I = I0;                     % Vector of I through time
U = U0;                     % Vector of U through time
t_vec = t0;                 % Vector of time
if interface.requiresInitialData()
    interface.writeData(meshName, DataNameI, vertex_ID, I0);
end
interface.initialize();
solverdt = 0.01;
precicedt = interface.getMaxTimeStepSize();
dt = min(solverdt, precicedt);

% Start simulation
t = t0 + dt;
while interface.isCouplingOngoing()

    % Record checkpoint if necessary
    if interface.requiresWritingCheckpoint()
        I0_checkpoint = I0;
        U0_checkpoint = U0;
        t_checkpoint = t;
    end
    precicedt = interface.getMaxTimeStepSize();
    dt = min(solverdt, precicedt);
    disp(['t: ' num2str(t) 9 ' solverdt: ' num2str(solverdt) 9 'precicedt: ' num2str(precicedt) 9 ' dt: ' num2str(dt)]);

    % Make Simulation Step
    [t_ode, I_ode] = ode45(@(t, y) f_I(t, y, U0), [t0 t], I0);
    I0 = I_ode(end);

    % Exchange data
    interface.writeData(meshName, dataNameI, vertex_ID, I0);
    interface.advance(dt);

    % Recover checkpoint if not converged, else finish time step
    if interface.requiresReadingCheckpoint()
        I0 = I0_checkpoint;
        U0 = U0_checkpoint;
        t = t_checkpoint;
    else
        U0 = interface.readData(meshName, dataNameU, vertex_ID, dt);
        U = [U U0];
        I = [I I0];
        t_vec = [t_vec, t];
        t0 = t;
        t = t0 + dt;
    end
end

% Stop coupling
interface.finalize();

% Analytical solution for comparison
I_an = Io*cos(w0*t_vec+phi);
U_an = -w0*L*Io*sin(w0*t_vec+phi);

% Make and save plot
figure(1)
subplot(2,1,1)
plot(t_vec, I_an)
hold on;
plot(t_vec, U_an)
ylim([-1,1])
legend('I', 'U')
title('Analytical')

subplot(2,1,2)
plot(t_vec, I)
hold on;
plot(t_vec, U)
ylim([-1,1])
title('Numerical')
legend('I', 'U')

save('outputs.mat', 'I', 'U')
saveas(gcf, 'Curves.png')

resonant-circuit/capacitor-matlab/Solver_I.m
```diff
@@ -0,0 +1,98 @@
...
+% Geometry IDs. As it is a 0-D simulation, only one vertex is necessary.
+meshName ="MeshI";
+vertex_ID = interface.setMeshVertex(meshName, [0 0]);
+
+% Data IDs
+dataNameI = "I";
+dataNameU = "U";
+
+% Simulation parameters and initial condition
+C = 2;                      % Capacitance
```
BenjaminRodenberg, line 15:
```suggestion
```
Can we remove this here? I also think this script `Solver_I.m` actually also represents the coil, not the capacitor. So it should be inside `coil-matlab`, no?

Btw: This also makes sense w.r.t black-box coupling: From the perspective of the coil, the capacitance `C` of the capacitor is considered "expert knowledge" that the coil should not have to know.

resonant-circuit/precice-config.xml
```diff
@@ -0,0 +1,50 @@
...
+      direction="write" from="MeshU" to="MeshI" constraint="consistent"/>
+    <mapping:nearest-neighbor
+      direction="read" from="MeshI" to="MeshU" constraint="consistent"/>
+  </participant>
+
+  <m2n:sockets acceptor="ParticipantI" connector="ParticipantU" exchange-directory=".." />
+  
+  <coupling-scheme:serial-implicit> 
+    <participants first="ParticipantI" second="ParticipantU" /> 
+    <max-time-windows value="1000" />
```
BenjaminRodenberg, line 44:
```suggestion
    <max-time value="10" />
```
This makes it easier to do convergence studies and is also more intuitive for me.

resonant-circuit/capacitor-matlab/Solver_I.m
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+end
+
+% Stop coupling
+interface.finalize();
+
+% Analytical solution for comparison
+I_an = Io*cos(w0*t_vec+phi);
+U_an = -w0*L*Io*sin(w0*t_vec+phi);
```
BenjaminRodenberg, line 77: At the very end of the run, we could also estimate the error by comparing to the analytical solution. Something along these lines (in python):

```
I_analytical = lambda t: Io*np.cos(t*w0 + phi)
error = I0 - I_analytical(10)
print(f"error: {error}")
```

resonant-circuit/precice-config.xml
```diff
@@ -0,0 +1,50 @@
...
+  <participant name="ParticipantU">
+    <provide-mesh name="MeshU" />
+    <receive-mesh name="MeshI" from="ParticipantI"/>
+    <write-data name="U"     mesh="MeshU" />
+    <read-data  name="I" mesh="MeshU" />
+    <mapping:nearest-neighbor
+      direction="write" from="MeshU" to="MeshI" constraint="consistent"/>
+    <mapping:nearest-neighbor
+      direction="read" from="MeshI" to="MeshU" constraint="consistent"/>
+  </participant>
```
BenjaminRodenberg, line 38:
```suggestion
  <participant name="Coil">
    <provide-mesh name="Coil-Mesh" />
    <write-data name="Current" mesh="Coil-Mesh" />
    <read-data  name="Voltage" mesh="Coil-Mesh" />
  </participant>

  <participant name="Capacitor">
    <provide-mesh name="Capacitor-Mesh" />
    <receive-mesh name="Coil-Mesh" from="Coil"/>
    <write-data name="Voltage" mesh="Capacitor-Mesh" />
    <read-data  name="Current" mesh="Capacitor-Mesh" />
    <mapping:nearest-neighbor
      direction="write" from="Capacitor-Mesh" to="Coil-Mesh" constraint="consistent"/>
    <mapping:nearest-neighbor
      direction="read" from="Coil-Mesh" to="Capacitor-Mesh" constraint="consistent"/>
  </participant>
```

We only need a mapping on one of the two participants. Additionally, some proposals for renaming.

NiklasVin: If I remove either mapping, so using this for example:

<details>
<summary> preCICE config file</summary>

```XML
<?xml version="1.0"?>



<precice-configuration>



  <data:scalar name="Current"  />

  <data:scalar name="Voltage"  />



  <mesh name="MeshCoil" dimensions="2">

    <use-data name="Current" />

    <use-data name="Voltage" />

  </mesh>



  <mesh name="MeshCapacitor" dimensions="2">

    <use-data name="Voltage" />

    <use-data name="Current" />

  </mesh>



  <participant name="Coil">

    <provide-mesh name="MeshCoil" />

    <write-data name="Current"     mesh="MeshCoil" />

    <read-data  name="Voltage" mesh="MeshCoil" />

  </participant>



  <participant name="Capacitor">

    <provide-mesh name="MeshCapacitor" />

    <receive-mesh name="MeshCoil" from="Coil"/>

    <write-data name="Voltage"     mesh="MeshCapacitor" />

    <read-data  name="Current" mesh="MeshCapacitor" />

    <mapping:nearest-neighbor

      direction="write" from="MeshCapacitor" to="MeshCoil" constraint="consistent"/>

    <mapping:nearest-neighbor

      direction="read" from="MeshCoil" to="MeshCapacitor" constraint="consistent"/>

  </participant>



  <m2n:sockets acceptor="Coil" connector="Capacitor" exchange-directory=".." />

  

  <coupling-scheme:serial-implicit> 

    <participants first="Coil" second="Capacitor" /> 

    <max-time value="10" />

    <time-window-size value="0.01" />

    <max-iterations value="3" />

    <exchange data="Current" mesh="MeshCoil" from="Coil" to="Capacitor"/>

    <exchange data="Voltage" mesh="MeshCapacitor" from="Capacitor" to="Coil"/>

  </coupling-scheme:serial-implicit>     

</precice-configuration>
```
</details>

I get the error

```
(0) 19:56:49 [config::Configuration]:87 in xmlEndTagCallback: ERROR: Participant "Coil" needs to use the mesh "MeshCapacitor" to be able to use it in the coupling scheme. Please either add a provide-mesh or a receive-mesh tag in this participant's configuration, or use a different mesh in the coupling scheme.

```
, while everything works fine with both mappings.

BenjaminRodenberg: The problem here is in the coupling scheme tag: The tag `<exchange data="Voltage" mesh="MeshCapacitor" from="Capacitor" to="Coil"/>` defines a data exchange between `Capacitor` and `Coil` of some data sitting on `MeshCapacitor`. But `Coil` does not know `MeshCapacitor`. If you use `MeshCoil` here things should work.

resonant-circuit/matlab-bindings-path.sh
BenjaminRodenberg, line 1: I would suggest to remove this file. In the matlab bindings we suggest to add the bindings to the `startup.m` file and I think this is a good way to do it and we can assume that our users have done this.

resonant-circuit/coil-matlab/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Import MATLAB-bindings path
+. ../matlab-bindings-path.sh
+
+# Add bindings to MATLAB path
+export MATLABPATH="$BINDINGS"
```
BenjaminRodenberg, line 7:
```suggestion
```
See comment below.

resonant-circuit/capacitor-matlab/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# Import MATLAB-bindings path
+. ../matlab-bindings-path.sh
+
+# Add bindings to MATLAB path
+export MATLABPATH="$BINDINGS"
+
```
BenjaminRodenberg, line 8:
```suggestion
```

BenjaminRodenberg (commented):
Looks good to me. @NiklasVin please also remove the case https://github.com/precice/matlab-bindings/tree/develop/tutorial from the matlab-bindings repo now that we moved it here.

I think it would be nice to get a review from somebody else, if possible. The installation process of the MATLAB bindings is not the best and adding the bindings to the path is not so easy, but let's try to solve this in the repository of the bindings. Not here.

resonant-circuit/README.md
```diff
@@ -0,0 +1,66 @@
...
+The circuit is described by the following system of ODEs:
+
+$U(t) = L \frac{\text{d}I}{\text{d}t}$
+
+$I(t) = -C \frac{\text{d}U}{\text{d}t}$
+
+where $I$ is the current and $U$ the voltage of the cirucit.
+
+Each of these equations is going to be solved by a different MATLAB solver. Note that as only one scalar is solved per equation, this is a 0+1 dimensional problem.
+
```
MakisH, line 23: You need a new section here to show the config visualization. This is a recent change, see #514. Merge from `develop`, then you can use the tool `toosl/visualize-configs.sh`.

resonant-circuit/capacitor-matlab/run.sh
```diff
@@ -0,0 +1,12 @@
...
+# Run MATLAB code without GUI
+LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6 matlab -nodisplay -nosplash -nodesktop -r "capacitor;exit;"
```
MakisH, line 10: This seems to be very system-specific (even though probably most systems we target should be fine). Is there any more portable way to describe the same paths?

Is `.6` the same at least across all Ubuntu versions we support?

resonant-circuit/coil-matlab/clean.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e -u
+
+find . -not -name "clean.sh" -not -name "run.sh" -not -name "Solver_Coil.m" -delete
```
MakisH, line 4: Do we really need to delete all files except these three? Are the result files so diverse?

You could also move that into a function in `tools/cleaning-tools.sh`.

resonant-circuit/coil-matlab/run.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Run MATLAB code without GUI
+LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6 matlab -nodisplay -nosplash -nodesktop -r "coil;exit;"
+
+stty echo
```
MakisH, line 12: Why is this needed?
At least a comment next to it would help maintain it.

MakisH (commented):
Structure-wise, this looks ready.
I have not tried to run it, but I would be interested, since we don't currently have any Matlab-based tutorial.

If everything is camera-ready today, we could squeeze it into this release, if you feel confident about it. But I would not wait for it, because we are already quite late.

resonant-circuit/precice-config.xml
fsimonis: Using time-windows is more robust though :wink:

resonant-circuit/capacitor-matlab/run.sh
BenjaminRodenberg: I opened an issue to understand this a bit better. Maybe we can just remove it (assuming our users fulfil certain requirements for the gcc / MATLAB version being used). See https://github.com/precice/matlab-bindings/issues/53.

BenjaminRodenberg:
```suggestion
LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6 matlab -nodisplay -r "capacitor;exit;"
```
Don't want to discuss `LD_PRELOAD` here, but I think we can remove some additional arguments. Only `-nodisplay` is needed.

resonant-circuit/coil-matlab/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Import MATLAB-bindings path
+. ../matlab-bindings-path.sh
+
+# Add bindings to MATLAB path
+export MATLABPATH="$BINDINGS"
+
+# Run MATLAB code without GUI
+LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6 matlab -nodisplay -nosplash -nodesktop -r "coil;exit;"
```
BenjaminRodenberg:
```suggestion
LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6 matlab -nodisplay -r "coil;exit;"
```
See above.

BenjaminRodenberg (commented):
I found another minor simplification. @NiklasVin if you think this is reasonable, please commit these suggestions to the branch.

resonant-circuit/capacitor-matlab/run.sh
BenjaminRodenberg: I updated the documentation in the MATLAB bindings (see https://github.com/precice/matlab-bindings/commit/20f8653b6f4f20c890e8fd8c3a981f6571ed8884). If @NiklasVin can also fix the problem simply by updating his MATLAB installation, I would recommend to remove `LD_PRELOAD` from here and assume our users have a recent MATLAB installation.

@NiklasVin can you try this out and then take care of the rest?

resonant-circuit/precice-config.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<precice-configuration>
+  <data:scalar name="Current" />
+  <data:scalar name="Voltage" />
+
+  <mesh name="MeshCoil" dimensions="2">
```
BenjaminRodenberg:
```suggestion
  <mesh name="Coil-Mesh" dimensions="2">
```
Use, for example, https://github.com/precice/tutorials/blob/develop/perpendicular-flap/precice-config.xml as reference. Same for `Capacitor-Mesh`.

resonant-circuit/capacitor-matlab/run.sh
NiklasVin: >  If @NiklasVin can also fix the problem simply by updating his MATLAB installation, I would recommend to remove `LD_PRELOAD` from here and assume our users have a recent MATLAB installation.
> 

Updating MATLAB to the newest version didn't help, but I'd suggest removing it as it works for different MATLAB versions on your system. This issue is known and a workaround exists, so I would at most add a comment in the README file referring to the [troubleshooting](https://github.com/precice/matlab-bindings/blob/develop/README.md#troubleshooting) section

BenjaminRodenberg: Let's keep the issue https://github.com/precice/matlab-bindings/issues/53 open and try to understand why it's working for me and not for you over there (can you please also comment in the issue in case you did not do this?).

For this PR: Let's go for the approach that works for me. If it explodes without `LD_PRELOAD` it is well documented how to fix it.

Please don't put a link into this README. We also would have to maintain it etc. We already have this documented in the matlab bindings and I think this is sufficient. If this tutorial explodes for a user it is easy to google the error and navigate to the README of the matlab bindings.

resonant-circuit/capacitor-matlab/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Run MATLAB code without GUI
+matlab -nodisplay -r "capacitor;exit;"
+
```
BenjaminRodenberg:
```suggestion
```
No blank line needed here.

resonant-circuit/coil-matlab/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Run MATLAB code without GUI
+matlab -nodisplay -r "coil;exit;"
+
```
BenjaminRodenberg:
```suggestion
```
No blank line needed here.

BenjaminRodenberg (commented):
I think we are converging here! Can you please go once more through all the comments, briefly write an answer (if necessary) and resolve the conversations. @MakisH Do you want to do another quick check, approve, and merge (I cannot approve, because I created this PR originally).

resonant-circuit/matlab-bindings-path.sh
NiklasVin: I removed it 👍

resonant-circuit/README.md
NiklasVin: Thanks for mentioning it. I merged the branches and updated the README accordingly

resonant-circuit/coil-matlab/clean.sh
NiklasVin: I wasn't aware that in `tools/cleaning-tools.sh` there existed a function removing all files preCICE creates. I added the function `clean_matlab` to `tools/cleaning-tools.sh` and use that instead

resonant-circuit/coil-matlab/run.sh
NiklasVin: I forgot to remove that 😅. While testing, I had issues with the terminal after executing the program, but now it's fine

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Resonant Circuit
+keywords: MATLAB
+summary: We simulate a two-element LC circuit (one inductor and one capacitor).
+---
+
+
+## Setup
+
+The purpose of this tutorial is to illustrate the usage of preCICE to couple MATLAB code. Two different MATLAB solvers will be coupled to simulate a two-element LC circuit. This type of circuit consists on a very simple system with one inductor and one capacitor:
```
MakisH, line 10: Just a general picky comment (no need to fix now, but keep it in mind for future PRs): we want to keep language a bit more informal and engaging, see [content guidelines](https://precice.org/docs-meta-content-guidelines.html).

Don't be afraid to write exactly as you would imagine explaining something to a colleague over coffee. :smile:

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Configuration
+
+preCICE configuration (image generated using the [precice-config-visualizer](https://precice.org/tooling-config-visualization.html)):
+
+![preCICE configuration visualization](images/tutorials-resonant-circuit-precice-config.png)
+
+## Available solvers
+
+* <em>MATLAB</em> A solver using the [MATLAB bindings](https://github.com/precice/matlab-bindings).
```
MakisH: For consistency and to avoid injecting HTML into Markdown:

```suggestion
* MATLAB. A solver using the [MATLAB bindings](https://precice.org/installation-bindings-matlab.html).
```

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+ Before running this tutorial, follow the [instructions](https://github.com/precice/matlab-bindings?tab=readme-ov-file#matlab-bindings) to correctly install the MATLAB bindings.
```
MakisH:
```suggestion
 Before running this tutorial, follow the [instructions](https://precice.org/installation-bindings-matlab.html) to correctly install the MATLAB bindings.
```

If that page is not up-to-date, then we need to update it :smile:

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```MATLAB
+capacitor
+```
+
+The preCICE configuration file is available as `precice-config.xml`, and it is called directly in the solvers.
+
+#### Running from terminal
+
+If you prefer to not open the MATLAB GUIs, one can alternatively use two shells instead.
```
MakisH:
```suggestion
If you prefer to not open the MATLAB GUIs, you can alternatively use two shells instead.
```

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+For that, modify the path in the file `matlab-bindings-path.sh` found in the base directory of this tutorial to the path to your MATLAB bindings.
+
+By doing that, you can now open two shells and switch into the directories `capacitor-matlab` and `coil-matlab` and execute the `run.sh` scripts.
+
+## Visualization
```
MakisH: For consistency:

```suggestion
## Post-processing
```

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The solver for the current also records the current and voltage through time and at the end of the simulation saves a plot with the obtained curves, as well as the analytical solution.
+
+After successfully running the coupling, one can find the curves in the folder `capacitor-matlab` as `Curves.png`.
```
MakisH, line 68: You could also include this image here (under `resonant-circuit/images/`). In that case, name it `tutorials-resonant-circuit-curves.png`.

resonant-circuit/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## References
+
+[1] By First Harmonic - Own work, CC BY-SA 3.0, https://commons.wikimedia.org/w/index.php?curid=21991221
```
MakisH: Title is missing

MakisH (approved):
Nice! :rocket: 

I still have not tested it, but that would take much longer, as I first have to finish other tasks. Feel free to merge after fixing a few minor issues in the README.

resonant-circuit/README.md
BenjaminRodenberg: True. We need to migrate some content from the `README.md` to the website. But for now, let's the link as suggested above. Currently, the Website provides minimal information and points to the `README.md` for further information. This we can change in the future.

resonant-circuit/README.md
BenjaminRodenberg:
```suggestion
[1] Schematic of a simple parallel LC circuit by First Harmonic - Own work, CC BY-SA 3.0, https://commons.wikimedia.org/w/index.php?curid=21991221
```

resonant-circuit/README.md
MakisH: And how will we remember to do that? :smile: 
I would suggest to already use the link to the website and just do a copy-paste of the README now, if possible. Or just open an issue.

resonant-circuit/README.md
NiklasVin: Thanks for the tip. I'll keep that in mind :)

changelog-entries/480.md
```diff
@@ -0,0 +1 @@
+- Added new [resonant-circuit tutorial](https://precice.org/tutorials-resonant-circuit.html) with MATLAB (migrated from [matlab-bindings repository](https://github.com/precice/matlab-bindings and updated to follow tutorials structure)).
```
BenjaminRodenberg, line 1: @MakisH the CI pipeline complains here. I think we need to ignore certain rules that we check for the `README.md` files in these kind of files. Or should we just ignore the complete folder?

MakisH: I think I broke that exception recently. Ignore it, and I will fix it in a separate PR.

PR #483, opened 2024-03-19
Add hook to ensure we link to https://precice.org

fsimonis:
This PR adds a hook that converts `http://precice.org` to `https://precice.org`.


MakisH (approved):
Nice idea, nice implementation!

PR #485, opened 2024-03-20
Document script plot-all-displacements

MakisH:
Very use-case-specific, outdated, and trivial to reproduce from plot-displacement.sh

### DIFF
--- a/perpendicular-flap/README.md
+++ b/perpendicular-flap/README.md
@@ -77,6 +77,8 @@ You should get results similar to this one:
 
 ![All flap watchpoints](images/tutorials-perpendicular-flap-displacement-all-watchpoints.png)
 
+This plot has been generated with the `plot-all-displacements.sh` script, which requires executing and pre-processing the individual combinations manually.
+
 Reasons for the differences:
 
 * The CalculiX adapter only supports linear finite elements (deal.II uses 4th order, FEniCS 2nd order).
--- a/perpendicular-flap/README.md
+++ b/perpendicular-flap/README.md
@@ -77,8 +77,6 @@ You should get results similar to this one:
 
 ![All flap watchpoints](images/tutorials-perpendicular-flap-displacement-all-watchpoints.png)
 
-This plot has been generated with the `plot-all-displacements.sh` script, which requires executing and pre-processing the individual combinations manually.
-
 Reasons for the differences:
 
 * The CalculiX adapter only supports linear finite elements (deal.II uses 4th order, FEniCS 2nd order).
--- a/perpendicular-flap/plot-all-displacements.sh
+++ b/perpendicular-flap/plot-all-displacements.sh
@@ -29,16 +29,19 @@ gnuplot -p << EOF
 			 "watchpoints/openfoam-dune.log" using 1:4 with lines title "OpenFOAM-DUNE", \
 			 "watchpoints/openfoam-openfoam.log" using 1:4 with lines title "OpenFOAM-OpenFOAM", \
 			 "watchpoints/openfoam-solids4foam.log" using 1:4 with lines title "OpenFOAM-solids4Foam", \
+			 "watchpoints/openfoam-nutils.log" using 1:4 with lines title "OpenFOAM-Nutils", \
 	     "watchpoints/su2-calculix.log" using 1:4 with lines title "SU2-CalculiX", \
 	     "watchpoints/su2-dealii.log" using 1:4 with lines title "SU2-deal.II", \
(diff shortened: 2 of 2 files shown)


BenjaminRodenberg:
The purpose of this script is to quickly be able to reproduce this plot:

https://github.com/precice/tutorials/blob/master/perpendicular-flap/images/tutorials-perpendicular-flap-displacement-all-watchpoints.png

As far as I remember this is to some degree a regression test we do more or less regularly. We should at least somehow document how this plot is created.

BenjaminRodenberg (changes requested):
I think we should not delete this file and keep everything as it is for the distribution release:

1. Comparing runs `{SU2, OpenFOAM, nutils} X {FEniCS, CalculiX, SU2}` is valuable. E.g. #250 
2. Running all 9 cases and storing them under unique names requires manual effort. I don't think we can easily improve this situation. Giving a name to the watchpoint file that includes the names of both participants (i.e. `openfoam-calculix.log`) breaks the black-box. The participant that is creating the watchpoint file does not know which participant sits on the other side.
3. `plot-displacements.sh` only allows to plot a single curve, not multiple. Extending this to support plotting of several is not trivial in this setting.

Possible todo: Documentation how to do the runs, store the results, and create the plot (manual effort mentioned under 2) could certainly be improved, but this is not really something critical for the distribution release.

MakisH:
>     1. Comparing runs `{SU2, OpenFOAM, nutils} X {FEniCS, CalculiX, SU2}` is valuable. E.g. [Change C3D8 elements to C3D8I elements in perpendicular-flap solid-calculix #250](https://github.com/precice/tutorials/pull/250)

The comparison is valuable, the script as-it-is is mostly confusing. I restored the script and added documentation.

>     2. Running all 9 cases and storing them under unique names requires manual effort. I don't think we can easily improve this situation. Giving a name to the watchpoint file that includes the names of both participants (i.e. `openfoam-calculix.log`) breaks the black-box. The participant that is creating the watchpoint file does not know which participant sits on the other side.

Automation could help, but that would be an overkill even for the system tests, as we need a system with all the solvers installed. We have that already in the VM, but even that would probably take hours to run. Reminder: we now have a cross-product of 3x7 (=21) solvers, with at least Nutils taking quite some time to run. Such automation would need to take fault tolerance into account.

>     3. `plot-displacements.sh` only allows to plot a single curve, not multiple. Extending this to support plotting of several is not trivial in this setting.

It is. Compare:

```diff
gnuplot -p << EOF                                                               
	set grid                                                                        
	set title 'x-displacement of the flap tip'                                        
	set xlabel 'time [s]'                                                           
	set ylabel 'x-displacement [m]'
+	set term pngcairo enhanced size 900,654
+	set output "images/tutorials-perpendicular-flap-displacement-all-watchpoints.png"
- 	plot "$1/precice-Solid-watchpoint-Flap-Tip.log" using 1:4 with lines title "$1"
+       plot "watchpoints/openfoam-calculix.log" using 1:4 with lines title "OpenFOAM-CalculiX", \
	     "watchpoints/openfoam-dealii.log" using 1:4 with lines title "OpenFOAM-deal.II", \
+            ...
EOF
```

The difference is in the hard-coding of the filenames (just more than one arguments). The two `set` should actually also have been used in the simple `plot-displacement.sh`.

> Possible todo: Documentation how to do the runs, store the results, and create the plot (manual effort mentioned under 2) could certainly be improved, but this is not really something critical for the distribution release.

> I would keep the script and add 2 sentences of documentation to the case README.

I added more documentation.

We now essentially need to do the manual run of everything and update the picture. I will try to do that next week in the VM, in a separate PR.

uekerman (approved):

perpendicular-flap/README.md
```diff
@@ -77,6 +77,8 @@ You should get results similar to this one:
 
 ![All flap watchpoints](images/tutorials-perpendicular-flap-displacement-all-watchpoints.png)
 
+This plot has been generated with the `plot-all-displacements.sh` script, which requires executing and pre-processing the individual combinations manually.
+
```
BenjaminRodenberg, line 81: Isn't more or less the same written a few lines above (line 74)

```
There is moreover a script `plot-all-displacements.sh` to plot and compare all possible variants. This script expects all watchpoint logs to be available in a subfolder `watchpoints` in the format `openfoam-dealii.log` or similar. If you want to use this script, you need to copy the files over accordingly.
```

BenjaminRodenberg (approved):
Minor comment. Rest looks ok.

perpendicular-flap/README.md
MakisH: Did not see that, thanks! Removed.

PR #487, opened 2024-03-20
Flow over a heated plate CalculiX: make domains to have same thickness

### LINKED ISSUE #455: Flow-over-heated-plate-two-meshes: Different thickness of each domain
I am not sure if there was some technical reason we had to do this like this (e.g., surface area of the interface), but this currently looks strange and makes visualization complicated:

Screenshot from 2024-01-31 19-48-08

edit: originally observed in https://github.com/precice/tutorials/pull/271#issuecomment-1210179681

MakisH:
closes #455.

Screenshot from 2024-03-20 15-11-04


carme-hp (approved):
Looks good

PR #488, opened 2024-03-20
Remove FEniCSx participant from the partitioned-heat-conduction tutorial

IshaanDesai:
The FEniCSx participant from the partitioned-heat-conduction tutorial is moved to the branch: https://github.com/precice/tutorials/tree/fenicsx-cases because it is not intended to be included in the upcoming distribution release.


IshaanDesai:
> Looks good. Can you please open an overview issue here that we do not forget that a tutorial for FEniCSx existed at some point in time? Maybe we do not even need a branch, but pointing to https://github.com/precice/tutorials/tree/f6401828cc59b6b6252696345b9d40802035724e/partitioned-heat-conduction/fenicsx should be good enough, if anybody wants to revive the case it in the future.

https://github.com/precice/tutorials/issues/491

PR #489, opened 2024-03-20
Fix structure of volume-coupled-diffusion

BenjaminRodenberg:
See #461


volume-coupled-diffusion/drain-fenics/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -e -u
+
+python3 ../solver-fenics/volume-coupled-diffusion.py --drain
```
MakisH, line 4: We need to merge back from develop once #479 is merged (or merge this beforehand).

volume-coupled-diffusion/solver-fenics/output/.gitkeep
MakisH, line 1: Is this still needed for the `solver-fenics` directory? Should it be at the `source-fenics`, instead?

volume-coupled-diffusion/drain-fenics/clean.sh
MakisH, line 1: This cleaning script deletes also the `output/.gitkeep`, which then shows up in the `git status`.

I have the impression we don't need these files at all.

BenjaminRodenberg: Done. Good point.

volume-coupled-diffusion/solver-fenics/output/.gitkeep
BenjaminRodenberg: Done. Good point.

volume-coupled-diffusion/drain-fenics/clean.sh
BenjaminRodenberg: True. Removed them.

volume-coupled-diffusion/drain-fenics/run.sh
MakisH: I will do that in my branch.

MakisH (approved):

PR #490, opened 2024-03-20
Fix inconsistencies in structure of some tutorials

uekerman:
Part of #461 

- [x] oscillator
- [ ] partitioned-heat-conduction/nutils
- [ ] partitioned-heat-conduction-direct


oscillator/mass-left-fmi/run.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+set -e -u
+
```
MakisH, line 3: We need to cross-merge with #479.

MakisH:
@uekerman I resolved the merge conflicts, but it looks like I don't have permission to push to your fork. Can you please enable that?

I would then merge this and @Fujikawas and @IshaanDesai will continue with the rest of the cases.

PR #492, opened 2024-03-21
added v3 compatible reference versions

valentin-seitz:
This PR introduced a set of new references versions running with precice v3. 

This is needed to generate references results for upcoming V3 systemtests.


valentin-seitz:
Started action https://github.com/precice/tutorials/actions/runs/8372020938 to add reference results in this PR

MakisH (approved):
I thought I updated this file already yesterday, but I can find no document proving that or any related wrong update... :thinking: The versions here look correct, thanks!

MakisH (approved):

PR #496, opened 2024-03-22
Fix inconsistenies in the Nutils participants of the partitioned-heat-conduction tutorial

IshaanDesai:
Fix for what is discussed in https://github.com/precice/tutorials/issues/461


partitioned-heat-conduction/dirichlet-nutils/run.sh
MakisH, line 1: Are you using the latest state of develop? The run scripts do not include the logging.

partitioned-heat-conduction/dirichlet-nutils/heat.py
MakisH, line 1: The `dirichlet-nutils/heat.py` and `neumann-nutils/heat.py` are almost identical, apart mainly from names of variables. This is ok for now, but ideally, the solver should be in `solver-nutils/heat.py` and either:

- each `run.sh` would pass a `dirichlet` / `neumann` keyword to the solver, or
- each case directory would include a config file. See other nutils tutorials for an example.

MakisH (commented):
The directory structure is fine, the run scripts need a couple more lines for logging, and the solver could be moved into `solver-nutils`.

I could run the case.

MakisH (approved):
Logging looks fine now.
Let's keep the duplication now, we can make it nicer later.

@IshaanDesai the PEP8 check is failing.

PR #497, opened 2024-03-22
solve structure inconsistency of `partitioned-heat-conduction-direct`

Fujikawas:
Refer to issue #461


partitioned-heat-conduction-direct/dirichlet-nutils/heat.py
MakisH, line 1: Similarly to #496 (cc @IshaanDesai), the two solvers are almost identical, apart from configuration. This is fine for now, but ideally the solver could be merged into `solver-nutils/heat.py` and each case would read a short configuration file, or just a keyword `dirichlet`/`neumann`.

MakisH (commented):
The directory structure is fine, thank you! You could still consider merging the solvers into one and read configuration files.

PR #498, opened 2024-03-22
Add python variant of resonant circuit

BenjaminRodenberg:
Add a python version of the tutorial provided via #480. Uses a black-box time stepper with time interpolation and substeps for high accuracy. 

Planning to merge #480 first.


BenjaminRodenberg:
It's to some degree an open question where we want to go from here: I added adaptive time stepping and use time interpolation in this PR. This is similar to what I did in #500. From my perspective the resonant circuit is even a bit simpler and might help to illustrate advanced time stepping techniques quite well.

MakisH:
I ported the fix for the markdownlint CI check. This is now only failing for autopep8.

BenjaminRodenberg (commented):
What's still missing here: We need a `requirement.txt` and update this case to use `venv` (see #547). Either I or @NiklasVin will take care of this.

resonant-circuit/README.md
```diff
@@ -31,6 +31,7 @@ preCICE configuration (image generated using the [precice-config-visualizer](htt
 
 * *MATLAB* A solver using the [MATLAB bindings](https://precice.org/installation-bindings-matlab.html).
  Before running this tutorial, follow the [instructions](https://precice.org/installation-bindings-matlab.html) to correctly install the MATLAB bindings.
+* *Python* A solver using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html). This solver also depends on the Python libraries `numpy` and `scipy`, which you can get from your system package manager or with `pip3 install --user <package>`.
```
MakisH: Let's not suggest the pip install method here, since this is now changing and disallowed, e.g., in Ubuntu 24.04.

```suggestion
* *Python* A solver using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html). This solver also depends on the Python libraries `numpy` and `scipy`.
```

BenjaminRodenberg, line 34:
```suggestion
* *Python* A solver using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html). 
```
Good point. I would even go one step further: Dependencies are specified in the `requirements.txt`. If users run the solver as intended they don't have to worry about any dependencies. Let's not mention any dependencies here.

BenjaminRodenberg:
```suggestion
* *Python* A solver using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html). 
```
Good point. I would even go one step further: Dependencies are specified in the `requirements.txt`. If users run the solver as intended they don't have to worry about any dependencies. Let's not mention any dependencies here.

resonant-circuit/README.md
```diff
@@ -61,9 +62,13 @@ For that, modify the path in the file `matlab-bindings-path.sh` found in the bas
 
 By doing that, you can now open two shells and switch into the directories `capacitor-matlab` and `coil-matlab` and execute the `run.sh` scripts.
 
+### Python
+
+This example can be executed by starting the `run.sh` file located in each of the directories `coil-python` and `capacitor-python`.
+
```
BenjaminRodenberg:
```suggestion
```
No need to explicitly mention this. We have the convention that all tutorials are run this way.

BenjaminRodenberg:
I did a quick round of checks for the newly added `requirements.txt` and modified `./run.sh` script and everything works. Thanks for adding @NiklasVin !

NiklasVin:
@MakisH Everything should be fine now. Could you please review this PR again?

resonant-circuit/README.md
```diff
@@ -63,7 +64,7 @@ By doing that, you can now open two shells and switch into the directories `capa
 
 ## Post-processing
 
-The solver for the current also records the current and voltage through time and at the end of the simulation saves a plot with the obtained curves, as well as the analytical solution.
+The MATLAB participant for the current also records the current and voltage through time and at the end of the simulation saves a plot with the obtained curves, as well as the analytical solution.
```
MakisH: How can I check what the Python participant computes?

MakisH (commented):
I can run the tutorial (but don't know what results to check if I only run Python-Python).

The files looks very clean and ready to merge.

resonant-circuit/README.md
BenjaminRodenberg: We have a `plot-solution.sh` script that uses the watchpoints.

BenjaminRodenberg: If this is not clear enough: Feel free to add some notes in the documentation.

MakisH: This should be in this section, otherwise I need to guess that there is a script.

Also: The watchpoints seem to not be specified in the precice-config.

BenjaminRodenberg: Good point. Fixed it.

resonant-circuit/README.md
```diff
@@ -63,7 +64,9 @@ By doing that, you can now open two shells and switch into the directories `capa
 
 ## Post-processing
 
-The solver for the current also records the current and voltage through time and at the end of the simulation saves a plot with the obtained curves, as well as the analytical solution.
+As we defined a watchpoint on the 'Capacitor' participant (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-solution.sh.` You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `./plot-solution.sh capacitor-python`. The resulting graph shows the voltage and current exchanged between the two participants.
+
+Additionally, the MATLAB participant for the current also records the current and voltage through time and at the end of the simulation saves a plot with the obtained curves, as well as the analytical solution.
```
BenjaminRodenberg:
```suggestion
Additionally, the MATLAB participant `capacitor-matlab` records the current and voltage over time. At the end of the simulation it creates a plot with the computed waveforms of current and voltage, as well as the analytical solution.
```

resonant-circuit/README.md
MakisH: plot

Works, thanks!

MakisH (approved):

PR #499, opened 2024-03-25
Use dedicated OpenFOAM solver clang format to make solver compile again

davidscn:
Takes over the clang format we have in the adapter. The default clang-format is not compatible with the way OpenFOAM code looks like. The adapter style guide is also more readable (compare e.g. the `while` loop, where the physics are solved).

### DIFF
--- a/partitioned-heat-conduction/solver-openfoam/.clang-format
+++ b/partitioned-heat-conduction/solver-openfoam/.clang-format
@@ -1,0 +1,175 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+#
+# Proposed clang-format-11 style for OpenFOAM, trying to follow the OpenFOAM style guide:
+#   https://develop.openfoam.com/Development/openfoam/-/wikis/coding/style/style
+# Configuration developed for the OpenFOAM-preCICE adapter code:
+#   https://github.com/precice/openfoam-adapter
+# Contribute to the discussion at the respective OpenFOAM issue:
+#   https://develop.openfoam.com/Development/openfoam/-/issues/1634
+#
+# Keep `public:` at the first indentation level
+AccessModifierOffset: -4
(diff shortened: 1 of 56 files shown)


partitioned-heat-conduction/solver-openfoam/.clang-format
```diff
@@ -0,0 +1,175 @@
...
+AccessModifierOffset: -4
+# Undocumented guideline: align arguments after an open bracket.
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+# Align operands after operators (+,*,<<) (see BreakBeforeBinaryOperators)
+AlignOperands: AlignAfterOperator
```
MakisH, line 22: This seems to be failing for the pre-commit, which uses clang-format 8.
Let's maybe update to 11, used by the OpenFOAM adapter?

https://github.com/precice/openfoam-adapter/blob/develop/.github/workflows/check-format.yml

MakisH (commented):
The code compiles, but let's also upgrade the pre-commit hook.

partitioned-heat-conduction/solver-openfoam/.clang-format
MakisH: We should make the CI succeed before merging, otherwise it will show as failing forever, until we merge again into master.

MakisH: Updating to 14.0.0 is fine, and touching more files in this PR is also fine.

davidscn: Done

MakisH (approved):
I have not checked if all code compiles. Feel free to merge once you are confident about that.

PR #500, opened 2024-03-25
Oscillator: Use adaptive black-box time stepper and dense output.

BenjaminRodenberg:
Let black-box time stepper directly access preCICE `read_data` function to avoid need for interpolation. Sample from dense output to circumvent need for substeps.


oscillator/solver-python/oscillator.py
```diff
@@ -133,22 +133,22 @@ class Participant(Enum):
...
-    u_new, v_new, a_new = time_stepper.do_step(u, v, a, f, dt)
+    u_new, v_new, a_new, sol = time_stepper.do_step(u, v, a, f, dt)
     t_new = t + dt
 
-    write_data = [connecting_spring.k * u_new]
-
-    participant.write_data(mesh_name, write_data_name, vertex_ids, write_data)
-
-    participant.advance(dt)
+    n_pseudo = 5 * len(sol.ts)  # 5 times no. substeps should be on the safe side.
```
uekerman: Could you turn the "5" into a variable and iterate on the comment. I guess this is the most important part to understand the magic here.

uekerman (approved):
Clever strategy.
Do we need to update anything in the README? There, we still write about 4 time steps.

oscillator/solver-python/oscillator.py
BenjaminRodenberg:
```suggestion
    samples_per_step = 5  # create n samples_per_step of time stepping scheme. Degree of dense interpolating function is usually larger 1 and, therefore, we need multiple samples per step.
    n_time_steps = len(sol.ts)  # number of time steps performed by adaptive time stepper
    n_pseudo = samples_per_step * n_time_steps  # samples_per_step times no. time steps per window.
```
Is this clearer? Ideally we would not do a uniform sampling (`dt_pseudo = dt / n_pseudo`) below but sample non-uniformly depending on `sol.ts`. But this is future work :smirk:

uekerman: yes, good now :+1:

BenjaminRodenberg:
@MakisH The CI Pipeline is failing. I think at some point you said this is a general problem you can fix somehow?

MakisH:
> @MakisH The CI Pipeline is failing. I think at some point you said this is a general problem you can fix somehow?

Fixed. Sorry for the mess...

BenjaminRodenberg:
I did a final round of testing and will merge this PR now.

PR #449, opened 2024-01-24
Add error computation for partitioned-heat-conduction OpenFOAM solver

davidscn:
To help testing and comparing.

Related to https://github.com/precice/tutorials/issues/445

In our tutorials, with OpenFOAM - OpenFOAM, I get

Dirichlet:
```
Error metrics at t = 1s:
l2 error (sqrt(sum(err^2)/n)):		6.2733e-05
Maximum absolute error (max(err)):	7.72464e-05
Global absolute error (sum(err^2)):	3.93543e-05
```

Neumann
```
Error metrics at t = 1s:
l2 error (sqrt(sum(err^2)/n)):		6.24377e-05
Maximum absolute error (max(err)):	7.4964e-05
Global absolute error (sum(err^2)):	3.89846e-05
```

PR #501, opened 2024-03-25
Clean up coded boundary conditions for OpenFOAM cases

davidscn:
Follow-up to #428

Cleans-up the run/clean scripts as we use coded boundary conditions and initial conditions since #428, i.e., tools such as `funkySetFields` and `groovyBC` are not required anymore. Also removes the `0.orig`, as it is not needed anymore. Applies some formatting to the coded boundary conditions itself and applies the (promised) boundary condition correction in the `heatTransfer´ solver before entering the time loop.


partitioned-heat-conduction/dirichlet-openfoam/system/fvSolution
MakisH, line 1: Why do we need the changes in this file?

davidscn: Not needed, strictly speaking, but it makes sense since we have the error computation #449 to have the numerical settings consistent in terms of accuracy.

MakisH (approved):
I modified the `.gitignore` to add the generated `dynamicCode` directory (but I guess we don't want to clean it, as we don't clean the solver codes as well).

I ran the `partitioned-heat-conduction` case. I trust you have checked the `turek-hron-fsi3` case.

davidscn:
> (but I guess we don't want to clean it, as we don't clean the solver codes as well)

changing the OpenFOAM version might lead to obscure errors, but we can still add something like this later on, if we find that it is important.

Issue #504, opened 2024-04-02
Conversion errors in FEniCS FSI cases


MakisH:
I cannot run the following cases/solvers (anymore):

- [`perpendicular-flap/solid-fenics/solid.py`](https://github.com/precice/tutorials/blob/da7a1494f5c36b4ef509daf2a43bfee42fb32d9d/perpendicular-flap/solid-fenics/solid.py)
- [`elastic-tube-3d/solid-fenics/solid.py`](https://github.com/precice/tutorials/blob/da7a1494f5c36b4ef509daf2a43bfee42fb32d9d/elastic-tube-3d/solid-fenics/solid.py)

While I can successfully run:

- [`flow-over-heated-plate/solid-fenics/solid.py`](https://github.com/precice/tutorials/blob/da7a1494f5c36b4ef509daf2a43bfee42fb32d9d/flow-over-heated-plate/solid-fenics/solid.py)
- [`partitioned-heat-conduction/solver-fenics/heat.py`](https://github.com/precice/tutorials/blob/da7a1494f5c36b4ef509daf2a43bfee42fb32d9d/partitioned-heat-conduction/solver-fenics/heat.py)

The solver starts, but at the time it tries to ?read? a checkpoint, the following happens:

```
---[precice]  iteration: 1 of 30 (min 1), time-window: 1 of 100, time: 0, time-window-size: 0.0001, max-time-step-size: 0.0001, ongoing: yes, time-window-complete: no, write-iteration-checkpoint 
Traceback (most recent call last):
  File "/home/gc/repos/precice/tutorials/elastic-tube-3d/solid-fenics/solid.py", line 158, in <module>
    res = m(avg(a_n, a_np1, alpha_m), v) + k(avg(u_n, du, alpha_f), v)
  File "/home/gc/repos/precice/tutorials/elastic-tube-3d/solid-fenics/solid.py", line 101, in k
    return inner(sigma(u), sym(grad(v))) * dx
  File "/home/gc/repos/precice/tutorials/elastic-tube-3d/solid-fenics/solid.py", line 91, in sigma
    return lambda_ * nabla_div(u) * Identity(dim) + 2 * mu * epsilon(u)
  File "/home/gc/.local/lib/python3.10/site-packages/ufl/operators.py", line 391, in nabla_div
    f = as_ufl(f)
  File "/home/gc/.local/lib/python3.10/site-packages/ufl/constantvalue.py", line 501, in as_ufl
    raise ValueError(
[...]

PR #505, opened 2024-04-02
Fix ufl compatibility issues for `perpendicular-flap/solid-fenics` and `elastic-tube-3d/solid-fenics`

BenjaminRodenberg:
See #504.

Replacing `nabla_div` and `nabla_grad` with `div` and `grad`.

### DIFF
--- a/elastic-tube-3d/solid-fenics/solid.py
+++ b/elastic-tube-3d/solid-fenics/solid.py
@@ -1,9 +1,8 @@
 # Import required libs
 from fenics import Constant, Function, AutoSubDomain, VectorFunctionSpace, interpolate, \
-    TrialFunction, TestFunction, Point, Expression, DirichletBC, nabla_grad, \
-    Identity, inner, dx, ds, sym, grad, lhs, rhs, dot, File, solve, assemble_system
+    TrialFunction, TestFunction, Point, Expression, DirichletBC, \
+    Identity, inner, dx, ds, sym, grad, div, lhs, rhs, dot, File, solve, assemble_system
 from mshr import Cylinder, generate_mesh
-from ufl import nabla_div
 import numpy as np
 from fenicsprecice import Adapter
 import math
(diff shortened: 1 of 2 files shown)


MakisH (commented):
For the perpendicular-flap, this seems to do the trick and produce the same results:

tutorials-perpendicular-flap-displacement-all-watchpoints

I ported the same changes in the elastic-tube-3d. This takes too long to run till the end, but running till t=0.0042 seems to already produce reasonable results. At least the result looks symmetric and I can see the pulse propagating. I will eventually let it run completely and update the images in https://precice.org/tutorials-elastic-tube-3d.html#post-processing.

I also removed some left-over mentions to ufl in the form of `if ufl: <A> else: <B>`, since the first would now be a dead branch and confusing.

We still use ufl in the `solid-dunefem` case (I still want to run that).

MakisH (approved):
The following cases work for me:

- `elastic-tube-3d/solid-fenics`
- `perpendicular-flap/solid-fenics`
- `flow-over-heated-plate/solid-dunefem` (which still uses ufl, but there seems to be no dependency issue)

elastic-tube-3d/solid-fenics/solid.py
```diff
@@ -140,8 +131,8 @@ def update_fields(u, u_old, v_old, a_old):
     v0_vec, a0_vec = v_old.vector(), a_old.vector()
 
     # call update functions
-    a_vec = update_a(u_vec, u0_vec, v0_vec, a0_vec, ufl=False)
-    v_vec = update_v(a_vec, u0_vec, v0_vec, a0_vec, ufl=False)
+    a_vec = update_a(u_vec, u0_vec, v0_vec, a0_vec)
+    v_vec = update_v(a_vec, u0_vec, v0_vec, a0_vec)
```
BenjaminRodenberg, line 135: The `if `ufl`: ... else: ...` is not a dead branch. See below.

elastic-tube-3d/solid-fenics/solid.py
```diff
@@ -152,8 +143,8 @@ def avg(x_old, x_new, alpha):
     return alpha * x_old + (1 - alpha) * x_new
 
 
-a_np1 = update_a(du, u_n, v_n, a_n, ufl=True)
-v_np1 = update_v(a_np1, u_n, v_n, a_n, ufl=True)
+a_np1 = update_a(du, u_n, v_n, a_n)
+v_np1 = update_v(a_np1, u_n, v_n, a_n)
```
BenjaminRodenberg, line 147: Here `ufl=True`

BenjaminRodenberg (commented):
I don't think that removing the `ufl` part is a good idea. It's also explicitly done this way in the tutorial that this case is based on. See https://comet-fenics.readthedocs.io/en/latest/demo/elastodynamics/demo_elastodynamics.py.html. @MakisH Can you please revert this part?

PR #508, opened 2024-04-03
Revert some of the ufl-relateve changes from #505

MakisH:
I took too much liberty in #505 and removed some code that was not supposed to be removed, and merged before confirming. This PR reverts that change.

FYI @BenjaminRodenberg 

I will test the cases and merge, even if I don't get a review until then, since I am currently trying to disentangle a bunch of inter-dependent work packages. Please complain again if you spot any mistakes.


MakisH (commented):
Comparing the files from this state and an older state, I only see the changes related to `ufl_grad` and `ufl_div`. Both the perpendicular-flap and the elastic-tube-3d are running.

PR #509, opened 2024-04-03
Update elastic-tube-3d images

MakisH:
This PR updates the images used in the elastic-tube-3d tutorial, based on:

- Ubuntu 22.04
- preCICE v3.1.0
- OpenFOAM v2312 and OpenFOAM-preCICE v1.3.0
- CalculiX v2.20 and CalculiX-preCICE v2.20.1
- FEniCS 2019.2.0.13.dev0 and fenicsprecice v2.1.0

The OpenFOAM-FEniCS case seems to be producing significantly different results than before. @BenjaminRodenberg @IshaanDesai is this something you would expect? They definitely look in better agreement now with OpenFOAM-CalculiX.

The surface plot is OpenFOAM-FEniCS.

[precice-Solid-FEniCS-watchpoint-Tube-Midpoint.log](https://github.com/precice/tutorials/files/14858282/precice-Solid-watchpoint-Tube-Midpoint.log)

[precice-Solid-CalculiX-watchpoint-Tube-Midpoint.log](https://github.com/precice/tutorials/files/14858284/precice-Solid-watchpoint-Tube-Midpoint.log)


IshaanDesai:
Previously we justified the disagreement in the results by stating that the FEniCS structural solver is linear, and we use nonlinear geometry in CalculiX. This is still the case, so the sudden agreement is puzzling. But in principle agreement of results is a good thing.

IshaanDesai (approved):

PR #511, opened 2024-04-03
Update multiple-perpendicular-flaps picture and documentation

MakisH:
@davidscn had some issues reproducing this picture at some point, I hope that the additional details and the missing scaling factor helps.

This also changes the threshold for the size check to 750KB. I think that, back in the day, we set it to 250KB because is was easy to achieve, but I think we are suffering too much without a concrete reason. Some images are just larger, and here I want to show the details of the mesh.


multiple-perpendicular-flaps/README.md
```diff
@@ -107,6 +107,8 @@ The scenario settings are implemented similarly for the nonlinear case.
 
 After the simulation has finished, you can visualize your results using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `fluid-openfoam.foam` file. Looking at the fluid results is enough to obtain information about the behaviour of the flaps. You can also visualize the solid participants' vtks though.
 
+Example visualization of the fluid velocity and the solid displacement at t=2.3s, after synchronizing the results with a `Temporal Shift Scale` filter (scale for the deal.II results: 0.1) and a `Warp by vector` with `cellDisplacement` and `Displacement` (with scale 1):
```
davidscn, line 110: Why do you need the warp by vector here? IIRC both, OpenFOAM and deal.II export the deformed grid.

MakisH: Ah, thanks, I had the impression deal.II VTK files were not deformed. Is this the case for all of our deal.II tutorials?

I updated the README and the picture (showing strain_xx instead of displacement, I think this gives a better impression).

The gap in the downstream flap is still there, not sure if we can do anything about it.

MakisH: tutorials-multiple-perpendicular-flaps-results

I guess I am just overthinking here.

davidscn: A bit hard to tell, but let's not worry here at the moment.

uekerman: Could also be the mapping error.

MakisH (commented):
I assume from the comments that this has been implicitly approved.

PR #512, opened 2024-04-04
Decrease runtime of flow around controlled moving cylinder by stopping at t=3 already

fsimonis:
When running the `flow-around-controlled-moving-cylinder`, the simulation runs for longer than it needs to. We could easily already stop at `t=3` without loosing any value.

Currently, then watchpoints look like this:

image

image

image

image


uekerman (approved):
Anything in the README that needs updating?

PR #516, opened 2024-04-08
Make Nutils participant `micro-nutils` in the two-scale-heat-conduction tutorial faster

IshaanDesai:
This PR removes the initial correction calculation of the phase field in the `micro-nutils` participant of the two-scale-heat-conduction tutorial. This calculation was originally implemented so that the micro simulation will start with a very accurate grain description. However, if the micro structure (grain) is not represented extremely accurately, the resultant phase field and the average porosity will still yield sensible values. Therefore this calculation can be commented out to cut down the initialization time and make the `micro-nutils` participant faster.

## Approximate runtime

- On 4 processors: 6.5 minutes
- On 2 processors: 10.5 minutes

### DIFF
--- a/two-scale-heat-conduction/micro-nutils/micro-manager-config.json
+++ b/two-scale-heat-conduction/micro-nutils/micro-manager-config.json
@@ -9,7 +9,7 @@
     },
     "simulation_params": {
       "macro_domain_bounds": [0.0, 1.0, 0.0, 0.5],
-      "decomposition": [4, 1],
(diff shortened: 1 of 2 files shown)


mathiskelm (approved):

MakisH:
How long was it before?
Does the time hint in the README need to be updated?

IshaanDesai:
> How long was it before? Does the time hint in the README need to be updated?

Good point, I will update the README hint.

PR #521, opened 2024-04-12
Add su2 adapter tests and update reference results

MakisH:
Adds an SU2 component and tests using the SU2 adapter.

@valentin-seitz somehow this gives me empty logs (`stdout.log`, `stderr.log`) after running for a long time. Any clue why?

I know that at least the meson build step in the Dockerfile is running, and I think it probably gets stuck in the ninja install.

See this run: https://github.com/precice/tutorials/actions/runs/8665208853

Reproduce the run:

```shell
gh workflow run run_testsuite_manual.yml -f suites=su2_test -f systests_branch=add-su2-adapter-tests -f loglevel=DEBUG -f build_args="PRECICE_REF:v3.1.1,OPENFOAM_ADAPTER_REF:v1.3.0,PYTHON_BINDINGS_REF:v3.1.0,FENICS_ADAPTER_REF:v2.1.0,TUTORIALS_REF:45650f2" --ref=add-su2-adapter-tests
```

Reproduce locally:

```shell
python3 systemtests.py --suites=su2_test --log-level=DEBUG --build_args="PRECICE_REF:v3.1.1,OPENFOAM_ADAPTER_REF:v1.3.0,PYTHON_BINDINGS_REF:v3.1.0,FENICS_ADAPTER_REF:v2.1.0,TUTORIALS_REF:45650f2"
```


MakisH:
Note that I don't always get empty logs. In previous states of debugging, I was getting output. I think that, whenever it is building the adapter, the log gets too long or something similar. But even when I build based on the cached layer (that's why I split the meson build and the ninja install), I get no output.

MakisH:
Or maybe the Docker layers were cached, so nothing was printed in the `stdout.log`, but somehow the running did not print any logs.

MakisH:
It is actually failing in the build step, due to dependencies (e.g., mpi4py).
Logging continues to be a puzzle: sometimes the `stdout.log` has content, sometimes not.

MakisH:
New observation: [This run](https://github.com/precice/tutorials/actions/runs/86663342450) actually completed the build step, and failed in the beginning of the run step. No `stdout.log` was written at all (not an empty file; just no file at all), while we got simulation logs.

MakisH:
The system tests now succeed: https://github.com/precice/tutorials/actions/runs/8672889794/job/23783659705

I have updated all reference results in this PR as well (we cannot only update reference results for specific tests with the current workflow).

I also needed to increase the global timeout, as the SU2-FeniCS test currently takes 8min.

@valentin-seitz could you please have a quick sanity check? Have I forgotten anything? Have I misused anything?

Regarding the logging issue, let's postpone and offload it here: https://github.com/precice/tutorials/issues/522

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -126,3 +126,31 @@ RUN git clone https://github.com/precice/calculix-adapter.git && \
     git checkout ${CALULIX_ADAPTER_REF} &&\
     make CXX_VERSION=${CALULIX_VERSION} ADDITIONAL_FFLAGS="-fallow-argument-mismatch" -j $(nproc) && \
     ln -s /home/precice/calculix-adapter/bin/ccx_preCICE /home/precice/.local/bin/ccx_preCICE
+
+FROM python_bindings as su2_adapter
```
MakisH, line 130: @valentin-seitz this is something I would like feedback on. I see that the FEniCS adapter starts from `precice_dependencies`, and then it copies files from `python_bindings`, but both should be `python_bindings`, right?

MakisH: It sounds like there was intention in making things this way in the FEniCS adapter. Since I don't want to open new boxes before this release, I will fix it in a separate PR.

valentin-seitz:
> The system tests now succeed: https://github.com/precice/tutorials/actions/runs/8672889794/job/23783659705
> 
> I have updated all reference results in this PR as well (we cannot only update reference results for specific tests with the current workflow).
> 
> I also needed to increase the global timeout, as the SU2-FeniCS test currently takes 8min.
> 
> @valentin-seitz could you please have a quick sanity check? Have I forgotten anything? Have I misused anything?
> 
> Regarding the logging issue, let's postpone and offload it here: #522

I think we discussed most of it already via phone. 
The last safeguard is the adding of the reference results. Here the pipeline should not fail. Which it did not, so :+1:

valentin-seitz (approved):

PR #524, opened 2024-04-16
Quickstart: Switch to directly downloading release artifacts

MakisH:
- Pointing to the specific releases
~~- Adding a note that they might still not be compatible, linking to the distribution.~~


quickstart/README.md
```diff
@@ -52,7 +52,7 @@ To get a feeling what preCICE does, watch a [short presentation](https://www.you
     sudo apt install build-essential pkg-config cmake git
     ```
 
-4. Download and install the [OpenFOAM-preCICE adapter](https://precice.org/adapter-openfoam-get.html):
+4. Download and install the [OpenFOAM-preCICE adapter](https://precice.org/adapter-openfoam-get.html) from the [latest release](https://github.com/precice/openfoam-adapter/releases/latest), or get the latest `master` branch of the repository:
```
davidscn, line 55: Didn't we want users to use the release archives? Also to accommodate for the different OpenFOAM versions available.

quickstart/README.md
```diff
@@ -70,6 +70,12 @@ To get a feeling what preCICE does, watch a [short presentation](https://www.you
 
 If you prefer to easily try everything in an isolated environment, you may prefer using our [demo virtual machine](https://precice.org/installation-vm.html).
 
+{% note %}
+We are assuming here that the latest `master` branches refer to the latest releases, and that they all work together.
+This might not be the case in transition periods after a major preCICE release (e.g., v2 to v3).
+Versions that are guaranteed to work together are described in the [preCICE Distribution](https://precice.org/installation-distribution.html).
+{% endnote %}
+
```
davidscn, line 78: I'm not too sure about this Note. The transition period essentially ends with the release, where you want to include it. To me, it confuses more than helps. It gives overall the impression of a fragile system, where you have to write down which versions and packages are compatible with each other. That's not really the case for preCICE and its components.

quickstart/README.md
MakisH: Yes, but how would you write that in one command? We would always need to update the version number manually.

quickstart/README.md
MakisH: I see your point, and I was mainly aiming for the v3->v4 transition. I can remove this, but then we can just close this PR.

quickstart/README.md
MakisH: @davidscn we can still merge this part, so that at least we hint towards releases.

We could add the versions as variables, but these would only render on the website, and we would need to remember to update them whenever we want users to use something new.

davidscn:
```suggestion
4. Download and install the [OpenFOAM-preCICE adapter](https://precice.org/adapter-openfoam-get.html) from the [latest release](https://github.com/precice/openfoam-adapter/releases/latest)´
```
Just like this? In the release, we anyway distinguish the OpenFOAM version?!

quickstart/README.md
MakisH: Removed

quickstart/README.md
MakisH: Yes, good point, we already specify the OpenFOAM version here, so let's keep everything fail-safe: we ask users to install specific versions. Even if we don't immediately update everything, these should still work.

quickstart/README.md
```diff
@@ -52,23 +52,25 @@ To get a feeling what preCICE does, watch a [short presentation](https://www.you
...
 
     ```bash
-    git clone --branch=master --depth 1 https://github.com/precice/tutorials.git
+    wget https://github.com/precice/tutorials/releases/tag/v202404.0
+    tar -xzf v202404.0.tar.gz
     cd tutorials/quickstart
     ```
 
-If you prefer to easily try everything in an isolated environment, you may prefer using our [demo virtual machine](https://precice.org/installation-vm.html).
+If you prefer to easily try everything in an isolated environment, you may prefer using our [demo virtual machine](https://precice.org/installation-vm.html). If you are looking for previous versions, check out the [preCICE Distribution](https://precice.org/installation-distribution.html).
```
MakisH, line 73: I added again a mention to the preCICE Distribution, this time only as a hint for older versions.

I am not sure who would look for older versions, but I feel this is a good place to tell users "hey! you are looking for something you found in a paper? Maybe look here, instead before you start guessing which versions you need".

quickstart/README.md
```diff
@@ -52,23 +52,25 @@ To get a feeling what preCICE does, watch a [short presentation](https://www.you
     sudo apt install build-essential pkg-config cmake git
     ```
 
-4. Download and install the [OpenFOAM-preCICE adapter](https://precice.org/adapter-openfoam-get.html) from the [latest release](https://github.com/precice/openfoam-adapter/releases/latest), or get the latest `master` branch of the repository:
+4. Download and install the [OpenFOAM-preCICE adapter](https://precice.org/adapter-openfoam-get.html) from the [latest release](https://github.com/precice/openfoam-adapter/releases/latest), unpack it, and build the adapter:
 
     ```bash
-     git clone --branch=master --depth 1 https://github.com/precice/openfoam-adapter
-     cd openfoam-adapter
+     wget https://github.com/precice/openfoam-adapter/archive/refs/tags/v1.3.0.tar.gz
```
davidscn, line 58: If you would want, we could also make this version agnostic and just download the latest version:
```suggestion
LATEST_RELEASE_JSON=$(curl -s https://api.github.com/repos/precice/openfoam-adapter/releases/latest)

# Use jq to parse JSON and get the tarball URL, we could potentially use grep here as well
TARBALL_URL=$(echo $LATEST_RELEASE_JSON | jq -r ".tarball_url")
wget $TARBALL_URL
```

quickstart/README.md
davidscn:
```suggestion
If you prefer to easily try everything in an isolated environment, you may prefer using our [demo virtual machine](https://precice.org/installation-vm.html). If you are looking for previous versions or face compatibility issues, check out [preCICE Distribution](https://precice.org/installation-distribution.html).
```

davidscn: The intrinsic problem here is also, that people typically don't recognize compatibility issues as such. If someone couldn't locate a `SolverInterface.hpp`, most likely this doesn't sound as if you would use the wrong version, right? It's rather a bogus installation..

quickstart/README.md
MakisH: But then we are back to square one: We make a preCICE release that does not work with the latest tutorial release, and we have the same issue.

quickstart/README.md
MakisH: Isn't this then the same as the note I originally had and showed a fragile system? I agree we should not give that impression, but I am trying to find a subtle way to inform the users about such a potential (but maybe unlikely) issue.

quickstart/README.md
davidscn: We can only refer to v1.3.0 here because it's out now. If it wouldn't be out, we would face the incompatibility anyway, right?

quickstart/README.md
davidscn: Hm, I guess it's more about the wording. Here, it says: if you face problems, look in the distri. In the former version it was: All versions which are known to be compatible are in our distribution. To me, there is a slight difference in the underlying assumption.

quickstart/README.md
MakisH: Yes, but then we would only update this page once all releases of the required components are out.

quickstart/README.md
MakisH: Convincing enough for me, thanks! (and added back a missing `the`)

quickstart/README.md
davidscn: I think the minor releases, which would require an update here are more frequent than the major releases you have in mind. Still, I don't have a strong opinion, so you could also just stick to fixed versions, if you think it is more suitable.

MakisH: Good point about the minor releases. If you want to try and extract this from the strings curl returns, go ahead. I would be very happy if we can merge this today.

uekerman (commented):
I am not sure I can still follow the motivation of this PR. What we want in the quickstart is sth that just works. We will have this again once the tutorials are released. And if we want to use a different preCICE or OpenFOAM adapter version, we can easily change this in the future as long as we render the develop branch.

In the end of the page, we point to the docs. That's where users than learn how to install things, which versions are compatible, what the distribution is etc.

We should be careful not to overwhelm the reader here.

MakisH:
@uekerman so, you prefer master branches over release artifacts for specific versions?

uekerman:
> @uekerman so, you prefer master branches over release artifacts for specific versions?

Currently we do a release for preCICE and master branches for the OpenFOAM adapter and the tutorials. We could switch to releases for the last two as well if things do not get more complicated for the user, but not a must for me. What we should not do is to add additional links etc IMO.

quickstart/README.md
MakisH: Let's keep the pinned versions for now, we can attempt a better solution later.

MakisH:
What I am trying to avoid is running again into a situation where the Quickstart does not work.

I removed all links and text changes. The only change now is `git clone master` --> `wget + tar`.

The only downside of this is that we have to update whenever we want that users use newer versions.

uekerman (approved):

PR #528, opened 2024-04-22
Two scale heat conduction: Enable running DuMux participants in parallel

mathiskelm:
Enables running both DuMux participants in parallel.

Checklist:

### DIFF
--- a/two-scale-heat-conduction/README.md
+++ b/two-scale-heat-conduction/README.md
@@ -59,7 +59,7 @@ If you want to use DuMu<sup>x</sup>, use `cd macro-dumux` instead of `cd macro-n
 
 ## Running the simulation in parallel
 
-Only the `micro-nutils` participant can be run in parallel. To run it in parallel, run:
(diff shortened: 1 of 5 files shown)


Fujikawas:
Thanks for the push. Should we also mention that if the micro manager config is not changed, `micro-dumux` only works with `-p 2`?

mathiskelm:
> Thanks for the push. Should we also mention that if the micro manager config is not changed, `micro-dumux` only works with `-p 2`?

Both micro participants should be behaving the same way with respect to parallelisation, i.e. you can pass the number of processes to the run script and the decomposition in `micro-X/micro-manager-config.json` has to produce at least that many subdomains as [README.md#L69](https://github.com/precice/tutorials/blob/4f2323ca0e9359283ee6850eb82cb35b1e063d84/two-scale-heat-conduction/README.md?plain=1#L69) already mentions.

IshaanDesai (approved):
Glossed over and looks good :+1: Testing I leave to @Fujikawas

Fujikawas:
Test with both parallel `macro-dumux` and `micro-dumux` could work :)

Fujikawas (approved):

Issue #526, opened 2024-04-16
Python packages not installed automatically for all tutorials


MakisH:
We now install Nutils in a venv for every related tutorial. We could do the same for:

- [ ] fmi runner
- [ ] fenicsprecice

fsimonis:
Note that the fmi-runner and the micro-manager can already be installed via pipx for the current user.

In general, using a venv directly is always an option though.

MakisH:
If both are only stand-alone applications (I assume, but did not test), this is probably what we should document and maybe use in the VM.

But, independent of the compatibility issue, I think that formally specifying the dependencies is a goal we should aim for.

PR #533, opened 2024-04-29
Directly use micro_manager in two-scale- tutorial

fsimonis:
This tutorial directly uses the micro_manager binary. This allows using the tutorial with a pipx-installed mirco-manager.

Related to https://github.com/precice/vm/pull/94 https://github.com/precice/tutorials/issues/526

Checklist:


IshaanDesai (approved):
Thanks! In the past this did not work for parallel execution of the Micro Manager. But now it appears to work, which is good news!

PR #535, opened 2024-05-06
Fix shebangs to `#!/usr/bin/env bash` in all .sh files

cheriimoya:
This allows all scripts to be run without hardcoding `/bin/bash` as an interpreter.
Maybe this is an improvement, maybe not.
Someone should take a look at https://unix.stackexchange.com/a/29620 and document a decision for/against `/bin/(ba)sh` vs `/usr/bin/env bash`

Checklist:

- [ ] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`. -> will probably not affect many users (at least not any users using debian/ubuntu)

### DIFF
--- a/breaking-dam-2d/fluid-openfoam/clean.sh
+++ b/breaking-dam-2d/fluid-openfoam/clean.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 set -e -u
 
 . ../../tools/cleaning-tools.sh
--- a/channel-transport-reaction/chemical-fenics/run.sh
+++ b/channel-transport-reaction/chemical-fenics/run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e -u
 
 . ../../tools/log.sh
--- a/aste-turbine/clean.sh
+++ b/aste-turbine/clean.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 set -e -u
 
 . ../tools/cleaning-tools.sh
--- a/breaking-dam-2d/fluid-openfoam/run.sh
+++ b/breaking-dam-2d/fluid-openfoam/run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e -u
 
 . ../../tools/log.sh
--- a/breaking-dam-2d/solid-calculix/clean.sh
+++ b/breaking-dam-2d/solid-calculix/clean.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 set -e -u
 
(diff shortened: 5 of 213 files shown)


breaking-dam-2d/fluid-openfoam/clean.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/urs/bin/env bash
```
davidscn:
```suggestion
#!/usr/bin/env bash
```

davidscn (commented):
>  Maybe this is an improvement, maybe not.

I would be in favor of using `env` for this purpose. We had already similar issues in ASTE. Problem: on compute clusters, the hardcoded path might not lead to the version you actually want to use. Here, `env` takes care of this.

The only change I am not sure about is that you changed the requirement from a POSIX shell to bash. Why not using `#!/urs/bin/env sh` instead?

MakisH:
I thought of this while first creating all these scripts (or at least when making them consistent, in their current version), but I don't remember why I decided to keep the hard-coded path.

If this has already triggered issues, then let's indeed change it. The version argument is a valid one, apparently not only for clusters, but [also for macOS](https://unix.stackexchange.com/a/772779/36693).

However, I agree with @davidscn: I chose `sh` instead of `bash` as the most portable option, and we have invested some effort to keep scripts compatible with `sh`. Remember that some people use `zsh`, some `csh`, some `ksh`. Not all shells are compatible with `bash`. Let's keep it as `sh`.

breaking-dam-2d/fluid-openfoam/clean.sh
cheriimoya: good catch!

cheriimoya:
Okay, so this time it's using `#!/usr/bin/env sh` everywhere

cheriimoya:
But interesting: Look at the failing pipeline!

channel-transport-reaction/chemical-fenics/run.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
```
davidscn: To be more concise with my previous comment: the way we call the shell should not alter the shell environment. Not all scripts are POSIX compatible.

MakisH: I agree with David: it is intentional that most scripts are `sh` but a few are `bash`. In any case, the kind of shell is not something we should change here.

cheriimoya:
Okay, I've split the commit in two parts. Now I got what you meant @davidscn :D
Also grep'ed for `/urs`;)

Commands I used for reference: `sed 's~#!/bin/sh~#!/usr/bin/env sh~' $(rg '#!/bin/sh' -l) -i` (rg is ripgrep which is roughly equivalent to `grep -r`)

davidscn:
Thanks!

PR #539, opened 2024-06-10
Perpendicular flap: Document a stronger coupling condition

Fujikawas:
With this change the coupling becomes stronger for the perpendicular flap case. It will require implicit coupling AND acceleration methods to converge. With $\rho=42$, it requires only implicit coupling, no acceleration is necessary. With current $\rho=3000$, both explicit and implicit coupling could lead to convergence.
With $\rho=1$, the result from `fluid-openfoam`-`solid-dealii`  looks like:
image
With $\rho=42$, the result from `fluid-openfoam`-`solid-dealii`  looks like:
image

Checklist:


uekerman:
Worth considering on top: We could make the fluid inflow time-dependent. Then, also the $\rho=3000$ scenario could look interesting again.

MakisH:
@Fujikawas please add in a comment the new figures you got in the context of https://github.com/MakisH/ofw19-training

MakisH:
This has high educational value, as it demonstrates a case in which implicit coupling is important.
To not trigger a cascade of side changes that require a lot of effort, let's convert this to a documentation note, encouraging the user to try a different density and observe the effect. (@Fujikawas)

MakisH (approved):
I directly fixed a few issues and removed the intermediate step of ρ=42 and no acceleration, as that I think is more confusing than helps.

Current rendering:

Screenshot from 2025-07-31 21-26-50

PR #541, opened 2024-06-17
Add solid fake for perpendicular flap tutorial

NiklasVin:
I added a solid fake implemented in Python to the perpendicular flap tutorial, similar to the already existing fluid-fake.


perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,150 @@
...
+import numpy as np
+import precice
+
+
+def displacement_param(t):
+    # this parameter exists to define a time-dependent displacement of the flap's tip
+    if t > 1:
+        return 0.5
+    else:
+        return 0.5 * t
```
BenjaminRodenberg:
```suggestion
def displacement_param(t):
    # this parameter exists to define a time-dependent displacement of the flap's tip
    max_d = 0.5
    if t > 1:
        return max_d
    else:
        return max_d * t
```
To avoid code duplication.

You could also do `min(max_d, max_d*t)` but I personally prefer the `if...else`-way of writing things.

MakisH:
Is this ready for review?
Don't forget to add it into the README of the tutorial. I can generate some of the plots when reviewing.

I will fix the markdownlint, it is failing also elsewhere.

MakisH:
@NiklasVin please rebase/merge from develop, the markdownlint issue is now fixed.

NiklasVin:
> Is this ready for review? Don't forget to add it into the README of the tutorial. I can generate some of the plots when reviewing.
> 
> I will fix the markdownlint, it is failing also elsewhere.

Yes, now this should be ready for review.

MakisH:
@NiklasVin can you please enable edits by maintainers? It's a checkbox on the right.

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import division
+
+import numpy as np
+import precice
```
MakisH, line 4: Please add a `requirements.txt` and a `run.sh`, same as in the `fluid-fake`.
See also #548 for an update.

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,76 @@
...
+y_top = y_bottom + H  # top of the flap
+
+n = 24  # Number of vertices per side
+t = 0
+
+vertices_mid = np.zeros((2 * n, dimensions))
+vertices_mid[:n, 1] = np.linspace(y_bottom, y_top, n)
+vertices_mid[n:, 1] = np.linspace(y_bottom, y_top, n)
+
+vertex_ids = interface.set_mesh_vertices(mesh_name, vertices_mid)
```
MakisH: Trying to run the `solid-fake`, either with `fluid-fake` or with `fluid-openfoam`, I got the following error (preCICE 3.1.2):

```
---[precice] ERROR:  The interpolation matrix of the RBF mapping from mesh "Solid-Mesh" to mesh "Fluid-Mesh" is not invertable. This means that the mapping problem is not well-posed. Please check if your coupling meshes are correct (e.g. no vertices are duplicated) or reconfigure your basis-function (e.g. reduce the support-radius).
```

NiklasVin: This problem should be fixed now

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,78 @@
...
+
+vertices = np.zeros((2 * n, dimensions))
+vertices[:n, 1] = np.linspace(y_bottom, y_top, n)
+vertices[n:, 1] = np.linspace(y_bottom, y_top, n)
+vertices[:n, 0] = x_left
+vertices[n:, 0] = x_right
+
+vertex_ids = interface.set_mesh_vertices(mesh_name, vertices)
+
+interface.initialize()
```
BenjaminRodenberg, line 68: Can you add a few lines above to provide initial data? Compare [here](https://github.com/precice/tutorials/blob/0df4f0085a3aca71f42e7806dc3d762eab54624c/perpendicular-flap/fluid-fake/fake.py#L34-L38). It's not required by the `precice-config.xml` of this case but if we initialize data this makes the script more flexible and I assume you also still remember all the pain we had with `fluid-fake` and missing initial data :smirk:

BenjaminRodenberg (changes requested):
I checked the case with `fluid-openfoam`. Generally runs and VTK output looks good. Only a minor point is left + some readability comments.

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,78 @@
...
+import precice
+
+
+def displace_flap(x, y, t, flap_tip_y):
+    x_displ = np.zeros_like(x)
+    y_displ = np.zeros_like(y)
+    # first, get displacement independent of x, only dependent on y and t
+    # maximal displacement at flap tip should be 0.5
+    # initially, flap's displacement is 0
+    x_displ = np.minimum(((np.sin(3 * np.pi * t + np.arcsin(-0.95)) + 0.95) / 8) * y / flap_tip_y, 0.5 * y / flap_tip_y)
```
BenjaminRodenberg:
```suggestion
    # first, get displacement independent of x, only dependent on y and t
    # initially, flap's displacement is 0
    max_x_displacement = 0.5
    x_displ = np.minimum(((np.sin(3 * np.pi * t + np.arcsin(-0.95)) + 0.95) / 8) * y / flap_tip_y, max_x_displacement * y / flap_tip_y)
```

I would generally try to avoid raw numbers and rather use descriptive names. This often allows you to completely save some comments.

What does the `0.95`, the `3 * np.pi` and the `/8` do? Generally, I think if the formula you are using here works we can keep it. But a brief comment about the general rationale behind the `arcsin` and `sin` would be helpful.

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,78 @@
...
+
+    displ = np.zeros((len(x), 2))
```
BenjaminRodenberg:
```suggestion
    dimensions = 2
    displ = np.zeros((len(x), dimensions))
```
Again: Makes it more explicit that this only works in 2D.

perpendicular-flap/solid-fake/fake.py
```diff
@@ -0,0 +1,78 @@
...
+y_top = y_bottom + H  # top of the flap
+
+n = 24  # Number of vertices per side
+t = 0
+
+vertices = np.zeros((2 * n, dimensions))
+vertices[:n, 1] = np.linspace(y_bottom, y_top, n)
+vertices[n:, 1] = np.linspace(y_bottom, y_top, n)
+vertices[:n, 0] = x_left
+vertices[n:, 0] = x_right
```
BenjaminRodenberg:
```suggestion
# define n vertices for left / upstream side of flap
vertices[:n, 1] = np.linspace(y_bottom, y_top, n)
vertices[:n, 0] = x_left
# define n vertices for right / downstream side of flap
vertices[n:, 1] = np.linspace(y_bottom, y_top, n)
vertices[n:, 0] = x_right
```

perpendicular-flap/solid-fake/fake.py
BenjaminRodenberg: I think this was somehow cached. Cannot find this piece of code anymore and is not relevant.

perpendicular-flap/solid-fake/fake.py
NiklasVin: I added some comments that hopefully clarify the idea behind the formula

BenjaminRodenberg (approved):
Thank for the changes! Looks good now. Let's get this merged :tada:

PR #544, opened 2024-06-26
volume-coupled-flow: add y and z components in source term

### LINKED ISSUE #542: volume-coupled-flow: Source term incomplete
https://github.com/precice/tutorials/blob/0a05e29d021290611eec04c51b0611dc31dabc21/volume-coupled-flow/fluid-openfoam/constant/fvOptions#L29

This probably also needs y and z.

MakisH:
Closes #542.

At the moment, the source term only considers the x-component of the read velocity in the volume coupling, even though we do exchange 2D velocities. Even though this example is completely artifical, reading only the `x` component seems to be unmotivated, and it already triggered a related question by @fsimonis when using this tutorial as a starting point.

Even though we exchange 2D data in preCICE, this is copying between OpenFOAM objects, which are both 3D structures.

@tirgendetwas do I miss something here, or was this simply something that slipped the review?

Results currently:

Screenshot from 2024-06-26 16-26-28

Results with the change:

Screenshot from 2024-06-26 16-25-22

Checklist:


MakisH:
Ready to review (maybe @davidscn?).
This tutorial remains a very arbitrary/constructed or even plain wrong example just to demonstrate the concept (we knew that since the design phase), but I think that this PR still fixes a bug.

davidscn:
I think the changes here require rather a conceptual review than a technical one and I am not familiar with the tutorial, what it is supposed to showcase or what the coupling is supposed to comprise. So, let me remove my review..

uekerman (approved):
The change makes sense to me, the tutorial not completely :grin:

Issue #547, opened 2024-07-02
Missing `requirements.txt` and `venv` in some Python-based examples


MakisH:
A generalization of #190.

Everything Nutils-based is already covered. We need to adapt our FEniCS and all Python-only examples.

BenjaminRodenberg:
A list of open ToDos:

- [x] https://github.com/precice/tutorials/tree/develop/channel-transport-reaction/chemical-fenics
- [x] https://github.com/precice/tutorials/tree/develop/channel-transport-reaction/fluid-fenics
- [x] https://github.com/precice/tutorials/tree/develop/elastic-tube-1d/fluid-python https://github.com/precice/tutorials/pull/564 @MakisH
- [x] https://github.com/precice/tutorials/tree/develop/elastic-tube-1d/solid-python https://github.com/precice/tutorials/pull/564 @MakisH
- [x] https://github.com/precice/tutorials/tree/develop/elastic-tube-3d/solid-fenics
- [x] https://github.com/precice/tutorials/tree/develop/flow-around-controlled-moving-cylinder/solid-python
- [x] https://github.com/precice/tutorials/tree/develop/flow-around-controlled-moving-cylinder/controller-fmi **fmi** @uekerman 
- [x] https://github.com/precice/tutorials/tree/develop/flow-over-heated-plate/solid-fenics
- [x] https://github.com/precice/tutorials/tree/develop/oscillator/mass-left-fmi **fmi** @uekerman 
- [x] https://github.com/precice/tutorials/tree/develop/oscillator/mass-right-fmi **fmi** @uekerman 
- [x] https://github.com/precice/tutorials/pull/556
    - [x] https://github.com/precice/tutorials/tree/develop/oscillator/mass-right-python
    - [x] https://github.com/precice/tutorials/tree/develop/oscillator/mass-left-python
    - [x] https://github.com/precice/tutorials/tree/develop/oscillator-overlap/mass-left-python
    - [x] https://github.com/precice/tutorials/tree/develop/oscillator-overlap/mass-right-python
- [x] https://github.com/precice/tutorials/tree/develop/partitioned-heat-conduction-complex/dirichlet-fenics
- [x] https://github.com/precice/tutorials/tree/develop/partitioned-heat-conduction-complex/neumann-fenics
- [x] https://github.com/precice/tutorials/tree/develop/partitioned-heat-conduction-overlap/left-fenics
[...]

PR #548, opened 2024-07-02
Set NumPy version requirement for Nutils cases

MakisH:
Our Nutils cases implicitly depend on NumPy 1. NumPy 2 was recently released, which breaks such installations. This adds a `numpy >1, <2` in all `requirements.txt` that also specify Nutils.

This brings up the issue that we should add `requirements.txt` for all other Python examples as well (in a separate PR). See #547

Checklist:

- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`. -> N/A

### DIFF
--- a/channel-transport/fluid-nutils/requirements.txt
+++ b/channel-transport/fluid-nutils/requirements.txt
@@ -1,2 +1,3 @@
 nutils==7
+numpy >1, <2
 pyprecice==3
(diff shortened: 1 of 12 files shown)


IshaanDesai (commented):
Looks good, but a cleaner solution would be to restrict the NumPy version in the Python bindings dependencies, and release a new version of the bindings. Adding the line in every tutorial is error prone and will increase maintainability. 

*pip would still update NumPy, even if we restrict it to a lower version at the bindings stage*

IshaanDesai (approved):

PR #551, opened 2024-07-11
Add channel-transport transport OpenFOAM

fsimonis:
This PR adds an OpenFOAM variant of the transport participant of the channel-transport tutorial.

The case uses a modified version of the scalarTransportFoam application. Main difference is that we assume `U` to change every timestep. Therefore we need to recompute $\phi$ every timestep.

@MakisH Can you give this PR a thorough review to make it as easy to use as possible?

Result:

[Screencast_20251210_145936.webm](https://github.com/user-attachments/assets/3fbc4817-9426-4cdb-8ce4-5061975f3061)

Checklist:

_edit: updated to the working state_


channel-transport/README.md
```diff
@@ -41,6 +41,8 @@ Transport participant:
 
 * Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0.
 
+* OpenFOAM. This case uses a modified version of scalarTransportFoam, which recomputes `phi` after the preCICE adapter reads the velocity field. Build it with `wmake`. Read more details in the [OpenFOAM adapter](https://precice.org/adapter-openfoam-overview.html). Note that T accumulates in front of the outlet as this case conserves overall T.
```
MakisH, line 44: Where is the solver? I would expect it here in a `solver-openfoam` or `solver-<more_specific_name>` directory.

MakisH:
```suggestion
* OpenFOAM. This case uses a modified version of scalarTransportFoam, which recomputes `phi` after the preCICE adapter reads the velocity field. Build it with `wmake`. Read more details in the [OpenFOAM adapter](https://precice.org/adapter-openfoam-overview.html). Note that `T` accumulates in front of the outlet as this case conserves overall `T`.
```

For consistency.

However: is this really intended, or a modelling issue?

channel-transport/solver-openfoam/.clang-format
MakisH, line 1: I start feeling very uneasy that we duplicate this file everywhere.
Opened an issue: https://github.com/precice/tutorials/issues/559

channel-transport/transport-openfoam/.gitignore
MakisH, line 1: We should have these in the central gitignore: https://github.com/precice/tutorials/blob/develop/.gitignore

The `0/` is already covered, the rest could go there as-is (strange we don't have them already).

channel-transport/solver-openfoam/Make/files
MakisH, line 1: You are mixing the solver with the case files. This is not a good idea.
I would expect the solver in a separate `solver-openfoam` or `solver-dynamicScalarTransportFoam` directory.

Why this name, btw?

channel-transport/solver-openfoam/dynamicScalarTransportFoam.C
MakisH, line 1: Can you explain the exact history of this file?
What did you get from where, what did you add yourself?

channel-transport/transport-openfoam/system/setExprFieldsDict
MakisH, line 1: A comment here would be nice.
I assume you are initializing a blob of `T` at (1,1).

channel-transport/transport-openfoam/system/fvSchemes
```diff
@@ -0,0 +1,50 @@
+/*--------------------------------*- C++ -*----------------------------------*\
+| =========                 |                                                 |
+| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
+|  \\    /   O peration     | Version:  v2312                                 |
+|   \\  /    A nd           | Website:  www.openfoam.com                      |
+|    \\/     M anipulation  |                                                 |
+\*---------------------------------------------------------------------------*/
```
MakisH: These headers are actually misleading, since we anyway have changed these configuration files a lot compared to what you might have started from.

(referring to this file and the `fvSolution`)

channel-transport/transport-openfoam/system/fvSchemes
```diff
@@ -0,0 +1,50 @@
...
+
+gradSchemes
+{
+    default         Gauss linear;
+}
+
+divSchemes
+{
+    default         none;
+    div(phi,T)      Gauss linearUpwind grad(T);
```
MakisH: Does tuning this maybe help with the strange values at the outlet?

MakisH (commented):
Besides the fact that solver and config are in the same directory, it looks good overall. I have not yet tried to run it.

channel-transport/README.md
fsimonis: It's a modelling issue that I couldn't find a solution for. I also ran out of people to ask for help

channel-transport/solver-openfoam/.clang-format
fsimonis: There is no uniform style in the tutorials repo. I don't see an easy solution for this.

channel-transport/transport-openfoam/.gitignore
fsimonis: It's the other way around.

`0/` is tracked and `0.*/` ignored by default. In my case I generate `0/` from `0.orig/`, which is the exact opposite of what the gitignore defines.

https://github.com/precice/tutorials/blob/3f253cc6323cb89559a98816ab9c34fb4f7ec060/.gitignore#L41-L43

channel-transport/solver-openfoam/Make/files
fsimonis: I wasn't aware of that. Thanks.

`scalarTransportFoam` with support for a dynamic velocity field = `dynamicScalarTransportFoam`.
Naming suggestions are welcome.

channel-transport/solver-openfoam/dynamicScalarTransportFoam.C
fsimonis: Where would you expect to find this information? In the file itself or in a README in the directory of the solver?

channel-transport/transport-openfoam/system/fvSchemes
fsimonis: How do you suggest me to handle this? (also in the solver source)

channel-transport/transport-openfoam/system/fvSchemes
fsimonis: Feel free to try. I wasn't able to change the behaviour.

channel-transport/README.md
MakisH: We should better understand this, and either add details, or fix it. I think right now it just looks wrong.

channel-transport/transport-openfoam/.gitignore
MakisH: Interestingly, we don't have a `.gitignore` in https://github.com/precice/tutorials/tree/develop/partitioned-pipe-two-phase, which should have the same issue.

But again, the `Make/`-related rules can go into the central `.gitignore`, or at least moved to the solver directory.

channel-transport/solver-openfoam/dynamicScalarTransportFoam.C
MakisH: Just here first, to decide on what to do with this header.
But eventually in the file itself.

channel-transport/transport-openfoam/system/fvSchemes
MakisH: I am suggesting to remove the copyright header from the `fvSchemes` and `fvSolution`, because it is wrong.

channel-transport/transport-openfoam/system/fvSchemes
MakisH: We can eventually try this together, but I don't have the time to debug this at the moment. That's why I am giving a hint on what I think could help first.

channel-transport/solver-openfoam/.clang-format
MakisH: There is a `.clang-format` file in the root directory: https://github.com/precice/tutorials/blob/develop/.clang-format

Let's discuss in #559, but keep the file here for now.

channel-transport/solver-openfoam/Make/files
MakisH: The name makes sense, I guess.

channel-transport/transport-openfoam/system/fvSchemes
fsimonis: What about the actual solver sources?

MakisH: There we need it, since it is actually a small extension of OpenFOAM code.
But this is discussed in https://github.com/precice/tutorials/pull/551#discussion_r1736768309

MakisH:
@fsimonis this seems to be close to merging. Does this need more work from your side, or just a new review iteration?

edit: Actually, no. There is still a blocking modeling issue.

MakisH:
Just finished a co-working session with @fsimonis. I pushed some new commits, which make the custom solver work with dynamic meshes and adjusts the case to apply dynamic mesh refinement with `dynamicRefineFvMesh`, based on the magnitude of the gradient of the transported value.

The settings are in `constant/dynamicMeshDict` and are currently rather arbitrary. Next step would be to refine these. We worked with an uncoupled scalar transport case based on the pitzDaily tutorial. Here is the draft (also with arbitrary settings):

[adaptivePitzDaily.tar.gz](https://github.com/user-attachments/files/23775520/adaptivePitzDaily.tar.gz)

Run with: `rm -rf 0.* && blockMesh && dynamicScalarTransportFoam`.

The next step would be to further understand and adjust the mesh refinement settings.

The coupled case currently fails, but this is expected since the adapter does not yet support remeshing (https://github.com/precice/openfoam-adapter/issues/382). I can start an implementation of that.

The accumulation of `T` at the outlet remains. Making the channel longer and refining the mesh showed that the accumulation is really on the last 2-3 layers of the mesh. Changing the `system/fvSchemes` strongly affected the behavior (to the worse):

```diff
divSchemes
{
    default         none;
-    div(phi,T)      Gauss linearUpwind grad(T);
+    div(phi,T)      Gauss linear grad(T);
}
```

However, the numerics and the boundary conditions are the same as in the pitzDaily case, so this might be unrelated. I would look next if the changes from the standard `scalarTransportFoam` have any effect (but I would expect not).

Some URLs for reference:

- https://precice.org/couple-your-code-moving-or-changing-meshes.html
- https://api.openfoam.com/2506/classFoam_1_1dynamicRefineFvMesh.html#details
- https://openfoamwiki.net/index.php/Parameter_Definitions_-_dynamicRefineFvMesh

fsimonis:
I got the non-AMR version running. I moved the AMR specific changes to #689 and will revert this PR to the non-AMR version.
Let's then first merge it and continue from there.

channel-transport/README.md
fsimonis: Solved

channel-transport/transport-openfoam/system/fvSchemes
fsimonis: I replaced the header with a link and description. Does this clarify the matter?

channel-transport/transport-openfoam/system/fvSchemes
fsimonis: Solved

channel-transport/transport-openfoam/system/setExprFieldsDict
fsimonis: Done

fsimonis:
@MakisH I think I implemented all suggestions. Could you review again?

channel-transport/solver-openfoam/dynamicScalarTransportFoam.C
```diff
@@ -0,0 +1,62 @@
...
+
+int main(int argc, char *argv[])
+{
+  argList::addNote(
+      "Dynamic scalar transport equation solver.");
+
+#include "addCheckCaseOptions.H"
+#include "createMesh.H"
+#include "createTime.H"
+#include "setRootCaseLists.H"
```
MakisH: In this order, the code doesn't compile. Last good commit is 8ae94048612ac9b126788ab88f11cd23f18b8047.

Correct order:


```suggestion
#include "setRootCaseLists.H"
#include "createTime.H"
#include "createMesh.H"
```

We need to update the `.clang-format` to disable reordering. Already discussing this in https://github.com/precice/tutorials/issues/559.

This is the only instance of this issue.

channel-transport/transport-openfoam/system/blockMeshDict
MakisH, line 1: I see the trick with the grading, it is already much better, but I am not yet sure that the case is correctly configured. The accumulation at the outlet is gone, but if I just change the mesh (I apply the same `blockMeshDict` from `fluid-openfoam`), I still see some accumulation at the outlet and an unphysical jump on the second-last layer:

https://github.com/user-attachments/assets/78d9224c-1692-46de-b3ef-a894f3ac38e5

While I could accept that the origin of the issue is related to the mesh, I would like to better understand it.

## Nutils

The combination `fluid-openfoam`/`transport-nutils` does not show this issue (solution available on mesh nodes only, Paraview interpolates):

https://github.com/user-attachments/assets/dd47b864-9eba-4061-ac9b-044016d84e56

## Fine grid

Even the fine grid case still shows some temporary accumulation near the center of the outlet when `T` first reaches the outlet:

https://github.com/user-attachments/assets/22672f8d-0e03-4ed7-a558-0435b46e7948

If I restrict the value range to 0-0.2, some cells near the corner also look a bit strange, as the values at the last layer of cells are getting smaller than the values in the layer before:

https://github.com/user-attachments/assets/4d03610a-8ab5-42aa-ae74-215cc3c26be8

## Monolithic

To check the monolithic case, I removed the function object from the `controlDict`, I applied a similar initial condition as in the `fluid-openfoam` in `U`: `internalField   uniform (10 0 0);`, and I used the same `blockMeshDict` as in the `fluid-openfoam`.

I observe the same "jump" before the obstacle (the accumulation is expected, since the flow field is constant and fully horizontal), but no issue with the outlet:

https://github.com/user-attachments/assets/28783987-e7de-42e1-b9f1-4fdd1802eafa

Same with the standard `scalarTransportFoam`:

https://github.com/user-attachments/assets/2fd07b82-5a75-43e3-988d-18c20edab385

## Conclusions

From this, I converge to:

- The issue at the outlet must be related to the combination of the volume coupling and the update of the outlet boundary, since it does not appear in the monolithic cases. But the issue is observed on the `T`, not on the `U`, which is the one being read. I am confused.
- The integration scheme of `T` could still be improved, as the monolithic simulation shows discontinuities. 

Edit: Discussion continues in the main thread: https://github.com/precice/tutorials/pull/551#issuecomment-4788119511

MakisH (commented):
I applied a few more changes directly, and in terms of usability and integration to the rest of the tutorials, the case is ready. It still needs reference results for the system tests (I can add them).

I still have doubts about the outlet conditions and the space integration of `T`. But I am also still confused.

Fujikawas:
I tried setting `div(phi,T)      bounded Gauss upwind;` in `fvSchemes` and it can remove the false accumulation of the scalar at the outlet both with BC  `advective` or `zeroGradient` for outlet.

MakisH:
Thank you, @Fujikawas! I can confirm that this scheme works just fine. Here is what I get with the coarse mesh and `zeroGradient`:

https://github.com/user-attachments/assets/59d2ec74-69ea-4d64-8c8a-b32315fa8818

I will then use this scheme, the coarse mesh, the simpler `zeroGradient` condition, I will remove the custom meshing script.

MakisH (approved):
This looks good to me now. Since I changed the mesh, @fsimonis could you please confirm that this is fine for you?

Since this is coming from a fork, I will add the reference results directly on develop.

fsimonis:
Such a simple solution and we spent hours debugging this 😂

Having the simpler mesh is actually better as it makes the base of the AMR version more realistic.

@MakisH feel free to merge

Issue #552, opened 2024-07-11
Specify default build arguments


MakisH:
We currently cannot really skip providing all the versions of the involved components as `--buid_args`. We should have a better mechanism to set defaults.

(To add more details later)

PR #553, opened 2024-07-16
Fix typo in components.yaml

uekerman:
There are likely more place where this needs to be updated:

- [ ] https://github.com/precice/precice/pull/2052
- [ ] add more


MakisH:
That's a known one, we were mainly avoiding touching it because it touched many other files already (pretty much every workflow specifying a CalculiX version). But good time to fix it, before we add more workflows.

Should not need to update many files after addressing #552.

MakisH (approved):
Let's merge this and observe what fails (and fix that as well).

PR #554, opened 2024-08-04
Fix crash for FEniCS-OpenFOAM perpendicular flap example with subcycling 

NiklasVin:
Addresses the crashes when using subcycling as described in https://github.com/precice/fenics-adapter/pull/172.

The current implementation of `solid-fenics.py` operates directly on the dof vectors of `u`, `a` and `v` at one point. This also changes the values stored in the checkpoint which ultimately leads to the crash described in the PR above.
Updating values of a function object via `assign()`, does not change the values stored in a checkpoint, since the old dof vector of the function to be updated is overwritten by calling `assign()` .

### DIFF
--- a/perpendicular-flap/solid-fenics/solid.py
+++ b/perpendicular-flap/solid-fenics/solid.py
@@ -77,7 +77,7 @@ precice = Adapter(adapter_config_filename="precice-adapter-config-fsi-s.json")
 precice.initialize(coupling_boundary, read_function_space=V, write_object=V, fixed_boundary=fixed_boundary)
 
 precice_dt = precice.get_max_time_step_size()
-fenics_dt = precice_dt / 10  # if fenics_dt == precice_dt, no subcycling is applied
+fenics_dt = precice_dt  # if fenics_dt == precice_dt, no subcycling is applied
 # n_substeps = 5  # number of substeps per window
 # fenics_dt = precice_dt / n_substeps  # if fenics_dt < precice_dt, subcycling is applied
 dt = Constant(np.min([precice_dt, fenics_dt]))
--- a/perpendicular-flap/solid-fenics/solid.py
+++ b/perpendicular-flap/solid-fenics/solid.py
@@ -77,7 +77,7 @@ precice = Adapter(adapter_config_filename="precice-adapter-config-fsi-s.json")
 precice.initialize(coupling_boundary, read_function_space=V, write_object=V, fixed_boundary=fixed_boundary)
 
 precice_dt = precice.get_max_time_step_size()
-fenics_dt = precice_dt  # if fenics_dt == precice_dt, no subcycling is applied
+fenics_dt = precice_dt / 10  # if fenics_dt == precice_dt, no subcycling is applied
 # n_substeps = 5  # number of substeps per window
 # fenics_dt = precice_dt / n_substeps  # if fenics_dt < precice_dt, subcycling is applied
 dt = Constant(np.min([precice_dt, fenics_dt]))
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg:
I think this clearly improves the readability of `solid.py`.

It also hints at a problem with the FEniCS adapter: From my perspective it should not be possible to alter a checkpoint -- intentionally or by accident -- as soon as it is handed over to the FEniCS adapter. Meaning: As far as I understand this PR cures the symptom but we can only solve the original problem with https://github.com/precice/fenics-adapter/pull/172.

NiklasVin:
Yes, it is true that it only cures the symptom. With this PR, I also wanted to show, that the current issue is easily avoidable from the user's side. And since I am not working that long with FEniCS, I am not sure how common it is to directly change the values by directly accessing `vector()`. If it is more of an exception, I wouldn't be that content with the deepcopy solution, since it is, in the rest of the cases, a bit pointless, but I don't want to be a nitpicker. So yes, the deepcopy solves the original problem for all cases.

BenjaminRodenberg (approved):
Looks good to me and definitely improves readability. Let's merge it.

Let's discuss the `deepcopy` issue directly in https://github.com/precice/fenics-adapter/pull/172 because it is now not really related to this tutorial anymore.

perpendicular-flap/solid-fenics/solid.py
```diff
@@ -158,16 +158,14 @@ def update_v(a, u_old, v_old, a_old, ufl=True):
...
-    v_vec = update_v(a_vec, u0_vec, v0_vec, a0_vec, ufl=False)
+    a_new = update_a(u, u_old, v_old, a_old)
+    v_new = update_v(u, u_old, v_old, a_old)
 
-    # assign u->u_old
-    v_old.vector()[:], a_old.vector()[:] = v_vec, a_vec
-    u_old.vector()[:] = u.vector()
+    # update values
+    a_old.assign(project(a_new, V))
+    v_old.assign(project(v_new, V))
```
BenjaminRodenberg, line 167: Just discovered one more thing: Do we really need this projection? If yes, why?

I'm just wondering because no projection was needed in `v_old.vector()[:], a_old.vector()[:] = v_vec, a_vec`. But maybe this also happened somehow implicitly...

BenjaminRodenberg: I played around with the following example and have the impression you do not really need the projection:
```
from fenics import Function, FunctionSpace, UnitIntervalMesh

mesh = UnitIntervalMesh(5)
V = FunctionSpace(mesh,'P',1)
u = Function(V)
v = Function(V)
u.vector()[:]=[1,2,3,4,5,6]  # array([1., 2., 3., 4., 5., 6.])
u_sum = u+u
u_sum.vector()[:]  # AttributeError: 'Sum' object has no attribute 'vector'
v.assign(u_sum)
v.vector()[:]  # array([ 2.,  4.,  6.,  8., 10., 12.])
```
Conclusion: The assign function also accepts an UFL expression (`u_sum`). The documentation (`help(Function.assign)`) also states that the following arguments are accepted:
```
A Function or a linear combination of Functions. If a linear
            combination is passed all Functions need to be in the same
            FunctionSpaces.
```
This should not be a problem since we are only dealing with linear combinations of functions from `V` (Multiplying by a constant is also allowed).

If I pass, for example, `assign(u*u)` I also directly get an error: `RuntimeError: Expected only linear combinations of Functions in the same FunctionSpaces`. If you do not see this error it should be safe to remove the projection and continue.

NiklasVin: I think the variables of type `Constant` are the issue here. I tested different scenarios and two used e.g. in `update_a()` are problematic, namely a Constant to the power of some value (as used in `update_a()`) - removing the `dt**2` and using `dt * dt` should fix the problem- and using a `Constant` variable as a divisor seem to be the issue.
I tried it with your code from above, introducing a single Constant variable and multiplying/dividing the corresponding variable to the second summand of u_sum, so e.g.:
```diff
...
- u_sum = u+u
+ t = Constant(2)
+ u_sum = u+u*t**2 # apparently no linear combination anymore??
...
```
or:
```diff
...
- u_sum = u+u
+ t = Constant(2)
+ u_sum = u+u/t # AttributeError: 'Constant' object has no attribute 'value_rank'
...
```

BenjaminRodenberg: What about `u+u*t*t`? And `u+u*1/t`? But if we cannot get it working with the kind of UFL expressions we are using here let's just use the (more general) projection. I'm just wondering why it worked without a projection before...

NiklasVin: `1/t` also does not work. You could use a workaround that looks similar to `t_inv = Constant(1/t(0))` and then multiply by `t_inv` but I think this is an ugly solution.
> I'm just wondering why it worked without a projection before...

If you mean by _before_ the already existing implementation, then it makes sense to not use the projection, since the values of the dof vector of the functions `a` and `v` are directly modified. In this PR, in `update_a` and `update_v`, we create a new function of some function space, instead of just modifying the already exisiting function of function space `V`.

BenjaminRodenberg: > `1/t` also does not work. You could use a workaround that looks similar to `t_inv = Constant(1/t(0))` and then multiply by `t_inv` but I think this is an ugly solution.

I totally agree. Let's go with the projection.

PR #556, opened 2024-08-19
Add missing `requirements.txt` to oscillator and oscillator-overlap tutorial

NiklasVin:
This PR adds the `requirements.txt` files for both python solvers of this tutorial, that were missing.

### DIFF
--- a/oscillator/mass-left-python/requirements.txt
+++ b/oscillator/mass-left-python/requirements.txt
@@ -1,0 +1,2 @@
+numpy >1, <2
+pyprecice~=3.0
--- a/oscillator/mass-left-python/requirements.txt
+++ b/oscillator/mass-left-python/requirements.txt
@@ -1,2 +1,3 @@
 numpy >1, <2
 pyprecice~=3.0
+scipy
--- a/oscillator/mass-left-python/run.sh
+++ b/oscillator/mass-left-python/run.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -e -u
 
+python3 -m venv .venv
+. .venv/bin/activate
+pip install -r requirements.txt
+
 . ../../tools/log.sh
 exec > >(tee --append "$LOGFILE") 2>&1
 
--- a/oscillator/mass-right-python/requirements.txt
+++ b/oscillator/mass-right-python/requirements.txt
@@ -1,0 +1,2 @@
+numpy >1, <2
+pyprecice~=3.0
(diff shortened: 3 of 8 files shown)


oscillator/mass-left-python/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy >1, <2
+pyprecice~=3.0
```
MakisH, line 2: We still need scipy in both participants:

```
  File "/home/gc/repos/precice/tutorials/oscillator/mass-left-python/../solver-python/oscillator.py", line 12, in <module>
    import timeSteppers
  File "/home/gc/repos/precice/tutorials/oscillator/solver-python/timeSteppers.py", line 4, in <module>
    import scipy as sp
ModuleNotFoundError: No module named 'scipy'
```

oscillator/mass-left-python/run.sh
MakisH, line 1: In this case, we have the solver in `../solver-python`.
Would it make sense to create the `requirements.txt` and the `venv` there, to avoid duplication?

MakisH (commented):
Thank you for spotting this!
I think this case is a bit special, since we have the solver in a dedicated directory.

There might be more similar cases to update accordingly.

BenjaminRodenberg (commented):
~~Maybe consider adding `oscillator-overlap` in this PR as well? I think everything should be pretty identical to `oscillator`.~~

~~We could also do this in a new PR, but if @NiklasVin is doing another round of testing before merging then I think doing `oscillator-overlap` in the same sweep would be a good idea.~~

Took care of this since I was running the case anyways. Will merge this PR now.

oscillator/mass-left-python/requirements.txt
BenjaminRodenberg, line 2:
```suggestion
numpy >1, <2
pyprecice~=3.0
scipy
```

oscillator/mass-right-python/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy >1, <2
+pyprecice~=3.0
```
BenjaminRodenberg, line 2:
```suggestion
numpy >1, <2
pyprecice~=3.0
scipy
```

BenjaminRodenberg (approved):
I did a quick test run of this case. After adding scipy everything works.

oscillator/mass-left-python/run.sh
BenjaminRodenberg: I think that's a good point that could also save us **a lot** of implementation & maintainment work in https://github.com/precice/tutorials/issues/547.

Let's do the following here: Let's get this PR merged with the structure as it is. Moving `requirements.txt` and the `venv` into the solver is a general thing that we should discuss in #574 and then solve consistently (maybe even for the nutils cases). I'll add a comment there.

PR #557, opened 2024-08-22
pySDC & FEniCS solver for partitioned-heat-equation

TobiasEppacher:
- Addition of a solver program for the `partitioned-heat-equation' tutorial, based on pySDC


partitioned-heat-conduction/dirichlet-pySDC-fenics/precice-adapter-config.json
BenjaminRodenberg, line 1: Is this JSON file identical to the one provided in `dirichlet-fenics`?

partitioned-heat-conduction/dirichlet-pySDC-fenics/clean.sh
BenjaminRodenberg, line 1: Is this `clean.sh` file identical to the one provided in `dirichlet-fenics`?

partitioned-heat-conduction/dirichlet-pySDC-fenics/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e -u
+
+. ../../tools/log.sh
+exec > >(tee --append "$LOGFILE") 2>&1
+
+python3 ../solver-fenics/heat_pySDC.py Dirichlet
```
BenjaminRodenberg: The only difference to the `run.sh` file in `dirichlet-fenics` is that you use `../solver-fenics/heat_pySDC.py` instead of `../solver-fenics/heat.py`. Correct?

I think we could actually live without `dirichlet-pySDC-fenics` and put everything into `dirichlet-fenics`. Maybe you could add an optional argument to the `run.sh` file such that `./run.sh` runs the "normal" case and `./run.sh sdc` would use `heat_pySDC.py` instead of `heat.py`.

@MakisH Does this approach fit into our framework?

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+W = V_g.sub(0).collapse()
+
+
+##################################################################
+# Problem definition and preCICE initialization
+##################################################################
+
+# Time step size
+# Should be integer fraction of the used time window size
+precice_time_window_size = 0.1
```
BenjaminRodenberg: This is something that we should get from the precice config. Please use the following:
```suggestion
precice_time_window_size = precice.get_max_time_step_size()
```
Note: At the beginning of the window the max step size equals the window size.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+precice_waveform_deg = 1
+pySDC_dt = precice_time_window_size/precice_waveform_deg
```
BenjaminRodenberg: Why do you need the degree? There is no API function to get the degree from preCICE.

I think it should be the other way around: Your time stepping scheme has certain requirements and then the user should be able to configure preCICE in a way that fulfills the requirements.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+description['level_params'] = level_params
+description['step_params'] = step_params
+
+# Controller for time stepping
+controller = controller_nonMPI(num_procs=1, controller_params=controller_params, description=description)
+
+# Reference to problem class for easy access to exact solution
+P = controller.MS[0].levels[0].prob
+
+##################################################################
```
BenjaminRodenberg: I'm not a big fan of this "section"-style. Is it possible to put this into a function with an descriptive name? This also makes it clearer which configuration parameters come from the outside and which ones are local.

Example:

```
def create_pysdc_controller(dt, ...):
    ...
    # REVIEW COMMENT: These are some local parameters that we can hide from the outside context
    # initialize sweeper parameters
    sweeper_params = dict()
    sweeper_params['quad_type'] = 'LOBATTO'
    sweeper_params['num_nodes'] = 4
    ...
    return controller

... # usual stuff before coupling loop

contoller = create_pysdc_controller(pySDC_dt, ...)

... # enter coupling loop
```

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+description['sweeper_class'] = imex_1st_order_mass
+description['sweeper_params'] = sweeper_params
+description['level_params'] = level_params
+description['step_params'] = step_params
+
+# Controller for time stepping
+controller = controller_nonMPI(num_procs=1, controller_params=controller_params, description=description)
+
+# Reference to problem class for easy access to exact solution
+P = controller.MS[0].levels[0].prob
```
BenjaminRodenberg:
```suggestion
```
Is this actually needed?

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+            error_out << error_pointwise
+
+    precice_dt = precice.get_max_time_step_size()
+    dt.assign(np.min([pySDC_dt, precice_dt]))
+
+    # Update start time of the current preCICE time step within
+    # the problem class.
+    # This is necessary, that the coupling expression update in the problem class
+    # is executed correctly.
+    P.set_t_start(t)
```
BenjaminRodenberg: Ok. Here `P` is used. I would actually consider writing the following instead:

```suggestion
    controller.MS[0].levels[0].prob.set_t_start(t)
```

Yes. This is a bit more lengthy but it saves you one indirection and clarifies the connection to the controller.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Retrieve the result at the end of the timestep from pySDC
+    # Possible statistics generated by pySDC can be accessed via the 'stats' variable if needed
+    uend, stats = controller.run(u_n, t0=t, Tend=t + float(dt))
+    
+    # Update the buffer with the new solution values
+    u_np1 = uend.values
```
BenjaminRodenberg: You could consider putting these lines into a function `u_np1 = do_time_step(controller, u_n, t, dt)`. This is the only point where you interact with pySDC? Really nice that the interface is so clean.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Update start time of the current preCICE time step within
+    # the problem class.
+    # This is necessary, that the coupling expression update in the problem class
+    # is executed correctly.
+    P.set_t_start(t)
+
+    # Retrieve the result at the end of the timestep from pySDC
+    # Possible statistics generated by pySDC can be accessed via the 'stats' variable if needed
+    uend, stats = controller.run(u_n, t0=t, Tend=t + float(dt))
```
BenjaminRodenberg:
```suggestion
    uend, _ = controller.run(u_n, t0=t, Tend=t + float(dt))
```
It's convention in python to replace unused variables with a `_`. You do not use `stats`, do you?

Even better would be `uend = controller.run(...)` but I guess the function always returns two values. In this case ignoring the `stats` and only keeping the `uend` is cleaner.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+# Manufactured solution parameters
+alpha = 3  # parameter alpha
+beta = 1.2  # parameter beta
+
+
+# create sympy expression of manufactured solution
+x_sp, y_sp, t_sp = sp.symbols(['x[0]', 'x[1]', 't'])
+
+temporal_deg = 1  # temporal degree of the manufactured solution
+u_D_sp = 1 + x_sp * x_sp + alpha * y_sp * y_sp + beta * (t_sp ** temporal_deg)
```
BenjaminRodenberg, line 183: This, for example, is one difference to `heat.py`, but integrating this feature into `heat.py` is not difficult. Again: We could add an option parameter `heat.py --polynomial-degree=123` that defaults to `1`.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+
+from pySDC.core.Problem import ptype
+from pySDC.implementations.datatype_classes.fenics_mesh import fenics_mesh, rhs_fenics_mesh
+
+
+class fenics_heat_2d(ptype):
+    dtype_u = fenics_mesh
+    dtype_f = rhs_fenics_mesh
+    
+    def __init__(self, mesh, function_space, forcing_term_expr, solution_expr, coupling_boundary, remaining_boundary, coupling_expr, precice_ref, participant_name):
```
BenjaminRodenberg: Can you add a docstring here and explain the arguments? See https://peps.python.org/pep-0257/#multi-line-docstrings.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+        # define the homogeneous Dirichlet boundary for residual correction
+        def FullBoundary(x, on_boundary):
+            return on_boundary
+        self.homogenousBC = df.DirichletBC(self.V, df.Constant(0), FullBoundary)
+        
+        
+    def solve_system(self, rhs, factor, u0, t):
+        r"""
+        Solves (M - factor \cdot A) \vec{u} = \vec{rhs}.
+        """
```
BenjaminRodenberg: I've never seen a docstring starting with `r"""` before. Is this the documentation style in pySDC? Is the latex code rendered on the website or similar?

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+        r"""
+            Returns the exact solution at time t.
+        """
+        self.solution_expr.t = t
+        
+        me = self.dtype_u(interpolate(self.solution_expr, self.V), val=self.V)
+        return me
+
+    def set_t_start(self, t_start):
+        self.t_start = t_start
```
BenjaminRodenberg: Hmm... not sure if you should actually use a function here. Why not simply call `controller.t_start = new_t_start` instead of this function in `heat_pySDC.py`? Currently, this function does not really add a lot of additional value (no additional logic, security checks, documentation...).

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+        return me
+
+    def u_exact(self, t):
+        r"""
+            Returns the exact solution at time t.
+        """
+        self.solution_expr.t = t
+        
+        me = self.dtype_u(interpolate(self.solution_expr, self.V), val=self.V)
+        return me
```
BenjaminRodenberg:
```suggestion
        return self.dtype_u(interpolate(self.solution_expr, self.V), val=self.V)
```
Would be shorter. Or any reasons for doing it differently?

PS: If this is convention in pySDC. Ok, since you are also doing it similarly in `apply_mass_matrix`

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+        f.expl = self.apply_mass_matrix(f.expl)
+
+        return f
+    
+    def fix_residual(self, res):
+        """
+            Applies homogeneous Dirichlet boundary conditions to the residual
+        """
+        self.homogenousBC.apply(res.values.vector())
+        return None
```
BenjaminRodenberg:
```suggestion
```
I think you don't need this.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
...
+        self.solution_expr.t = t
+        self.remainingBC.apply(T, b.values.vector())
+         
+        if self.participant_name == 'Dirichlet':     
+            # Coupling BC needs to point to correct time
+            dt = t - self.t_start
+            read_data = self.precice.read_data(dt)
+            self.precice.update_coupling_expression(self.coupling_expression, read_data)   
+            
+            self.couplingBC.apply(T, b.values.vector())
```
BenjaminRodenberg: From the perspective of coupling this is probably the most interesting part. Would be helpful to point to your thesis here or somewhere in the documentation. I guess you are explaining these few lines of code in detail in your thesis?

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,130 @@
+from fenics import *
+import dolfin as df
```
BenjaminRodenberg:
```suggestion
import fenics
```
Generally, I would suggest to avoid wildcard imports. See https://stackoverflow.com/questions/3615125/should-wildcard-import-be-avoided.

Do you really need to import `dolfin`? Usually all the relevant functions from `dolfin` are also in `fenics` (which is to my understanding the high-level wrapper).

What I usually do:
```
from fenics import function_that_I_use, another_function_that_I_use, ...
```
Then you also do not have to bother with all the `fenics.some_fenics_function_I_use_but_did_not_import(...)` below and just call `some_fenics_function_I_imported(...)`

BenjaminRodenberg (changes requested):
Hi @TobiasEppacher! Thanks for opening this PR. Can you please run `autopep8` on your code? Currently the CI is failing due to some formatting issues. Simply install `autopep8` and use the arguments given [here](https://github.com/precice/tutorials/blob/293c25b9595c4f7a055a228517664166d806c272/.pre-commit-config.yaml#L18).

I did not test the functionality yet but generally it already looks good. The way how you are implementing pySDC into the solver really looks good and is not invasive at all. With some refactoring we can make this even clearer.

partitioned-heat-conduction/dirichlet-pySDC-fenics/clean.sh
TobiasEppacher: Yes, it is identical

partitioned-heat-conduction/dirichlet-pySDC-fenics/precice-adapter-config.json
TobiasEppacher: Also this one is identical

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,346 @@
...
+controller_params['logger_level'] = 30
+
+# fill description dictionary for easy step instantiation
+description = dict()
+description['problem_class'] = fenics_heat_2d
+description['problem_params'] = problem_params
+description['sweeper_class'] = imex_1st_order_mass
+description['sweeper_params'] = sweeper_params
+description['level_params'] = level_params
+description['step_params'] = step_params
```
BenjaminRodenberg: I'm wondering a bit about how you are creating this dictionary. 

Instead of
 
```python
my_a = dict()
my_a['b'] = 42

description = dict()
description['a'] = my_a
```

I would suggest

```python
description = {
    'a': {
        'b':42
         }
}
```

This makes reading much easier, since you are going from top ('description') to bottom ('42') and not from in a zig-zag-fashion bottom  to top (first 'a', then 'b', then 'description').

For spacing and line breaks I would suggest to look for formatting suggestions. Some examples are here: https://stackoverflow.com/questions/44689546/how-to-print-out-a-dictionary-nicely-in-python

TobiasEppacher: Changed in new commit

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
TobiasEppacher: Fixed

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
TobiasEppacher: This was a remnant of the testing for my thesis. I removed it in the latest commit.

partitioned-heat-conduction/dirichlet-pySDC-fenics/run.sh
TobiasEppacher: I tried to unify the run-scripts for the default first-order FEniCS and my pySDC implementation. It would now work using an argument to the bash script which also includes Niklas Vinnitchenkos high-order FEniCS solver (`highorder` or `sdc`, everything else or no argument executes the first-order FEniCS solver)

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+
+precice_dt = precice.get_max_time_step_size()
+dt = Constant(0)
+dt.assign(np.min([pySDC_dt, precice_dt]))
+
+coupling_expression = precice.create_coupling_expression()
+
+
+controller, P = setup_problem(
+    participant_name=precice.get_participant_name(),
```
BenjaminRodenberg:
```suggestion
```
I think you do not need this: You are handing over `precice`. So you can also just call `precice.get_participant_name()` inside `setup_problem`. Same argumentation also holds for `__init__`

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+    w = TrialFunction(V_g)
+    v = TestFunction(V_g)
+
+    a = inner(w, v) * dx
+    L = inner(grad(u), v) * dx
+    solve(a == L, flux)
+
+
+def setup_problem(
+        participant_name,
```
BenjaminRodenberg:
```suggestion
```

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+        coupling_expression,
+        dt,
+        res_tol=1e-11,
+        maxiter=40,
+        quad_type='LOBATTO',
+        num_nodes=4,
+        logger_level=30):
+    """
+    Setup the problem for pySDC controller
+    :param participant_name: name of the participant
```
BenjaminRodenberg:
```suggestion
```

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+        'problem_params': {
+            'mesh': domain_mesh,
+            'function_space': V,
+            'coupling_boundary': coupling_boundary,
+            'remaining_boundary': remaining_boundary,
+            'solution_expr': u_D,
+            'forcing_term_expr': forcing_expr,
+            'precice_ref': precice,
+            'coupling_expr': coupling_expression,
+            'participant_name': participant_name
```
BenjaminRodenberg:
```suggestion
            'participant_name': precice.get_participant_name()
```
But maybe you do not even need this?

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+from pySDC.core.Problem import ptype
+from pySDC.implementations.datatype_classes.fenics_mesh import fenics_mesh, rhs_fenics_mesh
+
+
+class fenics_heat_2d(ptype):
+    dtype_u = fenics_mesh
+    dtype_f = rhs_fenics_mesh
+
+    def __init__(self, mesh, function_space, forcing_term_expr, solution_expr, coupling_boundary,
+                 remaining_boundary, coupling_expr, precice_ref, participant_name):
```
BenjaminRodenberg:
```suggestion
                 remaining_boundary, coupling_expr, precice_ref):
```

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+        Args:
+            mesh: FEniCS mesh object
+            function_space: FEniCS function space object
+            forcing_term_expr: FEniCS expression for the forcing term
+            solution_expr: FEniCS expression for the manufactured solution
+            coupling_boundary: FEniCS SubDomain object for the coupling boundary
+            remaining_boundary: FEniCS SubDomain object for the remaining boundary
+            coupling_expr: FEniCS expression for the coupling boundary condition
+            precice_ref: preCICE-FEniCS adapter object reference
+            participant_name: Name of the participant (Dirichlet or Neumann)
```
BenjaminRodenberg:
```suggestion
```

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+
+        # Set precice reference and coupling expression reference to update coupling boundary
+        # at every step within pySDC
+        self.precice = precice_ref
+        self.coupling_expression = coupling_expr
+        self.t_start = 0.0
+        self.participant_name = participant_name
```
BenjaminRodenberg:
```suggestion
        self.participant_name = self.precice.get_participant_name()
```
Or completely remove `self.participant_name` and just call `self.precice.get_participant_name()`? The operation is not costly and it also helps you to avoid bugs if you avoid storing redundant data.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Forcing term
+        self.forcing_term_expr = forcing_term_expr
+
+        # Solution expression for error comparison and as boundary condition
+        # on the non-coupling boundary
+        self.solution_expr = solution_expr
+
+        # Currently only for Dirichlet boundary, has to be changed for Neumann boundary
+        if self.participant_name == 'Dirichlet':
```
BenjaminRodenberg: See above

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        u = self.dtype_u(u0)
+        T = self.M - factor * self.K
+        b = self.dtype_u(rhs)
+
+        self.solution_expr.t = t
+        self.remainingBC.apply(T, b.values.vector())
+
+        # Update the coupling boundary condition for the Dirichlet participant
+        if self.participant_name == 'Dirichlet':
```
BenjaminRodenberg: See above

BenjaminRodenberg (changes requested):
I think you can simplify the code.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+precice_dt = precice.get_max_time_step_size()
+dt = Constant(0)
+dt.assign(np.min([pySDC_dt, precice_dt]))
+
+coupling_expression = precice.create_coupling_expression()
+
+
+controller, P = setup_problem(
+    participant_name=precice.get_participant_name(),
+    domain_mesh=domain_mesh,
```
BenjaminRodenberg: I believe you can also simplify your code here: The `mesh` is accessible via the `FunctionSpace` using `FunctionSpace.mesh()`.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
```diff
@@ -0,0 +1,372 @@
...
+    coupling_boundary=coupling_boundary,
+    remaining_boundary=remaining_boundary,
+    V=V,
+    u_D=u_D,
+    forcing_expr=forcing_expr,
+    precice=precice,
+    coupling_expression=coupling_expression,
```
BenjaminRodenberg: I think providing a list of `DirichletBC`s here would allow you to simplify your code quite a bit. Additionally, we could also do something along the following lines in the future:

If Dirichlet problem
```
setup_problem(
    ...
    dirichlet_bcs = [a vector of DirichletBC]
    neumann_bcs = [a vector of Expressions+SubDomains]
    ...
)
```
Both arguments would be optional. Additionally, pySDC should not have to worry whether a `DirichletBC` comes from a `coupling_expression` or not.

If you ignore the Neumann part:
```
setup_problem(
    ...
    dirichlet_bcs = [DirichletBC(V, coupling_expression, coupling_boundary),DirichletBC(V, u_D, remaining_boundary)]
    ...
)
```

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Solution expression for error comparison and as boundary condition
+        # on the non-coupling boundary
+        self.solution_expr = solution_expr
+
+        # Currently only for Dirichlet boundary, has to be changed for Neumann boundary
+        if self.participant_name == 'Dirichlet':
+            self.couplingBC = DirichletBC(self.V, coupling_expr, coupling_boundary)
+
+        self.remainingBC = DirichletBC(self.V, solution_expr, remaining_boundary)
```
BenjaminRodenberg: See my other comment. I think we could completely skip this part.
```suggestion
```

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+        self.solution_expr.t = t
+        self.remainingBC.apply(T, b.values.vector())
+
+        # Update the coupling boundary condition for the Dirichlet participant
+        if self.participant_name == 'Dirichlet':
+            dt = t - self.t_start                   # This dt is used to read data from the current time window
+            read_data = self.precice.read_data(dt)  # Read the data to update the coupling expression
+            self.precice.update_coupling_expression(
+                self.coupling_expression,
+                read_data)    # Update the coupling expression
```
BenjaminRodenberg, line 100: This part is a bit tricky. So we definitely need a handle to the `coupling_expression`

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+    def solve_system(self, rhs, factor, u0, t):
+        """
+        Solves (M - factor * A) u = rhs.
+        """
+        u = self.dtype_u(u0)
+        T = self.M - factor * self.K
+        b = self.dtype_u(rhs)
+
+        self.solution_expr.t = t
+        self.remainingBC.apply(T, b.values.vector())
```
BenjaminRodenberg, line 92:
```suggestion
```
Could be removed. See below.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Update the coupling boundary condition for the Dirichlet participant
+        if self.participant_name == 'Dirichlet':
+            dt = t - self.t_start                   # This dt is used to read data from the current time window
+            read_data = self.precice.read_data(dt)  # Read the data to update the coupling expression
+            self.precice.update_coupling_expression(
+                self.coupling_expression,
+                read_data)    # Update the coupling expression
+
+            self.couplingBC.apply(T, b.values.vector())  # Apply the coupling boundary condition
```
BenjaminRodenberg, line 102: Assuming `self.dirichlet_bcs` is a list of all `DirichletBC`s. I don't think we need to differentiate between coupling and non-coupling BCs.

```suggestion
    for bc in self.dirichlet_bcs:
        bc.apply(T, b.values.vector())  # Apply all Dirichlet boundary conditions
```

BenjaminRodenberg (changes requested):
I just spotted another simplification of the code. Sorry for the messy review...

partitioned-heat-conduction/dirichlet-fenics/run.sh
```diff
@@ -4,6 +4,33 @@ set -e -u
...
+        sdc)
+            echo "Running simulation with pySDC+FEniCS implementation"
+            python3 ../solver-fenics/heat_pySDC.py Dirichlet
+            ;;
+        *)
+            echo "Running simulation with default FEniCS implementation"
+            python3 ../solver-fenics/heat.py Dirichlet
+            ;;
+    esac
+fi
```
BenjaminRodenberg: @MakisH any comments here? I think it is generally a good idea to make the higher-order time-stepping schemes more accessible. The simple `./run.sh` still behaves as before.

MakisH: Having a default option + more options is totally fine, since there is a default that one can run without reading the details. :+1:

I have not tested this, but if shellcheck passes and you have tested all options, I don't see anything else to check.

partitioned-heat-conduction/dirichlet-pySDC-fenics/run.sh
MakisH: See https://github.com/precice/tutorials/pull/557#discussion_r1738261556, I think the approach with the default+more options is nice.

partitioned-heat-conduction/dirichlet-fenics/run.sh
BenjaminRodenberg: Still have to do the testing, but that's on my list for today. Then let's resolve this and keep it here.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
TobiasEppacher: The idea is good, but it won't reduce the complexity, as the expressions for the DirichletBCs have to be updated within the problem class if they are time-dependent. This is already the case with `self.solution_expr.t = t` before the application of the `DirichletBC` for the remaining boundary.
Similarly we would need to pass the `Expression` objects for any other DirichletBC into the problem class anyways to update them.

BenjaminRodenberg: True. Let's keep this refactoring out of this PR and wrap up.

partitioned-heat-conduction/dirichlet-fenics/run.sh
```diff
@@ -1,9 +1,40 @@
...
+    echo "Running simulation with default FEniCS implementation"
+    python3 ../solver-fenics/heat.py Dirichlet
+else
+    case "$1" in
+        -[hH]|--help)
+            echo "Usage: $0 [highorder|sdc|*]"
+            echo ""
+            echo "      highorder: Run simulation with higher order FEniCS implementation"
+            echo "      sdc      : Run simulation with pySDC+FEniCS implementation"
+            echo "      *        : For every other input run the simulation with default FEniCS implementation"
```
BenjaminRodenberg:
```suggestion
            echo "      irk : Run simulation with higher-order implicit Runge-Kutta schemes FEniCS implementation"
            echo "      sdc : Run simulation with pySDC+FEniCS implementation"
            echo "      *   : For every other input run the simulation with default FEniCS implementation"
```
`irk` instead of `highorder` because `sdc` is actually also high order. The `irk` in irksome (which was a template for Niklas' work) actually stands for Implicit Runge Kutta.

partitioned-heat-conduction/dirichlet-fenics/run.sh
```diff
@@ -1,9 +1,40 @@
...
+            exit 0
+            ;;
+        highorder)
+            echo "Running simulation with higher order FEniCS implementation"
```
BenjaminRodenberg:
```suggestion
        irk)
            echo "Running simulation with higher-order implicit Runge-Kutta schemes FEniCS implementation"
```

BenjaminRodenberg (approved):
Minor renaming. Rest looks good and will take care of merging after some final tests.

partitioned-heat-conduction/solver-fenics/requirements.txt
MakisH, line 1:
```
Running simulation with pySDC+FEniCS implementation
/usr/lib/python3/dist-packages/scipy/__init__.py:146: UserWarning: A NumPy version >=1.17.3 and <1.25.0 is required for this version of SciPy (detected version 1.26.4
  warnings.warn(f"A NumPy version >={np_minversion} and <{np_maxversion}"
  File "/home/gc/repos/precice/tutorials/partitioned-heat-conduction/dirichlet-fenics/../solver-fenics/heat_pySDC.py", line 44, in <module>
    from pySDC.implementations.sweeper_classes.imex_1st_order_mass import imex_1st_order_mass
ModuleNotFoundError: No module named 'pySDC'
```

BenjaminRodenberg: I got pySDC installed and the script working. Installation is, unfortunately, no fun since we cannot use the pySDC from pip but need to use the develop version from github. Additionally the develop version on github has some breaking changes. But the latest release `5.5.0` is still ok.

The pip pySDC package is still useful to install the dependencies.

The whole installation process is far from ideal but since the sdc time stepping is experimental and quite a niche feature I think we can live with this.

I will document this installation procedure that finally worked out for me and then I think we can merge.

MakisH: What if we just `pip install git+https://github.com/...@tag`?

BenjaminRodenberg: Good idea. I came up with a different solution: [d6a950f](https://github.com/precice/tutorials/pull/557/commits/d6a950f96733f0e908c149946942d384334fb0fc)

BenjaminRodenberg: `pip install git+https://github.com/Parallel-in-Time/pySDC@5.5.0` could actually lead us somewhere but I have to go now and try this next week. Thanks for the hint!

BenjaminRodenberg: With `pip install git+...` the case works nicely in a `.venv`.

partitioned-heat-conduction/dirichlet-fenics/run.sh
```diff
@@ -1,9 +1,43 @@
...
+            exit 0
+            ;;
+        irk)
+            echo "Running simulation with higher-order implicit Runge-Kutta schemes FEniCS implementation"
+            python3 ../solver-fenics/heatHigherOrder.py Dirichlet
+            ;;
+        sdc)
+            # install pySDC + its dependencies only if needed
+            pip install git+https://github.com/Parallel-in-Time/pySDC@5.4.3
+            pip install pySDC~=5.4
```
BenjaminRodenberg, line 32: I will take care of compatibility with `pySDC@develop/5.5.0` directly on develop. I would like to merge this PR first to make sure it's compatible with Tobias' work (he was using 5.4.something).

Only minor changes are required:

```diff
diff --git a/partitioned-heat-conduction/dirichlet-fenics/run.sh b/partitioned-heat-conduction/dirichlet-fenics/run.sh
index b75e8c84..1f7bcb9c 100755
--- a/partitioned-heat-conduction/dirichlet-fenics/run.sh
+++ b/partitioned-heat-conduction/dirichlet-fenics/run.sh
@@ -28,8 +28,8 @@ else
             ;;
         sdc)
             # install pySDC + its dependencies only if needed
-            pip install git+https://github.com/Parallel-in-Time/pySDC@5.4.3
-            pip install pySDC~=5.4
+            pip install git+https://github.com/Parallel-in-Time/pySDC@5.5.0
+            pip install pySDC~=5.5
             echo "Running simulation with pySDC+FEniCS implementation"
             python3 ../solver-fenics/heat_pySDC.py Dirichlet
             ;;
diff --git a/partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py b/partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
index ab952c20..7adb896e 100644
--- a/partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
+++ b/partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
@@ -10,11 +10,11 @@ from fenics import TrialFunction, TestFunction, dx, assemble, inner, nabla_grad,
 from my_enums import ProblemType
 
 
-from pySDC.core.Problem import ptype
+from pySDC.core.problem import Problem
 from pySDC.implementations.datatype_classes.fenics_mesh import fenics_mesh, rhs_fenics_mesh
 
 
-class fenics_heat_2d(ptype):
+class fenics_heat_2d(Problem):
     dtype_u = fenics_mesh
     dtype_f = rhs_fenics_mesh
```

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
BenjaminRodenberg: I will take care of this after merging.

partitioned-heat-conduction/solver-fenics/heat_pySDC.py
BenjaminRodenberg: Fixed. e889225ca17bee24f04072239004552831f7068d.

partitioned-heat-conduction/solver-fenics/heat_pySDC_problem_class.py
BenjaminRodenberg: Tobias mentions this in his thesis. But we cannot link to his thesis yet. So this is to some degree a hen-and-egg problem. Let's take care of this in #560.

MakisH (approved):
The current state works for me (`./run.sh sdc` installs everything in a venv and completes). However, if I run it again, it seems like PIP downloads the same extra packages again.

sdc

BenjaminRodenberg:
> However, if I run it again, it seems like PIP downloads the same extra packages again.

Maybe due to the `git` install? But let's keep it as it is and we can come back to this if it becomes problematic. SDC is quite niche and if we have some stable-but-not-perfect way to run things it's good enough. Will merge now. Thanks @TobiasEppacher for the very nice contribution! Of course, also thanks @MakisH for the valuable input!

Issue #378, opened 2023-10-04
Make `beta` in partitioned heat conduction tutorial consistent with paper


BenjaminRodenberg:
For some reason our tutorial here uses `beta = 1.3`:

https://github.com/precice/tutorials/blob/6023dbd9e67cda2dae61aef5dd7d0ab05b61f66a/partitioned-heat-conduction/fenics/heat.py#L70

However, the [FEniCS tutorial book, p.41](https://fenicsproject.org/pub/tutorial/pdf/fenics-tutorial-vol1.pdf) and the [quasi-Newton paper](https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443) uses `beta = 1.2`. To make things consistent, we should change the value here.

BenjaminRodenberg:
Also important: This change is also required for the nutils and OpenFOAM versions of this case.

BenjaminRodenberg:
Closed this issue via #379.

BenjaminRodenberg:
Marking this issue as a `bug`. I want to backport this fix and release it as a bugfix release via https://github.com/precice/tutorials/pull/558. Main motivation: consistency in results and literature when comparing the distributions for preCICE v2 and v3.

PR #558, opened 2024-08-27
Release v202211.1

BenjaminRodenberg:
See https://github.com/precice/precice.github.io/pull/414 for the motivation for this bugfix release.

I'm creating this branch and PR to start collecting relevant changes. Note that it is still unclear if and how we actually would merge these changes into `master`.

ToDos:

- [ ] backport #563 

Checklist:

### DIFF
--- a/partitioned-heat-conduction-complex/fenics/heat.py
+++ b/partitioned-heat-conduction-complex/fenics/heat.py
@@ -75,7 +75,7 @@ fenics_dt = .1
 # Error is bounded by coupling accuracy. In theory we can obtain the analytical solution.
 error_tol = 10 ** -6
 alpha = 3  # parameter alpha
-beta = 1.3  # parameter beta
+beta = 1.2  # parameter beta
 gamma = args.gamma  # parameter gamma, dependence of heat flux on time
 
 # Create mesh and separate mesh components for grid, boundary and coupling interface
--- a/partitioned-heat-conduction-direct/nutils/heat.py
+++ b/partitioned-heat-conduction-direct/nutils/heat.py
@@ -7,7 +7,7 @@ import numpy as np
 import precice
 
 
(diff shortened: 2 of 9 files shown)


partitioned-heat-conduction-complex/fenics/heat.py
```diff
@@ -75,7 +75,7 @@ def determine_gradient(V_g, u, flux):
 # Error is bounded by coupling accuracy. In theory we can obtain the analytical solution.
 error_tol = 10 ** -6
 alpha = 3  # parameter alpha
-beta = 1.3  # parameter beta
+beta = 1.2  # parameter beta
```
MakisH, line 78: Why do we need to change a model parameter in a bugfix release? I know that this was changed in the v202404, but wouldn't this confuse users if v202211.0 and v202211.1 have different values?

BenjaminRodenberg: The parameter was not consistent with literature. So I would consider it as a bug that we can also fix in a bugfix release. But let's continue this discussion in https://github.com/precice/tutorials/issues/378.

BenjaminRodenberg:
Additionally, if we want to make sure `v202211.1` is future-proof --- respectively even working with Ubuntu 24.04 --- we also could consider backporting #547. This also means that we would need to modify the folder structure of some cases or implement a solution that supports the old structure.

Looks to me like a rabbit hole: Therefore, I would leave things as they are. The user would have to do the `venv` stuff manually. This is to some degree also consistent with the `v202211.1` policy "make sure that your python environment works; we do not even provide the `requirements.txt` for some cases".

MakisH:
Since @BenjaminRodenberg has published the datasets elsewhere (please comment), this is not relevant anymore.

PR #563, opened 2024-08-31
Add missing `requirements.txt` and venvs for Python tutorials

NiklasVin:
Adds the missing `requirements.txt` files mentioned in https://github.com/precice/tutorials/issues/547 and modifies the run scripts accordingly.

I followed the approach suggested here: https://github.com/precice/tutorials/issues/547#issuecomment-2317190857 to prevent redundant requirement files.
This concerns the following tutorials:
- partitioned-heat-conduction
- partitioned-heat-conduction-complex
- partitioned-heat-conduction-overlap
- volume-coupled-diffusion
- oscillator-overlap
- oscillator


channel-transport-reaction/chemical-fenics/requirements.txt
```diff
@@ -1,3 +1,5 @@
 numpy >1, <2
 pyprecice~=3.0
-scipy
+mpi4py < 4
```
MakisH: Ah, nice! Did you face the same issue as in https://github.com/precice/precice/pull/2052#issuecomment-2321693938 and is this the solution?

MakisH (commented):
I have not tried to run or check if each `requirements.txt` included all and the correct dependencies, but the changes look good.

Thanks also for catching and fixing some small inconsistencies (e.g., forgotten `close_log`, or initializing the `venv` before starting to log)!

channel-transport-reaction/fluid-fenics/requirements.txt
BenjaminRodenberg, line 1: Strange that git interprets this as a renaming of `oscillator/mass-left-python/requirements.txt` but looks correct.

elastic-tube-1d/fluid-python/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pyprecice~=3.0
+numpy
```
BenjaminRodenberg:
```suggestion
numpy >1, <2
```
See #564. Let's generally use this for numpy at the moment since numpy 2 is not widely supported yet.

channel-transport-reaction/chemical-fenics/requirements.txt
BenjaminRodenberg: Is `mpi4py` a requirement of this case or is it actually a requirement of `fenicsprecice`? If `mpi4py < 4` helps us to solve some issue, we should consider adding this requirement in `fenicsprecice` and remove it here. This case should then also inherit the requirement from `fenicsprecice`. @NiklasVin can you open the corresponding PR in the FEniCS adapter repo if this makes sense?

elastic-tube-1d/fluid-python/requirements.txt
BenjaminRodenberg: We could also consider adding `numpy >1, <2` in `pyprecice` [here](https://github.com/precice/python-bindings/blob/1f750461c1d023f4de5d394804393be0566f904e/pyproject.toml#L3C1-L3C121). Then we would not need to specify the requirement here again. But I'm unsure what's the best practice here. Currently, we are duplicating the requirement but it also does not harm the way it is. I'm fine with either approach.

Assuming `pyprecice` might drop `numpy` at some point in time: If this case still needs `numpy` it would run into a problem. So I have a tendency towards keeping numpy here if there is an `import numpy` showing up anywhere.

elastic-tube-1d/solid-python/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pyprecice~=3.0
+numpy
```
BenjaminRodenberg: See above.

elastic-tube-3d/solid-fenics/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pyprecice~=3.0
```
BenjaminRodenberg:
```suggestion
```
`pyprecice` is no direct dependency of this case but of `fenicsprecice` we could consider removing it here.

flow-around-controlled-moving-cylinder/solid-python/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pyprecice~=3.0
+numpy
```
BenjaminRodenberg:
```suggestion
numpy >1, <2
```

elastic-tube-3d/solid-fenics/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+fenicsprecice
+numpy
```
BenjaminRodenberg:
```suggestion
numpy >1, <2
```

elastic-tube-3d/solid-fenics/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+mpi4py <4
```
BenjaminRodenberg:
```suggestion
```
Move this into `fenicsprecice`?

channel-transport-reaction/chemical-fenics/requirements.txt
```diff
@@ -1,3 +1,5 @@
...
+fenicsprecice
```
BenjaminRodenberg:
```suggestion
fenicsprecice~=2.0
```
Be aware that for `fenicsprecice` we had a breaking release consistently with preCICE 3.0.0. Therefore, we should not accept `fenicsprecice<2.0`.

BenjaminRodenberg (changes requested):
Thank you for the PR! This really gives a very good overview of the current state and where we have to clarify some questions. I added some comments to point to the open issues. I will briefly summarize here:

* Generally restrict numpy.
* Is `mpi4py` actually a `fenicsprecice` dependency? Move it there?
* If `pyprecice` is a `fenicsprecice` dependency move it there
* Find a good approach for FEniCS

I would suggest to merge #564 and #557 independently from this PR because I expect that we are quicker with these smaller PRs.

channel-transport-reaction/fluid-fenics/requirements.txt
```diff
@@ -1,3 +1,5 @@
 numpy >1, <2
 pyprecice~=3.0
-scipy
+mpi4py <4
+fenicsprecice
+
```
BenjaminRodenberg, line 3: In #557 I added the following:

```suggestion
# Assuming FEniCS from ppa:fenics-packages/fenics was installed https://fenicsproject.org/download/archive/
# Use --system-site-packages in venv
fenics-dijitso==2019.2.0.dev0
fenics-dolfin==2019.2.0.13.dev0
fenics-ffc==2019.2.0.dev0
fenics-fiat==2019.2.0.dev0
fenics-ufl-legacy==2022.3.0
```

I'm not sure that this is really the best approach but `--system-site-packages` pollutes the `venv` quite a lot. As far as I know, there is unfortunately no possibility to only use the packages related to FEniCS from the system and skip the rest.

Note that to some degree this is again also a FEniCS adapter problem. But the tutorials also directly depend on FEniCS. Therefore, listing `fenicsadapter` and FEniCS here looks ok.

elastic-tube-1d/fluid-python/requirements.txt
MakisH: I have not checked the details, but aren't we using the removed/changed functions also directly in the code we have in the tutorials? Then, specifying this as a restriction in the `pyprecice` would not be the right place.

Let's specify the exact restrictions here and remove the restrictions gradually, as things become compatible.

channel-transport-reaction/chemical-fenics/requirements.txt
NiklasVin: @MakisH Yes, I had the same issue. With `mpi4py < 4`, I didn't get an error and everything worked.

NiklasVin: > Is `mpi4py` a requirement of this case or is it actually a requirement of `fenicsprecice`? If `mpi4py < 4` helps us to solve some issue, we should consider adding this requirement in `fenicsprecice` and remove it here. This case should then also inherit the requirement from `fenicsprecice`. @NiklasVin can you open the corresponding PR in the FEniCS adapter repo if this makes sense?

I think this is a requirement of FEniCS. So, I guess, it makes sense to add this as a requirement to the FEniCS adapter.

MakisH: Yes, let's add it to the FEniCS adapter.

channel-transport-reaction/fluid-fenics/requirements.txt
NiklasVin: I wasn't content with the `system-site-packages` procedure as well. That's why I looked for an alternative and found [one](https://stackoverflow.com/questions/14571454/virtualenv-specifing-which-packages-to-use-system-wide-vs-local). By creating symbolic links to the system-wide installation of FEniCS we can circumvent including all packages with `system-site-packages`. And, since legacy FEniCS is not going to be updated, this shouldn't be a maintenance issue as well, because the version won't change.
Linking all dependencies of FEniCS by hand is a bit annoying, but adding the commands to the `run.sh` file shouldn't be a problem.
I also tested it with one tutorial case and it seems to work with a clean venv.

What's your opinion on this approach?

BenjaminRodenberg: Hmm... I also saw this solution but was not really convinced. I'm mainly concerned about different ways FEniCS might be installed. I think it's not necessarily always `/usr/lib/python3/dist-packages` and I don't know an automated way to find the right prefix.

Can you maybe put your solution here into the comments to get a better impression?

NiklasVin: Until now, I added the paths manually only, so I don't know if the following idea is a good approach. We can recursively search the directory to the packages (system-wide) by using, for example `find / -name fenics_dolfin-2019.2.0.13.dev0.egg-info`. From the output, you could extract the paths of each required package. We could also think about searching in the `/usr/lib/` only. If the packages are not installed there, we can use the `system-site-packages` as a fallback option.

BenjaminRodenberg: Searching for an explicit name (here `fenics_dolfin-2019.2.0.13.dev0.egg-info`) sounds like something that breaks easily and it is also not working with other versions. Imagine somebody is using an (even older) FEniCS.

I would suggest to use the `system-site-packages`. That's not perfect since it clutters the venv a bit but it is overall simple and good enough for the purpose.

MakisH: I would also vote in keeping the `system-site-packages` if it works.
If we face problems later, then we can think of alternative solutions.
But this solution seems to me the less error-prone of the two.

NiklasVin:
> If `pyprecice` is a `fenicsprecice` dependency move it there

I saw that `pyprecice` is a dependency of `fenicsprecice`, so I am going to remove it from the requirements👍

channel-transport-reaction/chemical-fenics/requirements.txt
```diff
@@ -0,0 +1,9 @@
+numpy >1, <2
+fenicsprecice~=2.0
+
```
BenjaminRodenberg, line 3: I added the following comment in https://github.com/precice/tutorials/blob/develop/partitioned-heat-conduction/solver-fenics/requirements.txt. But we can also skip this. I think it's generally useful to add some information about the somewhat "strange" situation with the FEniCS python package.

But this is your decision and we can also leave it as it is.

```suggestion

# Assuming FEniCS from ppa:fenics-packages/fenics was installed https://fenicsproject.org/download/archive/
# Use --system-site-packages in venv
```

elastic-tube-3d/solid-fenics/run.sh
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
 set -e -u
```
BenjaminRodenberg, line 3:
```suggestion

. ../../tools/log.sh
exec > >(tee --append "$LOGFILE") 2>&1
```
Why is this part missing here?

BenjaminRodenberg (approved):
Looks good :+1:

Just a few minor (non-critical) point left. We can merge this PR as it is or you can quickly take care about the fixes.

I will merge this PR latest over the weekend.

PR #569, opened 2024-09-06
Include "history.csv" in gitignore

j-signorelli:
For non-restart SU2 simulations, a history.csv file is outputted. This PR simply ensures that these are ignored.

Checklist:


MakisH (approved):
Perfect, thank you! :hugs: 

I don't think we really need a changelog entry for this.

PR #570, opened 2024-09-10
Improvements to plotting function for flow-over-heated-plate

j-signorelli:
This PR updates and improves the plot-final-interface-temperature.py file:

- Previously, the file would only run properly when all solid solvers were run; now it works regardless of how many have been run
- When data was not found, the last file failed because `data_id` was not defined, and `if not data_id` would cause an error. This updated script correctly prints what files/data is/are not found in the search.
- General improvements (such as using `enumerate`, removing unnecessary retrieval of other data arrays, etc.)
- Because either SU2 or OpenFOAM may be used as the fluid-participant, the labels were changed as `OpenFOAM-nutils` to `Fluid-nutils` for example.

This PR also adds SU2 to the metadata.yaml file in flow-over-heated-plate.

No user-facing changes.

Checklist:

- [X] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.
- [X] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.


MakisH (approved):
We somehow overlooked this for a while. I had a look now and directly modified a few things:

- Switched to reading `.vtu` files instead of `.vtk`, which is now the default. I had to adapt the reader for that.
- Removed the `metadata.yml` change. We should do this in a separate PR, and it is mainly meant for the system tests.

I tested this and it works:

Figure_1


The changes in the code also make sense.

PR #571, opened 2024-09-10
Add multigrid settings to flow-over-heated-plate SU2 config

j-signorelli:
This PR adds multigrid settings to the flow-over-heated-plate SU2 config file, speeding up runtime.

Checklist:


uekerman:
Could you please briefly document how much the speedup here is? Is it already significant for such a small mesh? Thanks!

j-signorelli:
Yes of course: on my system with 3 MPI procs, without multigrid took ~444 seconds and with these multigrid settings took 168 seconds, so the speedup is 2.5-3x improvement.

MakisH (approved):
I have not tried to run this, but I trust that this should only improve the speed of the tutorial.

I assume you have checked that results are still the same, right?

https://precice.org/tutorials-flow-over-heated-plate.html#compare-results

j-signorelli:
Yep! Results are indeed still the same.

PR #587, opened 2024-11-11
System tests: Enable preCICE debug messages and symbols via an argument

### LINKED ISSUE #586: System tests: Write preCICE debug logs
Would be useful to always write (also) the debug information, ideally directly into a file via defining a separate sink.

MakisH:
This adds an additional argument to the system tests to pass the CMake preset to the preCICE build. It defaults to `production-audit`, which also enables debug messages, but one could then manually switch it, e.g., to `optimized-debug`.

My motivation for making this an argument is that we should be testing something close to the user, but sometimes it would be useful to get additional debug symbols (e.g., in the coredump file that is included in the archive). Note that for this we would also need to get the binary from the Docker container. This is a bit more complicated, but we could get it from the VM, if needed.

It also now writes `debug` and `trace` log files for every tutorial.

Respective system tests job: https://github.com/precice/tutorials/actions/runs/11798415337

closes #586 

Checklist:


tools/tests/docker-compose.template.yaml
```diff
@@ -13,6 +13,8 @@ services:
     command: >
       /bin/bash -c "id &&
       cd '/runs/{{ tutorial_folder }}' && 
+      sed -i 's|</log>|<sink type=\"file\" output=\"precice-debug.log\" filter=\"%Severity% >= debug and %Rank% = 0\" enabled=\"true\" /></log>|g' precice-config.xml &&
+      sed -i 's|</log>|<sink type=\"file\" output=\"precice-trace.log\" filter=\"%Severity% >= trace and %Rank% = 0\" enabled=\"true\" /></log>|g' precice-config.xml &&
```
MakisH: Not sure if the trace output could cause any issues (file size, runtime).

fsimonis (approved):

PR #597, opened 2024-11-21
`oscillator-overlap`: Make consistent with `oscillator`

BenjaminRodenberg:
Updates `oscillator-overlap` to use time stepping schemes and typing like in `oscillator`

Checklist:

### DIFF
--- a/changelog-entries/597.md
+++ b/changelog-entries/597.md
@@ -1,0 +1,1 @@
+- Offer Runge-Kutta 4 scheme and black-box time-stepping with RadauIIA scheme in oscillator-overlap case. Demonstrates use of time interpolation and dense output.
--- a/changelog-entries/597.md
+++ b/changelog-entries/597.md
@@ -1,1 +1,1 @@
-- Offer Runge-Kutta 4 scheme and black-box time-stepping with RadauIIA scheme in oscillator-overlap case. Demonstrates use of time interpolation and dense output.
+- Added Runge-Kutta 4 scheme and black-box time-stepping with RadauIIA scheme as options to oscillator-overlap case. This demonstrates use of time interpolation and dense output.
--- a/oscillator-overlap/solver-python/oscillator.py
+++ b/oscillator-overlap/solver-python/oscillator.py
@@ -182,12 +182,6 @@ while participant.is_coupling_ongoing():
         t_write.append(t)
 
 # store final result
-u = u_new
-v = v_new
-a = a_new
-u_write.append(u)
-v_write.append(v)
-t_write.append(t)
 positions += u_write
 velocities += v_write
 times += t_write
--- a/oscillator-overlap/solver-python/oscillator.py
+++ b/oscillator-overlap/solver-python/oscillator.py
@@ -74,7 +74,7 @@ dimensions = participant.get_mesh_dimensions(mesh_name)
 
 vertex = np.zeros(dimensions)
 read_data = np.zeros(num_vertices)
-write_data = connecting_spring.k * u0 * np.ones(num_vertices)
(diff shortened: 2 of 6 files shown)


oscillator-overlap/solver-python/oscillator.py
```diff
@@ -68,7 +74,7 @@ class Participant(Enum):
 
 vertex = np.zeros(dimensions)
 read_data = np.zeros(num_vertices)
-write_data = connecting_spring.k * u0 * np.ones(num_vertices)
```
BenjaminRodenberg, line 71: This is actually a bug: In `oscillator` we exchange forces. Writing displacement * spring stiffness is correct. Compare https://github.com/precice/tutorials/blob/dea77cbfdbc3d5bdc0df1e745bbc3c3c5ed2b2d4/oscillator/solver-python/oscillator.py#L77. In this case, i.e. `oscillator-overlap` we have to write displacements. The multiplication with `connecting_spring.k` is incorrect.

changelog-entries/597.md
```diff
@@ -0,0 +1 @@
+- Offer Runge-Kutta 4 scheme and black-box time-stepping with RadauIIA scheme in oscillator-overlap case. Demonstrates use of time interpolation and dense output.
```
uekerman:
```suggestion
- Added Runge-Kutta 4 scheme and black-box time-stepping with RadauIIA scheme as options to oscillator-overlap case. This demonstrates use of time interpolation and dense output.
```

uekerman (approved):
Makes sense to me. I did not look into the details and did not run the case.

oscillator-overlap/solver-python/oscillator.py
```diff
@@ -6,13 +6,10 @@
...
 import os
+from typing import Type
 
 import problemDefinition
-
-
-class Scheme(Enum):
-    NEWMARK_BETA = "Newmark_beta"
-    GENERALIZED_ALPHA = "generalized_alpha"
+from timeSteppers import TimeStepper, TimeSteppingSchemes, GeneralizedAlpha, RungeKutta4, RadauIIA
```
NiklasVin, line 12: The `timeSteppers` module is missing.

oscillator-overlap/solver-python/oscillator.py
```diff
@@ -22,20 +19,30 @@ class Participant(Enum):
 
 parser = argparse.ArgumentParser()
 parser.add_argument("participantName", help="Name of the solver.", type=str, choices=[p.value for p in Participant])
-parser.add_argument("-ts", "--time-stepping", help="Time stepping scheme being used.", type=str,
-                    choices=[s.value for s in Scheme], default=Scheme.NEWMARK_BETA.value)
+parser.add_argument(
```
NiklasVin, line 22: This comment is more of a general usability concern of this feature : Assuming the user wants to use a different time stepping scheme but wants to use the `run.sh` files only, then, as far as I can see, the time stepping scheme cannot be changed.

NiklasVin (changes requested):

oscillator-overlap/solver-python/oscillator.py
BenjaminRodenberg: Fixed. Thanks for checking!

oscillator-overlap/solver-python/oscillator.py
BenjaminRodenberg: True. However, in `tutorials/oscillator` we also do not offer the option to select a different time stepping scheme using the `run.sh` script. Therefore, I would keep it as it is and not add a new feature here.

NiklasVin (approved):
Now everything works 👍

PR #599, opened 2024-11-26
Add channel transport with remeshing

fsimonis:
This PR adds support for adaptive mesh refinement to the transport solver of the channel transport tutorial.

The `transport-nutils` runs normally without arguments.
Using `./run.sh remesh=True` enables adaptive mesh refinement every 2 time windows. This mode starts with a mesh of a quarter of the total cells.
Using `./run.sh remesh=True frequency=5` sets the refinement frequency to every 5 time windows.

Checklist:


uekerman:
This is the case that we specifically set up as a testcase for remeshing. No need to copy in my opinion. Simply edit it directly.

https://github.com/precice/tutorials/pull/269

fsimonis:
@uekerman This way, we lose our simple volume-coupling case though.

uekerman:
There are enough other cases for volume coupling.

fsimonis:
Merged the solvers, and updated OP

channel-transport/precice-config.xml
```diff
@@ -33,10 +33,10 @@
 
   <m2n:sockets acceptor="Fluid" connector="Transport" exchange-directory=".." />
 
-  <coupling-scheme:serial-explicit>
+  <coupling-scheme:parallel-explicit>
```
MakisH, line 36: The config visualization needs to be updated as well. I assume this is a requirement for the remeshing at the moment?

channel-transport/transport-nutils/requirements.txt
```diff
@@ -1,3 +1,4 @@
 nutils==7
 numpy >1, <2
-pyprecice~=3.0
+pyprecice @ git+https://github.com/precice/python-bindings.git@develop
```
MakisH, line 4: Would be good to have a comment here describing why we need develop, to remember to set a released version once such a version is available.

channel-transport/transport-nutils/requirements.txt
```diff
@@ -1,3 +1,4 @@
...
+setuptools
```
MakisH, line 5: Why do we need setuptools now but not before?

channel-transport/transport-nutils/transport.py
```diff
@@ -96,15 +180,54 @@ def main():
...
+    # args = parser.parse_args()
+
+    # print(args)
+    # if args.remesh:
+    #    assert args.frequency > 0
+    #    assert len(args.limits) > 0
+
+    # cli.run(main)
+    # cli.run(main,args.remesh, args.frequency, args.limits, args.visualize)
+
```
MakisH: Why are these commented out?

channel-transport/README.md
```diff
@@ -39,7 +43,7 @@ Fluid participant:
 
 Transport participant:
 
-* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0.
+* Nutils with support for adaptive mesh refinement. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0.
```
MakisH: Maybe link to the documentation page for remeshing? There, I would also expect some statement on the minimum preCICE version.

channel-transport/transport-nutils/requirements.txt
fsimonis: Nutils needs this with a newer pip/python. This should impact all nutils tutorials.

channel-transport/precice-config.xml
fsimonis: Done

channel-transport/transport-nutils/transport.py
fsimonis: Cleaned up

channel-transport/transport-nutils/requirements.txt
MakisH: Opened a PR to update everything, then: https://github.com/precice/tutorials/pull/602

channel-transport/transport-nutils/requirements.txt
fsimonis: Done

channel-transport/README.md
```diff
@@ -39,7 +43,7 @@ Fluid participant:
 
 Transport participant:
 
-* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0.
+* Nutils with support for [adaptive mesh refinement](https://precice.org/couple-your-code-moving-or-changing-meshes.html#pseudo-reference-domain). For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0 and a preCICE release above 3.1.2 for remeshing support.
```
MakisH: I understand why you write "above 3.1.2", but this could easily be misinterpreted as "3.1.2 or higher". Yes, this is not what it means, but just how some people might interpret it.

You could formulate it as:

```suggestion
* Nutils with support for [adaptive mesh refinement](https://precice.org/couple-your-code-moving-or-changing-meshes.html#pseudo-reference-domain). For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v7.0 and a preCICE release with [remeshing support](couple-your-code-moving-or-changing-meshes.html#remeshing-using-precice).
```

and document the version requirements in that page. You can write there that it is planned for 3.2.0, and you will probably have to update that page soon after anyway.

You could also create an issue with TODOs after releasing the feature, to update that page, the `requirements.txt`, and anything else related.

MakisH (approved):
Very nice test case! :+1:

I ran it and the results look good: I see the updated mesh, and the simulation completes normally.

- The `fluid-nutils`/`transport-nutils` case took me 2h.
- The `fluid-openfoam`/`transport-nutils` case took me 20min.

Not sure if it makes sense to document somewhere that Nutils, especially the `fluid-nutils`, takes comparatively too long. Mainly to guide the user to prefer the faster OpenFOAM case and to not blame the runtime solely on preCICE.

channel-transport/README.md
fsimonis: Good idea to point to the changing meshes-page. Fewer places to update.

PR #600, opened 2024-11-27
Growing mesh case

fsimonis:
This PR adds a growing mesh case.

Two solvers define the same mesh, which grows at given points in time.

Inspiration for this case is the formation of additional residue layers on the floor of an ocean over time.

@MakisH is this how you envision the separate solver folder to work?

Checklist:


growing-mesh/README.md
```diff
@@ -0,0 +1,43 @@
...
+permalink: tutorials-growing-mesh.html
+keywords: python, remeshing
+summary: The growing mesh case is a showcase example of two solvers which grow their mesh at predefined points in time.
+---
+
+{% note %}
+Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/growing-mesh). Read how in the [tutorials introduction](https://precice.org/tutorials.html).
+{% endnote %}
+
+## Setup
```
MakisH: pending

growing-mesh/README.md
```diff
@@ -0,0 +1,43 @@
...
+preCICE configuration (image generated using the [precice-config-visualizer](https://precice.org/tooling-config-visualization.html)):
+
+![preCICE configuration visualization](images/tutorials-growing-mesh-precice-config.png)
+
+## Available solvers
+
+There are two solvers that define the same mesh:
+
+- A who runs first
+- B who runs second
```
MakisH: We so far avoid calling participants and data generic names, but try to relate to a specific physical domain and application use case. This is currently only partially described in the [naming conventions](https://precice.org/community-contribute-to-precice.html#naming-conventions).

Of course, this is fine for now (and fine for an integration test), but before merging, we should better have more descriptive names.

growing-mesh/README.md
```diff
@@ -0,0 +1,43 @@
...
+```bash
+cd fluid-cpp
+./run.sh 2
+```
+
+and
+
+```bash
+cd solid-cpp
+./run.sh 2
```
MakisH: fixme

growing-mesh/solver/solver.py
MakisH, line 1: The folder should be called `solver-python` (we might have more options later). See https://precice.org/community-contribute-to-precice.html#tutorial-specific-codes

MakisH (commented):
I know that this is still a draft, but I had a quick look anyway, so I left some comments that I know are probably already on your list.

growing-mesh/README.md
fsimonis: They both represent the same mesh and don't model any physical behaviour.

Not sure what we would name them.

growing-mesh/solver/solver.py
fsimonis: Done.

growing-mesh/README.md
fsimonis: Done

growing-mesh/README.md
uekerman: There is a physical motivation for this scenario. We could borrow names there?

fsimonis: The pyhsical motivation is to couple a hydrogeological and geomechanical model and let them iterate until they agree on effective stress.
Without the iterative scheme, there is no correction and the geomechanical model doesn't have a function.

So, I cannot see obvious names that we can use here.

MakisH: But then we need to better express the motivation in the description and say that the modeling is (extremely) simplified, as is the case in general in our tutorials.

With the current formulation, it is a great integration test, but it looks a bit odd as a tutorial.

fsimonis: I added the motivation and more details about the solver.

growing-mesh/solver-python/pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[project]
+name = "growing"
```
MakisH, line 2: Could we give a more descriptive name?
Not necessary, but If a user were to install this outside the venv, a `growing` binary lying around would be quite strange.

One direction could be `growing-mesh-solver-python`.

growing-mesh/README.md
```diff
@@ -0,0 +1,56 @@
...
+summary: The growing mesh case is a showcase example of two solvers which grow their mesh at predefined points in time.
+---
+
+{% note %}
+Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/growing-mesh). Read how in the [tutorials introduction](https://precice.org/tutorials.html).
+{% endnote %}
+
+## Motivation
+
+This case is an extremely simplified version of the scenario presented by Jean-Marc Gratien during Coupled 2023 *Implementing a Comprehensive Hydromechanical Model for Sedimentary Basins by Coupling a 3D Mechanical Code to a Classic Basin Fluid Flow Code with the PreCICE Library*.
```
MakisH, line 14: Do we need the title case (with `PreCICE`) here? 😬

growing-mesh/README.md
```diff
@@ -0,0 +1,56 @@
...
+The scheme iterates until both model agree on effective stress.
+At fixed points in time, geological events deposit new layers onto the mesh, effectively growing the mesh into z-direction.
+
+This tutorial aims to implement such a growing mesh scenario without the implicit coupling.
+
+## Setup
+
+The problem consists of a unit-square uniformly discretized by 512 x 512 nodes.
+The unit square is partitioned among the ranks of a solver according to an n x m grid, where 512 must be divisible by n and m.
+The mesh starts with 2 nodes in z direction and at a given frequency, 2 nodes are added to the mesh, changing only the load per rank, not the partitioning.
```
MakisH, line 26: The original was a bit difficult for me to read, maybe this is better:

```suggestion
The mesh starts with two nodes in z direction. In regular intervals, two nodes are added to the mesh, changing only the load per rank, but not the partitioning.
```

growing-mesh/README.md
```diff
@@ -0,0 +1,56 @@
...
+
```
MakisH, line 27: "A and B" are mentioned below, but not introduced. Compromise:

```suggestion

While motivated by a setup with a hydrogeological and a geomechanical model, we name these participants A and B, as this tutorial only aims to demonstrate the any such growing mesh case scenario.
```

growing-mesh/README.md
```diff
@@ -0,0 +1,56 @@
...
+cd A
+./run.sh 2
+```
+
+and
+
+```bash
+cd B
+./run.sh 2
+```
```
MakisH, line 56: A post-processing section would be nice, hinting to what the user should see in the end. Note that there is no other figure (besides the config visualization) in this tutorial. Someone that does not yet know what "growing mesh" means would need to read and imagine a bit first, before being able to get a first overview.

An image with the mesh in two different times (potentially with partitions shown) would be helpful.

growing-mesh/solver-python/solver.py
```diff
@@ -0,0 +1,137 @@
...
+        data = participant.read_data(mesh_name, read_data_name, vertex_ids, dt)
+        if rank == 0:
+            print(f"{participant_name}: data: {data[0]} * {len(data)}")
+
+        if args.remesh and requiresEvent(tw):
+            oldCount = len(coords)
+            coords = getMeshAtTimeWindow(tw)
+            if rank == 0:
+                print(
+                    f"{participant_name}: Event grows local mesh from {oldCount} to {len(coords)}"
```
MakisH: I fixed the formatting of these strings (here and in line 58). Cross-check.

MakisH (commented):
A few more minor changes. The case starts, I have not run it beyond `Compute consistent mapping` yet.

I directly edited the run scripts for consistency.

growing-mesh/precice-config.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<precice-configuration experimental="True" allow-remeshing="True">
+  <log>
+    <sink
+      filter="%Severity% >= debug and %Rank% = 0"
```
MakisH: Debug messages should probably not be enabled by default.

growing-mesh/solver-python/solver.py
MakisH: Interestingly, while the solver messages show up in the correct order when running with two ranks, they show up separately at the end when running serially:

```
--[precice]  time-window 11 (max: 12), t 10, Dt 1, max-dt 1
---[precice]  Time window completed
---[precice]  Mapping "Data-B" for t=11 from "B-Mesh" to "A-Mesh"
---[precice]  time-window 12 (max: 12), t 11, Dt 1, max-dt 1
---[precice]  Reinitializing Participant
---[precice]  Close distributed communication channels
---[precice]  Setting up primary communication to coupling partner/s
---[precice]  Primary ranks are connected
---[precice]  Setting up preliminary secondary communication to coupling partner/s
---[precice]  Prepare partition for mesh A-Mesh
---[precice]  Gather mesh A-Mesh
---[precice]  Send global mesh A-Mesh
---[precice]  Receive global mesh B-Mesh
---[precice]  Setting up secondary communication to coupling partner/s
---[precice]  Secondary ranks are connected
---[precice]  Time window completed
---[precice]  Computing "nearest-neighbor" mapping from mesh "B-Mesh" to mesh "A-Mesh" in "read" direction.
---[precice]  Mapping distance min:0 max:0 avg: 0 var: 0 cnt: 2621440
---[precice]  Mapping "Data-B" for t=12 from "B-Mesh" to "A-Mesh"
---[precice]  Reached end at: final time-window: 12, final time: 12
---[precice]  Implicitly finalizing in destructor
---[precice]  Close communication channels
Rank 0/1 has partition (0, 0)/(1, 1)
Each of 1 partitions has node size 512x512 = 262144 for a total of 262144 nodes on the base
A: data: 0.0 * 524288
A: data: 1.0 * 524288
A: data: 2.0 * 524288
A: Event grows local mesh from 524288 to 1048576 and global mesh from 524288 to 1048576
A: data: 3.0 * 1048576
A: data: 4.0 * 1048576
A: data: 5.0 * 1048576
A: Event grows local mesh from 1048576 to 1572864 and global mesh from 1048576 to 1572864
A: data: 6.0 * 1572864
A: data: 7.0 * 1572864
A: data: 8.0 * 1572864
A: Event grows local mesh from 1572864 to 2097152 and global mesh from 1572864 to 2097152
A: data: 9.0 * 2097152
A: data: 10.0 * 2097152
A: data: 11.0 * 2097152
A: Event grows local mesh from 2097152 to 2621440 and global mesh from 2097152 to 2621440
```

growing-mesh/README.md
MakisH: Does this case even produce any results to show how the mesh grows?

MakisH:
With a quick discussion, I understand that a few more updates are planned in this tutorial before another review is needed.

PR #602, opened 2024-12-03
Add setuptools to Nutils-related requirements.txt

MakisH:
@fsimonis tells me in https://github.com/precice/tutorials/pull/599#discussion_r1867678913 that we now need to specify `setuptools` as a requirement whenever we also specify `nutils` for newer PIP versions.

I have not verified this, but here is a PR to at least not forget (but it should not harm anyway).

Checklist:

- I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`. -> N/A

### DIFF
--- a/channel-transport/fluid-nutils/requirements.txt
+++ b/channel-transport/fluid-nutils/requirements.txt
@@ -1,3 +1,4 @@
+setuptools # required by nutils
 nutils==7
 numpy >1, <2
 pyprecice~=3.0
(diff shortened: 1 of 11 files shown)


fsimonis (approved):

fsimonis:
I think this may be a side effect of the [distutils deprecation in python 3.12](https://docs.python.org/3/whatsnew/3.10.html#distutils).

fsimonis:
@gertjanvanzwieten This could be interesting to you.

PR #603, opened 2024-12-04
Perpendicular flap (IGA solid participant G+Smo)

Crazy-Rich-Meghan:
Perpendicular Flap with G+Smo (IsoGeometric Analysis) and OpenFOAM

NOTE: `Stress` instead of `Force`.

This PR is still a draft, because the G+Smo adapter's documentation is still ongoing. 

I have added a README.md file to guide users on running the tutorial with G+Smo. It includes detailed instructions for setup and execution.


perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Navigate to the directory of the preCICE tutorial (solid-gismo-elasticity).
+```
+cd <path to your precice tutorials directory>/perpendicular-flap-stress/solid-gismo-elasticity/
+```
+
+2. Create the symbolic link between the executable file and the preCICE tutorial folder
+```
+ln -sf <path to your G+Smo build folder>/bin/solid-gismo-elasticity ./solid-gismo-elasticity
+```
```
uekerman: If I remember correctly, the plan was to move these points to an adapter docs eventually, correct? This would be good.

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Runnning the G+Smo Tutorial
```
uekerman:
```suggestion
### Runnning the G+Smo Tutorial

```
see also the md linter

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+Open two terminals and run:
+```
+cd solid-gismo-elasticity
```
uekerman:
```suggestion
cd solid-gismo
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Perpendicular Flap (with IGA Solid Participant Communicating Stress Data)
```
uekerman:
```suggestion
title: Perpendicular flap with stresses
```
There is nothing specifically IGA about the setup, right?

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+keywords: G+Smo, perpendicular flap
```
uekerman:
```suggestion
keywords: G+Smo, fluid-structure interaction, FSI, OpenFOAM, IGA
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+summary: This tutorial demonstrates the isogeometric analysis (IGA) solid solver version of the “Perpendicular Flap” tutorial. It focuses on using G+Smo to handle solid-structure interactions by exchanging stress data during simulations.
```
uekerman:
```suggestion
summary: This tutorial is a modified version of the “perpendicular flap” tutorial using stresses instead of forces.
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+---
+
+
+## Overview
+
+This example demostrates how the Geometry + Simulation Modules (G+Smo) can be utilised through a module-type adapter to couple with other codes using preCICE. G+Smo offers a robust framework for isogeometric analysis (IGA), seamlessly integrating geometric representations with numerical solvers to enable advanced simulations and efficient code coupling.
```
uekerman: Let's only describe the diff of this case to the original one here. As an example, you could look at https://precice.org/tutorials-partitioned-heat-conduction-direct.html

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+1. Navigate to the directory of the preCICE tutorial (solid-gismo-elasticity).
+```
+cd <path to your precice tutorials directory>/perpendicular-flap-stress/solid-gismo-elasticity/
+```
+
+2. Create the symbolic link between the executable file and the preCICE tutorial folder
+```
+ln -sf <path to your G+Smo build folder>/bin/solid-gismo-elasticity ./solid-gismo-elasticity
```
uekerman: Couldn't we install `solid-gismo-elasticity` instead?
How is this executable related to `solid-gismo-elasticity/solid-gismo.cpp`?

perpendicular-flap-stress/solid-gismo-elasticity/clean.sh
uekerman, line 1: Compare how cleaning works in OpenFOAM. We could add a general `clean_gismo` function here: https://github.com/precice/tutorials/blob/develop/tools/cleaning-tools.sh

perpendicular-flap-stress/solid-gismo-elasticity/run.sh
uekerman, line 1: compare the OpenFOAM `run-sh` script. We could use the logging functionality from there.

uekerman (commented):
Thanks for these great contributions!
I did not yet try it out, but wanted to give some initial tips already.

Guidelines worth checking out:

- [tutorials contribution guidelines](https://precice.org/community-contribute-to-precice.html#contributing-tutorials)
- [style guide](https://precice.org/docs-meta-content-guidelines.html) of the preCICE docs.

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,76 @@
...
+mkdir build
+cd build
+cmake .. -DGISMO_OPTIONAL="<other submodules>;gsPreCICE"
+```
+
+This will trigger a download of `gsPreCICE` from GitHub. Once `gsPreCICE` is downloaded, you can compile G+Smo with `gsPreCICE`:
+
+```
+make solid-gismo-elasticity -j<number of threads to use>
+```
```
hverhelst: we can call `make install`  to install it on the system
CC  @uekerman

perpendicular-flap-stress/README.md
Crazy-Rich-Meghan: Hi! In this setup, there has no IGA-based communication being used, just the typical vertex based communication

Crazy-Rich-Meghan: I will update examples with other coupling methods under this tutorial in the future!

perpendicular-flap-stress/README.md
Crazy-Rich-Meghan: Hi @uekerman, 

We have another variant of this benchmark uses Kirichhoff-Love shell elememt and will be called `solid-gismo-shell`. While the KL shell variant has been verified for both accuracy and convergence, the implementation is currently hard-coded and needs to be generalized to support other examples.

perpendicular-flap-stress/README.md
Crazy-Rich-Meghan: I will make another PR to the adapter docs

perpendicular-flap-stress/README.md
uekerman: Could we eventually keep both in the same folder and pick them within the run.sh script?

perpendicular-flap-stress/README.md
uekerman: OK, but then let's use the title I suggested. The setup is independent of IGA. Does not mean that we could not add an IGA solver later. OK?

perpendicular-flap-stress/README.md
uekerman: This will go to the adapter docs later, right? 
I would leave this up to the user. "Make it discoverable", for example by installing it to the system.

But the tutorial can assume that the executable is discoverable.

Crazy-Rich-Meghan: Hi @uekerman, we opened a separate PR for the adapter docs: https://github.com/precice/precice.github.io/pull/482

uekerman:
I am trying to install and run the case.

I am doing:

```
git clone gismo ... & cd gismo
mkdir build & cd build
cmake .. -DGISMO_OPTIONAL="gsKLShell;gsPreCICE;gsElasticity;gsStructuralAnalysis"
make solid-gismo-elasticity
```

and get

```
make: *** No rule to make target 'solid-gismo-elasticity'.  Stop.
```

And normal `make` works and gives me `libgismo.so` as expected. 

I am a bit puzzled. How do I build the `solid-gismo.cpp` here? How does the adapter know about the available tutorials? Do I need to clone gismo into the tutorial case?

I guess what we need to do is to install the lib / make is discoverable and then add a `CMakeLists.txt` to the tutorial, which builds the actual solver?

Crazy-Rich-Meghan:
> I am trying to install and run the case.
> 
> I am doing:
> 
> ```
> git clone gismo ... & cd gismo
> mkdir build & cd build
> cmake .. -DGISMO_OPTIONAL="gsKLShell;gsPreCICE;gsElasticity;gsStructuralAnalysis"
> make solid-gismo-elasticity
> ```
> 
> and get
> 
> ```
> make: *** No rule to make target 'solid-gismo-elasticity'.  Stop.
> ```
> 
> And normal `make` works and gives me `libgismo.so` as expected.
> 
> I am a bit puzzled. How do I build the `solid-gismo.cpp` here? How does the adapter know about the available tutorials? Do I need to clone gismo into the tutorial case?
> 
> I guess what we need to do is to install the lib / make is discoverable and then add a `CMakeLists.txt` to the tutorial, which builds the actual solver?

Hi Benjamin,

Thanks for testing this! The thing is, if you try to compile the file from the G+Smo side, we are using a different name for the filename since G+Smo and preCICE have different naming rules. On the G+Smo side, the solid participant filename is `perpendicular-flap-vertex-gismo.cpp` I have also changed the filename to `perpendicular-flap-vertex-gismo` under this PR. 

Since we have multiple CPP files on our side, we cannot use a simple name like `solid-gismo-elasticity.cpp`. We have discussed this internally and decided that in the next version, we will consolidate everything into a single CPP file. Additionally, we will specify the meshes, boundary conditions, participant names, and other configurations using an XML file.

uekerman:
With a few minor changes, I was able to run the case. Most importantly, I changed the data mapping constraint for the stresses to "consistent".

I now get:

plot

The missing difference is probably due to the elements or refinement?

> Thanks for testing this! The thing is, if you try to compile the file from the G+Smo side, we are using a different name for the filename since G+Smo and preCICE have different naming rules. On the G+Smo side, the solid participant filename is perpendicular-flap-vertex-gismo.cpp I have also changed the filename to perpendicular-flap-vertex-gismo under this PR.

The file `perpendicular-flap-vertex-gismo.cpp` here is not really needed, right? I now simply copied the executable over. And the name is fine. There is no standard for names of executables. The `run.sh` only needs to know the name.

I will continue cleaning up the case the next days. Please don't force push.

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,62 @@
...
+cd fluid-openfoam
+./run.sh
+```
+
+and
+
+```bash
+cd solid-gismo
+./run.sh
+```
```
uekerman, line 46: We could explain here how different variants could be run.

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Post-processing
+
+On the OpenFOAM side, you can open the `.foam` file with ParaView, or create VTK files with `foamToVTK`.
+
+TODO: how can we visualize G+Smo results?
```
uekerman: TODO

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+As we defined a watchpoint on the 'Solid' participant at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh.` You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-gismo`. The resulting graph shows the x displacement of the flap tip. You can modify the script to plot the force instead.
+
+![Flap watchpoint](images/tutorials-perpendicular-flap-stress-displacement-watchpoint.png)
+
+TODO: add picture
```
uekerman: Let's add a picture once we are happy with the results.

tools/cleaning-tools.sh
```diff
@@ -189,10 +189,27 @@ clean_fmi() {
...
+        cd "$1"
+        echo "- Cleaning up G+Smo case in $(pwd)"
+        for ext in pvd vts vtp log txt; do
+            for file in **/*."$ext"; do
+                # Check if the file exists
+                [ -f "$file" ] && rm -f "$file"
+            done
+        done
+        rm -fv pointData.csv
+        rm -rfv ./output/
```
uekerman: This looks still a bit arbitrary to me. Currently, the output folder is empty. Shouldn't the csv go there?

perpendicular-flap-stress/README.md
uekerman: I renamed the folder. We can easily call different version from the same run script. Should we?

perpendicular-flap-stress/README.md
uekerman: Discussed with @Crazy-Rich-Meghan and @hverhelst:
The vision of the adapter is to have one configurable solver for all preCICE-coupled applications. Let's keep things as they are here and potentially modify in the future.

perpendicular-flap-stress/plot-displacement.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+
+if [ "${1:-}" = "" ]; then
+    echo "No target directory specified. Please specify the directory of the solid participant containing the watchpoint, e.g. ./plot-displacement.sh solid-fenics."
```
uekerman:
```suggestion
    echo "No target directory specified. Please specify the directory of the solid participant containing the watchpoint, e.g. ./plot-displacement.sh solid-gismo."
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Fluid participant:
+
+* OpenFOAM (pimpleFoam). In case you are using a very old OpenFOAM version, you will need to adjust the solver to `pimpleDyMFoam` in the `Fluid/system/controlDict` file. For more information, have a look at the [OpenFOAM adapter documentation](https://precice.org/adapter-openfoam-overview.html).
+
+Solid participant:
+
+* G+Smo (perpendicular-flap-vertex-gismo). For more information, have a look at the [G+Smo adapter documentation](https://precice.org/adapter-gismo-overview.html).
+
+## Running the Simulation
```
uekerman:
```suggestion
## Running the simulation
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Open two separate terminals and start the desired fluid and solid participant by calling the respective run script `run.sh` located in the participant directory. For example:
```
uekerman:
```suggestion
Open two separate terminals and start the desired fluid and solid participants by calling the respective run scripts `run.sh` located in the participants' directories. For example:
```

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Post-processing
+
+On the OpenFOAM side, you can open the `.foam` file with ParaView, or create VTK files with `foamToVTK`.
+
+On the G+Smo side, you can open the `.pvd` file located in the `solid-gismo/output` folder using ParaView. If you prefer not to plot the simulation, simply edit the `run.sh` script and remove the  `--plot` option.
+
+As we defined a watchpoint on the 'Solid' participant at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh.` You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-gismo`. The resulting graph shows the x displacement of the flap tip. You can modify the script to plot the force instead.
+
+![Flap watchpoint](images/tutorials-perpendicular-flap-stress-displacement-watchpoint.png)
```
uekerman, line 56: I think, we don't need to show all other variants here. One non-linear one (e.g. deal.II) could be enough.
And we need a brief sentence that we have not yet found the reason for the difference.

tools/cleaning-tools.sh
```diff
@@ -189,10 +189,27 @@ clean_fmi() {
...
+        set -e -u
+        cd "$1"
+        echo "- Cleaning up G+Smo case in $(pwd)"
+        for ext in pvd vts vtp log txt; do
+            for file in **/*."$ext"; do
+                # Check if the file exists
+                [ -f "$file" ] && rm -f "$file"
+            done
+        done
+        rm -fv pointData.csv
```
uekerman:
```suggestion
```
@Crazy-Rich-Meghan Only removing the output folder should be enough, right? I assume this name is currently hard-coded into the solver, right? (no problem, I just want to understand)

uekerman:
@Crazy-Rich-Meghan The results look good now :+1: What did you change?
It seems that the linear and the non-linear version of the code give (nearly?) the same results. Is this correct or a copy-paste issue?

Crazy-Rich-Meghan:
> @Crazy-Rich-Meghan The results look good now 👍 What did you change? It seems that the linear and the non-linear version of the code give (nearly?) the same results. Is this correct or a copy-paste issue?

Hi @uekerman , for the nonlinear version we iterate with Jacobian matrix, and the linear version with stiffness matrix. In principle, if the problem we are solving is a linear problem, the nonlinear time integrator should converge within the first timestep and result should be the same as time-integrator with stiffness matrix, that's why you see they are overlapping. 

And we checked the source code from dealii, in linear_elasticity there is no geometric nonlinearlity so we changed a bit from our side.

uekerman (approved):

perpendicular-flap-stress/clean-tutorial.sh
MakisH, line 1: This should be a symbolic link, not the file itself.

perpendicular-flap-stress/README.md
MakisH, line 1: Trying to run this, I get, in the beginning, on the Fluid:

```
---[precice]  Automatic RBF mapping alias from mesh "Solid-Mesh" to mesh "Fluid-Mesh" in "read" direction resolves to "partition-of-unity RBF" .
---[precice]  Computing "partition-of-unity RBF" mapping from mesh "Solid-Mesh" to mesh "Fluid-Mesh" in "read" direction.
---[precice] ERROR:  The interpolation matrix of the RBF mapping from mesh "Solid-Mesh" to mesh "Fluid-Mesh" is not invertable. This means that the mapping problem is not well-posed. Please check if your coupling meshes are correct (e.g. no vertices are duplicated) or reconfigure your basis-function (e.g. reduce the support-radius).
```

and the Solid waits with:

```
Solving timestep 0...
	It. |R|/|R0|         |dU|/|U0|        
	0   0                0                
Finished
---[precice]  relative convergence measure: relative two-norm diff of data "Displacement" = inf, limit = 5.00e-03, normalization = 0.00e+00, conv = true
---[precice]  relative convergence measure: relative two-norm diff of data "Stress" = inf, limit = 5.00e-03, normalization = 0.00e+00, conv = true
---[precice]  All converged
---[precice] WARNING:  The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
---[precice] WARNING:  The IQN matrix has no columns.
---[precice]  Time window completed
---[precice]  Mapping "Stress" for t=0.01 from "Fluid-Mesh" to "Solid-Mesh"
---[precice]  iteration: 1 of 50 (min 1), time-window: 2, time: 0.01 of 5, time-window-size: 0.01, max-time-step-size: 0.01, ongoing: yes, time-window-complete: yes, write-iteration-checkpoint 
```

Am I missing something?

perpendicular-flap-stress/README.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Perpendicular flap with stresses
+keywords: G+Smo, fluid-structure interaction, FSI, OpenFOAM
+summary: This tutorial is a modified version of the “perpendicular flap” tutorial using stresses instead of forces.
+
+---
```
MakisH:
```suggestion
---
title: Perpendicular flap with stresses
permalink: tutorials-perpendicular-flap-stresses.html
keywords: G+Smo, fluid-structure interaction, FSI, OpenFOAM
summary: This tutorial is a modified version of the “perpendicular flap” tutorial using stresses instead of forces.
---
```

The tutorials sidebar on the website also needs to be extended accordingly, to add it as a subcase of the perpendicular flap. See the flow-over-heated-plate as an example.

tools/cleaning-tools.sh
MakisH, line 1: Thanks for fixing the inconsistent indentation! Looks good.

MakisH (commented):
Thanks a lot for the contribution! Looks already very neat! I also compared the differences to the perpendicular-flap and it looks very clean. I also think that it is fine having this as a separate subcase, and I expect we will have more solvers opting for stresses instead of forces.

I could not run the case, but maybe I am doing something wrong, or something still needs to be documented. Other than that, this is in good shape already.

perpendicular-flap-stress/README.md
Crazy-Rich-Meghan: Hi Makis, 
Thanks for testing! I tested this PR with a clean clone. I did not run into this problem. 

My procedure is:
- cloned this pr
- cloned a clean G+Smo 
   - in the build folder of G+Smo, I included submodule: gsPreCICE;gsKLShell;gsElasticity;gsStructuralAnalysis
   - make perpendicular-flap-vertex-gismo
   - sudo make install perpendicular-flap-vertex-gismo
- switched back to the pr, and run the run.sh script in fluid-openfoam and solid-gismo

I am able to run this example following these steps and I could not replicate the issue here.

perpendicular-flap-stress/clean-tutorial.sh
uekerman: done

perpendicular-flap-stress/README.md
MakisH: I tried again, after pulling the latest state of this branch and of Gismo:

- Tutorials `perpendicular-flap-gismo-elasticity-stress` Mar 24 (bd0610f1174fb5658043141141b332886348b98d)
- Gismo `stable` Mar 23 (390eaf68de077ecddb5ae9f899fa88bf8a84a244)
- preCICE `develop` Mar 21 (6c1b8590c1895502443901f73bdde71a53d776f2) with `PRECICE_FEATURE_PETSC_MAPPING    ON` and `PRECICE_FEATURE_GINKGO_MAPPING   OFF`
- OpenFOAM v2406
- OpenFOAM adapter `develop` Mar 21 (fc16b305bd43f455b9ce044acadd42c56b52fdf4)
- Running on my laptop: Ubuntu 22.04, Eigen 3.4.0, PETSc 3.15.

Same error.

MakisH: @Crazy-Rich-Meghan which state of preCICE are you using? I can try checking with that, to exclude any bugs in preCICE `develop`.

uekerman: It worked for me previously, but I am now trying to mimic @MakisH 's setup.

I, however, get an error when trying to build that version of gismo:

```
Building CXX object optional/gsElasticity/CMakeFiles/gsElasticity.dir/gsBiharmonicAssembler_.cpp.o
In file included from /home/uekermbn/Repos/gismo/optional/gsElasticity/gsBaseAssembler.h:17,
                 from /home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler.h:17,
                 from /home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler_.cpp:3:
/home/uekermbn/Repos/gismo/src/gsAssembler/gsAssembler.h: In instantiation of ‘void gismo::gsAssembler<T>::apply(ElementVisitor&, size_t, gismo::boxSide) [with ElementVisitor = gismo::gsVisitorBiharmonicMixed<double>; T = double; size_t = long unsigned int]’:
/home/uekermbn/Repos/gismo/src/gsAssembler/gsAssembler.h:462:13:   required from ‘void gismo::gsAssembler<T>::push(const ElementVisitor&) [with ElementVisitor = gismo::gsVisitorBiharmonicMixed<double>; T = double]’
/home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler.hpp:93:54:   required from ‘void gismo::gsBiharmonicAssembler<T>::assemble(bool) [with T = double]’
/home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler_.cpp:8:25:   required from here
/home/uekermbn/Repos/gismo/src/gsAssembler/gsAssembler.h:721:27: error: cannot convert ‘gismo::gsBasis<double>::domainIter’ {aka ‘gismo::gsDomainIteratorWrapper<double>’} to ‘gismo::gsDomainIterator<double>&’
  721 |         visitor_.assemble(domIt, quWeights);
      |                           ^~~~~
      |                           |
      |                           gismo::gsBasis<double>::domainIter {aka gismo::gsDomainIteratorWrapper<double>}
In file included from /home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler.hpp:20,
                 from /home/uekermbn/Repos/gismo/optional/gsElasticity/gsBiharmonicAssembler_.cpp:4:
/home/uekermbn/Repos/gismo/optional/gsElasticity/gsVisitorBiharmonicMixed.h:74:48: note:   initializing argument 1 of ‘void gismo::gsVisitorBiharmonicMixed<T>::assemble(gismo::gsDomainIterator<T>&, const gismo::gsVector<T>&) [with T = double]’
   74 |     inline void assemble(gsDomainIterator<T> & element,
      |                          ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```

hverhelst: @uekerman, did you pull inside `gismo/optional/gsElasticity`?

MakisH: I did not, and I have the `master` branch there (d6ccb6f512579a077b6e549191963d6696f66a61, Feb 17).

hverhelst: @MakisH that one should be the right one that would fix the error that @uekerman encounters

uekerman: @hverhelst Thanks for the quick reaction. I wasn't aware that this is necessary. I now used a fresh clone again.

And indeed I also run into the same problem as @MakisH 

- Tutorials: same as @‌MakisH
- Gismo: same as @‌MakisH
- preCICE develop Mar 24 (5909d6548) (PETSc off), almost the same as @‌MakisH
- OpenFOAM v2212
- OpenFOAM adapter v1.3.1
- Ubuntu 20.04, Eigen 3.4.0

hverhelst: Okay, that's somehow good. I think @Crazy-Rich-Meghan is on it right now. I'll wait for her response.

uekerman: I get a similar problem with preCICE v3.1.2

```
---[precice] ERROR:  The interpolation matrix of the RBF mapping from mesh "Solid-Mesh" to mesh "Fluid-Mesh" is not invertable. This means that the mapping problem is not well-posed. Please check if your coupling meshes are correct (e.g. no vertices are duplicated) or reconfigure your basis-function (e.g. reduce the support-radius).
```

uekerman: Visualizing the coupling mesh of G+Smo looks strange. I get a volume mesh, not a surface mesh. The tip of the flap:

Screenshot from 2025-03-24 13-18-30

uekerman: And it is a 3D mesh, so duplicated vertices for a mapping in 2D, which explains the error.

@hverhelst Did you change anything in the adapter in this direction recently?

Crazy-Rich-Meghan:
Hi @MakisH and @uekerman ,

I tried to do it as Makis suggested and I arrived at the similar error. So I set back G+Smo and submodules few commits and now it worked for me:

- Main gismo (stable at 460d8dc4)
- gsElasticity (master at 5eb301c)
- gsStructuralAnalysis (main at 99ee486)
- gsPreCICE (main at 8633422 **latest**)
- preCICE version (3.1.2) 
```
grep "VERSION" /usr/lib/x86_64-linux-gnu/cmake/precice/preciceConfigVersion.cmake
# The created file sets PACKAGE_VERSION_EXACT if the current version string and
# PACKAGE_VERSION_COMPATIBLE if the current version is >= requested version,
# The variable CVF_VERSION must be set before calling configure_file().
set(PACKAGE_VERSION "3.1.2")
```
- Tutorial version (this PR at bd0610f **latest**)

And my PC is Ubuntu 24.04. @hverhelst Maybe you have an idea about what is going wrong here, could be related to one of the recent commits to G+Smo.

Crazy-Rich-Meghan:
Hi @uekerman and @MakisH,

Thanks again for your help! We @hverhelst have merged PRs to stable G+Smo and the preCICE adapter to fix the compatibility with the newer version G+Smo. Now the previous mentioned issue should be resolved

perpendicular-flap-stress/fluid-openfoam/system/controlDict
```diff
@@ -0,0 +1,45 @@
...
+
+stopAt          endTime;
+
+endTime         5;
+
+deltaT          0.01;
+
+writeControl    adjustableRunTime;
+
+writeInterval   0.1;
```
MakisH, line 24: All the configuration seems to be in the code itself, so I am writing this here:

- OpenFOAM seems to be writing results every 0.1s
- G+smo seems to writing results every 0.01s

When putting both in ParaView, this looks a bit strange, while I am wondering if it also makes the simulation slower.

Would it make sense to change the structure solver to also write every 0.1s? Or do you need this resolution for some analysis?

perpendicular-flap-stress/README.md
MakisH: With the latest state of the stable branch of the G+smo adapter (9e567), the case run fine, and the results look correct! :tada:

Note that I had to do a clean build (delete the complete `build/` directory and update all `optional/` modules).

perpendicular-flap-stress/fluid-openfoam/system/controlDict
Crazy-Rich-Meghan: > All the configuration seems to be in the code itself, so I am writing this here:
> 
> * OpenFOAM seems to be writing results every 0.1s
> * G+smo seems to writing results every 0.01s
> 
> When putting both in ParaView, this looks a bit strange, while I am wondering if it also makes the simulation slower.
> 
> Would it make sense to change the structure solver to also write every 0.1s? Or do you need this resolution for some analysis?

@MakisH I’ve updated it on our side. You can now control the output frequency using the `-m` option, which sets the modulo for plotting. By default, it’s set to 10 to match the OpenFOAM output interval (i.e., writing every 0.1s). Let me know if you’d prefer a different setting or if anything else needs adjustment.

perpendicular-flap-stress/README.md
MakisH: I think still open here is only this part:

> we need a brief sentence that we have not yet found the reason for the difference.

since deal.II still shows a small difference. Or was it larger before?

MakisH:
@Crazy-Rich-Meghan @hverhelst side question: How do you normally rebuild G+smo? Do you always need to make a clean build? Whenever I try to pull (also in the submodules) and directly run `make`, it leads to all kinds of compilation errors.

I am just trying to save a bit of time, since a clean build takes a while. Not urgent (I already rebuilt successfully), but I was wondering.

perpendicular-flap-stress/fluid-openfoam/system/controlDict
MakisH: Perfect, thank you! I ran again with the latest state, and now the results are consistent. :rocket:

MakisH (approved):
Looks perfect to me now, thanks for the quick updates! :smile: 

See also https://github.com/precice/precice.github.io/pull/516

hverhelst:
> @Crazy-Rich-Meghan @hverhelst side question: How do you normally rebuild G+smo? Do you always need to make a clean build? Whenever I try to pull (also in the submodules) and directly run `make`, it leads to all kinds of compilation errors.
> 
> I am just trying to save a bit of time, since a clean build takes a while. Not urgent (I already rebuilt successfully), but I was wondering.

@MakisH I never run a clean build. I usually pull (incl. submodules) and run a `make target` instead of a `make`. In that way I avoid compiling all examples. In addition, for small changes it will only compile the changed files, so a build should be small.
Regarding the compilation errors: what do you get? We had some breaking changes recently, most of them are resolved since today. In addition, we have some deprecated functions right now that will go out in a new release (probably June).

MakisH:
Ah, ok, I guess I stumbled upon some of the breaking changes then. I was also previously building all targets, which probably increased the range of what can go wrong. I did not log the messages, as this was clearly a dirty build. Thanks!

perpendicular-flap-stress/README.md
Crazy-Rich-Meghan: > I think still open here is only this part:
> 
> > we need a brief sentence that we have not yet found the reason for the difference.
> 
> since deal.II still shows a small difference. Or was it larger before?

The difference was larger before, we made some changes to fix it and now the gap is smaller. One possible explanation is that we may be using different assumptions regarding plane stress vs plane strain.

uekerman: let's ignore this for the moment. not worth the hassle to dig even deeper IMO

PR #606, opened 2025-01-08
Add Fortran elastic-tube-1d solvers 

YonatanGM:
This PR adds Fortran solvers for elastic-tube-1d case using the Fortran bindings (not the Fortran module). I opened another  PR #607  with the module approach if you prefer that

The solvers follow the same project structure as the other solvers. LAPACK is needed for the fluid solver (same as the C++ case). Run ./run.sh to compile and execute the solvers via CMake. The plotting scripts (plot-diameter.sh and plot-all.sh) are updated to show results for the Fortran case. The precommit yaml didnt have hooks for Fortran files, so I maunally formatted them using fprettify.  

Results look consistent with the other solvers. I checked the VTK output, and the numbers match up to 5 decimal places on the last timestep. Also tested running the fortran case against the other solves and that seems to work fine too. 

Results:
tutorials-elastic-tube-1d-all
Figure-fortran
Figure-python

Checklist:
- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.
- [ ] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.
- [x] Update the image used in readme showing diameter and pressure of running all solvers against cpp to inlcude fortran case

PR #607, opened 2025-01-08
Add Fortran elastic-tube-1d solvers (precice Fortran module)

YonatanGM:
This PR adds Fortran solvers for the elastic-tube-1d case using the fortran module (precice.f90). I also opened PR #606 that doesn't rely on this module.

The `run.sh` script downloads precice.f90 from the Fortran module repository to get the latest version and the file gets compiled with the solvers source files using CMake. Since the module is just a wrapper around the precicef functions in the library, which are updated with preCICE release, I imagine it will stay compatible with the precice version used in the solvers (as long as the function names stay consistent)

LAPACK is needed for the fluid solver (same as the C++ case). Just run ./run.sh to build and execute via CMake. The plotting scripts (plot-diameter.sh and plot-all.sh) are updated to show results for the Fortran case. The precommit yaml didnt have hooks for Fortran files, so I maunally formatted them using fprettify.  

Results look consistent with the other solvers. I checked the VTK output, and the numbers match up to 5 decimal places on the last timestep. Also tested running the fortran case against the other solves and that seems to work fine too. 

tutorials-elastic-tube-1d-all
Figure-fortran
Figure-python

Checklist:

- [x] Update the image used in readme showing diameter and pressure of running all solvers against cpp to inlcude fortran case


elastic-tube-1d/fluid-fortran/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+cmake_minimum_required(VERSION 3.10)
+project(ElasticTube LANGUAGES Fortran C)
```
MakisH: Just curious (I don't know): why is the `C` required here?

elastic-tube-1d/fluid-fortran/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+set(CMAKE_Fortran_STANDARD 2003)
```
MakisH: Does the variable `CMAKE_Fortran_STANDARD` exist or do anything?
From [this CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/25006), I understand that setting Fortran standards is not intrinsically supported by CMake at the moment.

If removed, don't forget to also apply this (and any further changes) to the `solid-fortran/CMakeLists.txt` and in the other PR.

elastic-tube-1d/fluid-fortran/run.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -e -u
+
+. ../../tools/log.sh
+exec > >(tee --append "$LOGFILE") 2>&1
+
+if [ ! -f src/precice.f90 ]; then
+  echo "Fetching precice.f90 (Module for Fortran bindings of preCICE)..."
+  curl -o src/precice.f90 https://raw.githubusercontent.com/precice/fortran-module/master/precice.f90
+fi
```
MakisH: I think this step is fine, given that it first checks if the file exists, so one could skip it (we need to document this in a `fluid-fortran/README.md`, though).

However, what was your motivation for getting the source file, instead of linking to the module, as we do in the [example](https://github.com/precice/fortran-module/blob/develop/examples/solverdummy/README.md)? It is definitely easier for the user, I am just wondering if there are further reasons.

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
```diff
@@ -0,0 +1,182 @@
...
+
+        velocity = velocity_old
+        pressure = pressure_old
+
+        ! Stabilization intensity
+        alpha = 0.0  !(N * kappa * tau) / (N * tau + 1);
+        L = 10.0
+        dx = L/kappa  !1.0 / (N * kappa);
+
+        ! result variable
```
MakisH: Since this `info` is now (in the Fortran version) far away from where it is used, we need a more descriptive comment.

I would argue that we need a better name in all versions, but let's keep it as-is now, for consistency and to keep things simple.

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
```diff
@@ -0,0 +1,182 @@
...
+        info = 0
+
+        ! Nonlinear solver loop
+        do k = 1, max_iterations
+            ! Initialize residual vector
+            Res = 0.0
+
+            ! Compute residuals
+            do i = 2, N  ! Adjusted for 1-based indexing
```
MakisH: I have not checked all the indices here, but I looked at several, and it looks consistent with the C++ version. The bounds make sense.

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
```diff
@@ -0,0 +1,182 @@
...
+
+        real(dp) :: alpha, l, dx, velocity_in, tmp2, norm_1, norm_2, norm
+
+        ! LAPACK Variables
+        integer :: nlhs, nrhs
+        real(dp), allocatable :: res(:)
+        real(dp), allocatable :: lhs(:, :)
+        integer, allocatable :: ipiv(:)
+
+        nlhs = 2*N + 2
```
MakisH: There seems to be a mix-up between `n` and `N`. Let's use one of the two forms, as at least it helps text finding. The C++ version uses `N`.

Same with `res`/`Res`, `lhs`/`LHS` and maybe more.

(Remember: in Fortran, these two are the same variable)

MakisH (commented):
Hi @YonatanGM,

wearing my preCICE hat here :cowboy_hat_face: (it is a bit unfortunate that I am also supervising your SSE project, maybe @fsimonis can control my conflict of interest).

With a first look, I think we could keep both contributions, since we support both ways of using preCICE via Fortran. Let's start with this one, and at the same time port any changes to the other one as well (should not be too many). Not sure yet how we could reduce duplication, but I have the impression the simplest would be to keep both, in different directories (`fluid-fortran` and `fluid-fortran-module`).

I had a first look and I ran the code, mixing languages to check for potential incompatibilities. I arrive at the same results.

tutorials-elastic-tube-1d-all

In terms of the solver source code, I have only reviewed the `FluidComputeSolution.f90` so far, as well as the shell scripts and `CMakeLists.txt`. I will check the rest of the Fortran files during the day.

elastic-tube-1d/fluid-fortran/src/FluidSolver.f90
```diff
@@ -0,0 +1,194 @@
...
+  use precice
+  use FluidComputeSolution, only: fluid_compute_solution
+  use Utilities, only: write_vtk
+  implicit none
+  integer, parameter :: dp = kind(1.0d0) ! Double precision
+
+  character(LEN=50) :: configFileName
+  character(LEN=50)  :: solverName
+  character(LEN=50)  :: meshName, pressureName, crossSectionLengthName
+  character(LEN=256) :: outputFilePrefix
```
MakisH: I agree that 256 would probably be a better value than 50 everywhere (while we need to eventually drop this need altogether), but what led you give a larger value here? Any particular issues?

elastic-tube-1d/fluid-fortran/src/FluidSolver.f90
```diff
@@ -0,0 +1,194 @@
...
+  ! Initialize physical parameters
+  r0 = 1.0_dp / sqrt(pi)
+  a0 = r0**2 * pi
+  u0 = 10.0_dp
+  ampl = 3.0_dp
+  frequency = 10.0_dp
+  t_shift = 0.0_dp
+  p0 = 0.0_dp
+  vel_in_0 = u0 + ampl * sin(frequency * (t_shift) * pi)
+
```
MakisH: The C++ version has some comments on what these paramteres are. Since these are parameters someone might need to change at some point, let's port the comments here as well.

elastic-tube-1d/fluid-fortran/src/utilities.f90
```diff
@@ -0,0 +1,74 @@
...
+    real(dp), intent(IN)       :: t
+    integer, intent(IN)        :: iteration
+    character(LEN=*), intent(IN) :: filenamePrefix
+    integer, intent(IN)        :: nSlices
+    real(dp), intent(IN)       :: grid(:), velocity(:), pressure(:), diameter(:)
+
+    integer :: ioUnit, i, ioStatus
+    character(LEN=256) :: filename
+
+    write (filename, '(A,"_",I0,".vtk")') trim(filenamePrefix), iteration
```
MakisH: The `filenamePrefix` was already defined as 256 characters long, and `filename` is also defined as 256 characters long. So this could potentially fail, right?

MakisH (commented):
Some of my comments were still somehow pending publishing. @YonatanGM let me know if everything is clear.

YonatanGM:
> Some of my comments were still somehow pending publishing. @YonatanGM let me know if everything is clear.

Hi @MakisH, Thanks for the comments!! I will address them later today or tomorrow.

YonatanGM:
> Hi @YonatanGM,
> 
> wearing my preCICE hat here 🤠 (it is a bit unfortunate that I am also supervising your SSE project, maybe @fsimonis can control my conflict of interest).
> 
> With a first look, I think we could keep both contributions, since we support both ways of using preCICE via Fortran. Let's start with this one, and at the same time port any changes to the other one as well (should not be too many). Not sure yet how we could reduce duplication, but I have the impression the simplest would be to keep both, in different directories (`fluid-fortran` and `fluid-fortran-module`).
> 
> I had a first look and I ran the code, mixing languages to check for potential incompatibilities. I arrive at the same results.
> 
> tutorials-elastic-tube-1d-all
> 
> In terms of the solver source code, I have only reviewed the `FluidComputeSolution.f90` so far, as well as the shell scripts and `CMakeLists.txt`. I will check the rest of the Fortran files during the day.

Hi @MakisH, I just addressed your suggestions! I renamed the folder to `fluid/solid-fortran-module`. PR #606 uses `fluid/solid-fortran`.

elastic-tube-1d/fluid-fortran/CMakeLists.txt
YonatanGM: Without the C, I get a compilation error in the find_package(precice...) call, which seems to use the FindThread cmake module and this module requires C or C++ to set up threading support. SO we need to specify C or CXX.

elastic-tube-1d/fluid-fortran/CMakeLists.txt
YonatanGM: you're right, CMake doesn't seem to support Fortran standards.  I just removed it in both PRs. 
I think we could still specify it like so: add_compile_options($<$<COMPILE_LANG_AND_ID:Fortran,GNU>:-std=f2003>)

elastic-tube-1d/fluid-fortran/run.sh
YonatanGM: Getting precice.f90 and using CMake to compile it with the solvers felt simpler. The [example](https://github.com/precice/fortran-module/tree/develop) uses two Makefiles, one to generate the .mod file and another to build the solver. We could still streamline this in the run script (by just adding the Make commands), but I preferred CMake to match the C++ case. Also, I had in mind that, CMake was more flexible because it can generate platform specific files like VS code solutions and detect compilers (the makefiles are hardcoded to use GNU fortran i think). Also don't know if makefiles are supported on windows. Let me know what you think!

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
YonatanGM: I double checked them too and they seem correct.

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
YonatanGM: I fixed this!

elastic-tube-1d/fluid-fortran/src/FluidSolver.f90
YonatanGM: Actually i meant to use 256 for the full filename. The prefix is hard coded like this: 
 ```fortran
outputFilePrefix = './output/out_fluid'  
```
So i changed it to 50 like the others.

elastic-tube-1d/fluid-fortran/src/utilities.f90
YonatanGM: Yes, this is wrong. The prefix is small and hard-coded right now, so I changed it to 50. For the full filename, I still use 256 (it's probably too much for this tutorial case, I think there's only 100 time steps / iterations but probably safe to keep it this way)

elastic-tube-1d/fluid-fortran/src/FluidComputeSolution.f90
YonatanGM: I have updated the comment here

YonatanGM:
Hi @MakisH, I addressed your suggestions! I renamed the folder to `fluid/solid-fortran-module`. PR #606 uses `fluid/solid-fortran`.
One problem is the `plot-all.sh` and `plot-diameter.sh` scripts in this PR still reference the `fluid-fortran`folder (which is not there anymore), so if you try to run those scripts they will fail. But once both PRs are merged there should no issue. The results from both PRs are the same, so I though it makes sense to just plot one of them.

YonatanGM:
@MakisH I noticed there are no precice finalize calls in the C++ and python versions. Could that be problematic?

elastic-tube-1d/fluid-fortran/CMakeLists.txt
MakisH: I only have a very limited understanding of Fortran standards, but I have heard that they work a bit differently than the C++ standards. Let's not set anything and leave the defaults (i.e., as you already have it now).

elastic-tube-1d/fluid-fortran/run.sh
MakisH: I see your point. Directly building with the same stack would be more robust, and pulling from the internet should be possible on all the relevant systems. Let's keep it like this.

elastic-tube-1d/plot-all.sh
MakisH, line 1: > One problem is the `plot-all.sh` and `plot-diameter.sh` scripts in this PR still reference the `fluid-fortran`folder (which is not there anymore), so if you try to run those scripts they will fail. But once both PRs are merged there should no issue. The results from both PRs are the same, so I though it makes sense to just plot one of them.

@YonatanGM I would suggest to treat the `solid-fortran` and the `solid-fortran-module` as two distinct variants, and plot both. Since this case is also used as a test, it makes sense to run both and compare the results. They check different components.

This means: add a `solid-fortran-module` entry in this file and in the `plot-diameter.sh`.

elastic-tube-1d/fluid-fortran-module/output/.keepme
MakisH, line 1: Is this actually needed?
This is only for the preCICE export files, right? In recent versions, the directory does not need to pre-exist.

MakisH (commented):
Thank you very much for the prompt and very clean updates. Looks already very good! Just a couple of tiny points, mostly replying to your questions.

MakisH:
> @MakisH I noticed there are no precice finalize calls in the C++ and python versions. Could that be problematic?

Good observation, thanks for asking!

In the C++ API, the `finalize()` is called in the destructor of the `Participant` class, so it is called automatically when the program exits (i.e., when the object goes out of scope). This is a good practice, which has the strange name "Resource Acquisition Is Initialization (RAII)". I assume the Python bindings implement the same.

In C and Fortran, the whole paradigm is procedural, so we need to call this explicitly.

tl;dr: All good.

MakisH:
Also: Don't forget the changelog entry (in both PRs).

MakisH:
The code works. Still pending:

- [x] ~~removing the `output/.keepme` file and the `output/` directory (I tested it, and it works without)~~ create the directory in the run script.
- [x] adding a changelog file `changelog-entries/607.md`

(putting my SSE hat on: these are trivial, but still have some learning significance)

elastic-tube-1d/fluid-fortran-module/output/.keepme
YonatanGM: @MakisH  hmm, I'm not sure actually (the cpp and python versions have that folder with the .keepme in it). 
I just removed the .keepme and pushed it but git doesnt track empty folders it seems so the output folder is now gone. 
Should we create the folder when writing the vtk outputs maybe? or in the run script ?

YonatanGM:
@MakisH With both directory and file removed, it didnt work for me.
Getting errors like this:  Error: Unable to open file ./output/out_fluid_99.vtk
I think we have to keep the folder or create it first if it doesnt exist when writing the vtk files. What do you think?

elastic-tube-1d/fluid-fortran-module/output/.keepme
MakisH: You are right, I had the impression this was the output of preCICE, but we are actually not writing any export files in this tutorial. I checked again, and while the code works, it indeed does not produce files.

Let's revert the change and add the `.keepme`, but please add a comment in the file on why this is needed (pointing to the `FluidSolver.f90`).

MakisH: Actually, creating the folder in the run script would be even better.

YonatanGM: Just committed this

MakisH (approved):
Thank you for the very clean contribution and the very prompt and clear answers, @YonatanGM! I am merging this and I will merge #606 next.

MakisH:
After merging, I noticed one small thing that I would have done differently and slipped my review: the `precice.f90` is downloaded into `src/`. I would normally put it under `thirdparty/` (or even get the repository as a git module), to keep it clearly separated.

If you want to contribute this as well, that would be great, but I can also do it later.

Issue #596, opened 2024-11-18
Reference results: Archive Python environment information


MakisH:
As suggested by @fsimonis, we can use `pip freeze` to get a tree of all dependencies used in a venv:

https://pip.pypa.io/en/stable/cli/pip_freeze/

MakisH:
A file `pip-installed-packages.log` is included in all Python-related tutorials since #608.

Missing is including this in the reference results. This, however, would break many assumptions that we currently have in what we consider as reference results and how we archive them. We currently only consider the `precice-exports/`, which are defined per tutorial, while this file is defined per case.

In general, more files and potentially the complete results would be useful to be archived as reference results for debugging purposes. But this would increase the storage requirements a lot (even if we use Git LFS, getting the repository would download multiple multi-MB files), and would require several changes in the way we archive things.

As a much easier and more useful work-around, I suggest (manually) archiving the full build artifact outside this repository. For this reason, I have now included the run number in the name of the artifacts in https://github.com/precice/tutorials/pull/628. When someone triggers the generation of new reference results, they could also archive the complete artifact (since these are retained by GitHub for up to 90 days).

I will address this again in #629, in the context of https://github.com/precice/tutorials/issues/400, update the related documentation, and add reminders. Closing for now.

PR #608, opened 2025-01-11
Python-based tutorials: Log list of installed packages

MakisH:
In the `run.sh` of our Python-based tutorial, we create a venv and install packages. This PR adds a call to `pip freeze` that writes the list of installed packages (with their versions) into a file called `pip-installed-packages.log`. This additional information can increase reproducibility.

The log file is not touched by `clean.sh`, as the respective `.venv` is also not removed by it. It is also neither a case nor a preCICE log file, to be added to the respective cleaning functions.

This is related to #596.

Checklist:

### DIFF
--- a/changelog-entries/608.md
+++ b/changelog-entries/608.md
@@ -1,0 +1,1 @@
+- Added logging of installed Python packages for the Python-based tutorials into a file `pip-installed-packages.log`.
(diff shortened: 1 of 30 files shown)


uekerman (approved):
Makes sense to me. Whether or not to clean is probably debatable, but I understand your reasoning.
I did not try it and did not check whether you caught everything.

PR #614, opened 2025-01-31
Add Fortran module as submodule (elastic-tube-1d)

YonatanGM:
In the elastic-tube-1d fortran-module case,  the fortran module file (precice.f90) is downloaded into the `src` folder of the solvers. This PR adds the [fortran-module repository](https://github.com/precice/fortran-module) as a submodule instead inside both solver to keep it more separate. 

Checklists:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "elastic-tube-1d/fluid-fortran-module/fortran-module"]
+	path = elastic-tube-1d/fluid-fortran-module/fortran-module
+	url = https://github.com/precice/fortran-module
+[submodule "elastic-tube-1d/solid-fortran-module/fortran-module"]
+	path = elastic-tube-1d/solid-fortran-module/fortran-module
+	url = https://github.com/precice/fortran-module
--- a/elastic-tube-1d/fluid-fortran-module/CMakeLists.txt
+++ b/elastic-tube-1d/fluid-fortran-module/CMakeLists.txt
@@ -16,13 +16,17 @@ endif()
 find_package(precice 3 REQUIRED CONFIG)
 find_package(LAPACK REQUIRED)
 
+# specify the path to the preCICE Fortran module
+if(NOT DEFINED PRECICE_FORTRAN_MODULE)
+  message(FATAL_ERROR "PRECICE_FORTRAN_MODULE not set. Provide it with -DPRECICE_FORTRAN_MODULE=<path>")
+endif()
+
 add_executable(FluidSolver
   src/FluidComputeSolution.f90
   src/utilities.f90
   src/FluidSolver.f90
-  src/precice.f90
+  ${PRECICE_FORTRAN_MODULE}
 )
 
 target_link_libraries(FluidSolver PRIVATE precice::precice)
 target_link_libraries(FluidSolver PRIVATE ${LAPACK_LIBRARIES} ${LAPACK_LINKER_FLAGS})
-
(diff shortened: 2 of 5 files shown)


YonatanGM:
@MakisH I am not familiar with submodules and could use your feedback on these:

- The submodule branch isn’t set, so it defaults to the latest commit on the fortran-module default branch (develop I think). Should we explicitly specify this?
-  The run scripts are doing `submodule update --init`  which only updates the submodules to the commit recorded (in the gitlink entries) in the tutorial repo. Would it be better to always get the latest commit from the submodule repo by doing something like `submodule update --init --remote`? But if the tutorial repos gitlink is outdated, wouldn't this require users to have to stash when pulling from the tutorial again because their gitlink changes (after the run script)? There are no releases for the fortran bindings so this part is confusing which state of the repo to use
- The submodule is duplicated with this approach in two places. This keeps it contained inside the solvers. Or like you said there could be a thirdparty folder in the root of the tutorial repo or elasic-tube-1d and the submodule can be put there, then we wont have to add it twice?
- Now that we have the fotran-module repo, does it make sense to just rely on the provided Makefile to create the precice.mod and use Makefiles file to build the solvers like in the dummysolver example?
- Or simply could download the precice.f90 like before but place it in a thridpary folder

MakisH:
Thanks for the contribution, @YonatanGM!

Before diving into details, I think that the submodules approach is very nice to learn how this works, but it might not be the simplest for users, as it relies on Git.

Alternatively, one could download the source code artifact: https://github.com/precice/fortran-module/archive/refs/heads/master.zip

I see we need releases, I will go ahead and make one later today, unless anyone raises concerns about it. One could then get a `.tar.gz`, which should be easier to handle.

Now to reply to your questions:

>     * The submodule branch isn’t set, so it defaults to the latest commit on the fortran-module default branch (develop I think). Should we explicitly specify this?

That should be fine for now, as we don't have releases in that repository yet. Technically, setting the branch `master` would be safer.

>     * The run scripts are doing `submodule update --init`  which only updates the submodules to the commit recorded (in the gitlink entries) in the tutorial repo. Would it be better to always get the latest commit from the submodule repo by doing something like `submodule update --init --remote`? But if the tutorial repos gitlink is outdated, wouldn't this require users to have to stash when pulling from the tutorial again because their gitlink changes (after the run script)? There are no releases for the fortran bindings so this part is confusing which state of the repo to use

We use the same approach in the website repository. One sees the modified modules in their `git status` output, but it does not conflict with any pull/push operations, as long as the user does not change anything related to the submodules.

>     * The submodule is duplicated with this approach in two places. This keeps it contained inside the solvers. Or like you said there could be a thirdparty folder in the root of the tutorial repo or elasic-tube-1d and the submodule can be put there, then we wont have to add it twice?

Two approaches:

- Get the submodule twice, in two different folders: double the downloads, which we don't really need.
- Get the submodule once in one of the fluid/solid folders, and then copy/link, after checking for existence.

Both are fine, but I think in the end we cannot assume that both `fluid-fortran-module` and `solid-fortran-module` will run. Could be a different combination (with `fluid-python`, for example).

>     * Now that we have the fotran-module repo, does it make sense to just rely on the provided Makefile to create the precice.mod and use Makefiles file to build the solvers like in the dummysolver example?

Yes, that would make sense, actually.

>     * Or simply could download the precice.f90 like before but place it in a thridpary folder

That was what I originally thought of, which should be simpler than the submodule in this case. If it was a development repository, then submodules would indeed make more sense.

The issue here is that a user might only download a release artifact of the tutorials, without using Git.

YonatanGM:
Thanks!! I understand now

MakisH:
@YonatanGM with my project maintainer hat on and completely independently from SSE: Are you planning any updates on this, or do you prefer to close it?

YonatanGM:
Hi @MakisH, I dont have any updates, Let's close it!

PR #618, opened 2025-02-18
Add default profiling explicitly in xml configs

davidscn:
To raise awareness of this option and have a starting point for potential modifications.

Checklist:

### DIFF
--- a/breaking-dam-2d/precice-config.xml
+++ b/breaking-dam-2d/precice-config.xml
@@ -7,6 +7,8 @@
       enabled="true" />
   </log>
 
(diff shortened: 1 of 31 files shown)


MakisH (approved):

davidscn:
@MakisH do you think it is a good idea to include it everywhere, e.g., also the quickstart?

MakisH:
> @MakisH do you think it is a good idea to include it everywhere, e.g., also the quickstart?

Since profiling does not make anything more complicated or more prone to dependency issues (right?), then I think consistency would be better and have it everywhere.

Thanks for asking, I did not notice you did not update the Quickstart.

PR #619, opened 2025-02-18
System tests: Update reference versions and results, upgrade default to Ubuntu 24.04

MakisH:
In the context of https://github.com/precice/precice/pull/2052#issuecomment-2665800027

Also upgrading to Ubuntu 24.04 as a base image because (1) we eventually need to do it, and (2) the Python script that updates the reference results seems to require Python 3.11. I am not sure how this worked last time.

Checklist:

- I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`. -> N/A


MakisH:
@valentin-seitz please remind me: how did this work before, but not now? It seems like the `generate_reference_results.py` requires Python 3.11, but the VM has Ubuntu 22.04, i.e., Python 3.10, and leads to the following:

```
  File "/home/precice/runners_root/actions-runner-org/_work/tutorials/tutorials/tools/tests/generate_reference_results.py", line 148, in <module>
    main()
  File "/home/precice/runners_root/actions-runner-org/_work/tutorials/tutorials/tools/tests/generate_reference_results.py", line 140, in main
    ref_results_info = render_reference_results_info(
  File "/home/precice/runners_root/actions-runner-org/_work/tutorials/tutorials/tools/tests/generate_reference_results.py", line 62, in render_reference_results_info
    'sha256': sha256sum(reference_result.path),
  File "/home/precice/runners_root/actions-runner-org/_work/tutorials/tutorials/tools/tests/generate_reference_results.py", line 57, in sha256sum
    return hashlib.file_digest(f, 'sha256').hexdigest()
AttributeError: module 'hashlib' has no attribute 'file_digest'
```

What do I forget?

MakisH:
Related job that shows the Python 3.10 issue: https://github.com/precice/tutorials/actions/runs/13421240183/job/37494139990

I see that there is also Python 3.11 installed, which should be enough. I tried enforcing that in https://github.com/precice/tutorials/pull/619/commits/62fefe500fa60ac78c560c568816087e0ec0b775, but then I started getting random tutorials failing without output. I reverted the change (which I don't seem to have needed ever before), and I removed all Docker cache from the VM, and this job output is the result.

I am tempted to upgrade the VM, but I would like first to understand why this only appears now.

MakisH:
I see that the VM is set to not allow release upgrades (I guess because it should be externally managed), so I see two ways forwards:

1. Ask for the VM to be upgraded. This might trigger issues, but better now than in 2027.
2. Downgrade the script to use older hashing methods.

I still don't understand how this worked last time, though.

MakisH:
I changed the implementation of the script to be compatible with Python 3.10, and it worked. Job output: https://github.com/precice/tutorials/actions/runs/13435728598/job/37537442258

valentin-seitz:
@MakisH This is one of the things that makes me think of the 6 stages of debugging. (where this would be the: why did it ever work stage). 
I am not at all sure what is happening. I checked the environment used by the runner and I don't see why it should have changed.

The VM has python 3.11 installed, and I remember that I implemented the hashing function in 3.11 because it was a bit more neat, and it did work at some point in time. 

But you solved it for now i guess?

MakisH:
> @MakisH This is one of the things that makes me think of the 6 stages of debugging. (where this would be the: why did it ever work stage). I am not at all sure what is happening. I checked the environment used by the runner and I don't see why it should have changed.

Could it be that some virtual environment was enabled, which was reset after a restart? I see that the uptime of the VM is only 9 days at the moment.

In the bash history, I do see a venv enabled for certbot at some point (in a directory that does not exist anymore), as well as the package `python3.11-venv` installed (from a third-party PPA).

edit: I have also updated packages (a usual `apt upgrade`), maybe this changed some setting back.

> But you solved it for now i guess?

I would still like to understand the root cause, but yes, the specific problem has been worked around for now.

Do I correctly interpret your comment as a PR approval?

PR #620, opened 2025-02-20
Use linearly increasing ramp as inflow BC

### LINKED ISSUE #612: Introduce a ramp similar to `turek-hron-fsi3` in `perpendicular-flap`?
I saw that the boundary condition in `turek-hron-fsi3` is implemented as a [cosine-shaped ramp](https://github.com/precice/tutorials/blob/e91c4b5ce8f47c4ba55e9e08f005862ae5d67079/turek-hron-fsi3/fluid-openfoam/0/U#L54-L70). Why do we not use a similar kind of boundary condition for the perpendicular flap?

In https://github.com/precice/precice/issues/2172 I studied the case `perpendicular-flap` and made the observation that not all data equals zero at the beginning with the current boundary and initial conditions.

I implemented https://github.com/precice/precice/pull/2181 to plot the watchpoints at initial time. I get non-zero forces if I set `<exchange data="Force" ... initialize="true" />`:

```
  Time  Coordinate0  Coordinate1  Displacement0  Displacement1  Force0  Force1
 0.00000000e+00   0.00000000e+00   1.00000000e+00   0.00000000e+00   0.00000000e+00   2.29427968e+00  -3.02282515e-14
```

I assume that this is due to the initial condition (IC) where the fluid velocity in x direction is homogeneously set to 10 on the whole domain. However, the cross section of the channel is smaller at the x-coordinate of the flap. Here, one would expect a larger velocity due to mass conservation which makes the given IC unphysical. A ramp should lead to a more realistic scenario.

Additionally, I realized that using a ramp also leads to better results in a subcycling-based experiment I developed for my thesis. See [here](https://github.com/BenjaminRodenberg/tutorials/blob/9b2c12682df6e99fa21dd9488d27f0a762e0f964/perpendicular-flap/fluid-openfoam/0/U#L15-L23). I will probably also change this implementation to a cosine shaped ramp since this gives us C0 and C1 continuity in contrast to the piecewise linear ramp that is only C0.

Side note: Looking at the watchpoints I would claim that setting `initialize="false"` is actually wrong since it ignores the initial forces produced on the OpenFOAM side. Using a ramp "heals" this problem since here we really have zeros for all exchanged quantities. Related issues/PRs: https://github.com/precice/precice/issues/2033, https://github.com/precice/tutorials/pull/540

BenjaminRodenberg:
Use a linearly ramp as initial condition: The inflow velocity starts for $$t=0$$ at $$u_x=0$$ and linearly increases to $$u_x=10$$ at $$t=1$$. For $$t>1$$ we use a constant inflow velocity of $$u_x=10$$.

Todos:

- [ ] Make sure that the state of the systemtests is clarified before merging #620, since it will influence the reference results.

Checklist:

### DIFF
--- a/perpendicular-flap/fluid-openfoam/system/controlDict
+++ b/perpendicular-flap/fluid-openfoam/system/controlDict
@@ -17,7 +17,7 @@ stopAt          endTime;
 
 endTime         5;
 
-deltaT          0.01;
+deltaT          0.025;
 
 writeControl    adjustableRunTime;
 
--- a/perpendicular-flap/fluid-openfoam/0/U
+++ b/perpendicular-flap/fluid-openfoam/0/U
@@ -8,14 +8,18 @@ FoamFile
 
 dimensions      [0 1 -1 0 0 0 0];
 
-internalField   uniform (10 0 0);
+internalField   uniform (0 0 0);
 
 boundaryField
 {
     inlet
     {
-        type            fixedValue;
-        value           $internalField;
+        type            uniformFixedValue;
+        uniformValue    table
+                        (
+                          (0 (0 0 0))
+                          (1 (10 0 0))
+                        );
     }
     outlet
     {


perpendicular-flap/fluid-openfoam/system/controlDict
```diff
@@ -17,7 +17,7 @@ stopAt          endTime;
 
 endTime         5;
 
-deltaT          0.01;
```
BenjaminRodenberg, line 20: Using a ramp also allows us to use a larger time step size in OpenFOAM. I did not change the overall time window size in the `precice-config.xml` since I am not sure if we will see the same effect in nutils. And I'm also not sure if there is any potential for increasing the time step size in SU2.

Would be helpful and a useful step in the scope of this PR if anybody wants to evaluate this but also not necessary, since preCICE generally supports using a smaller time step size in the config than in the solver (if everything is implemented correctly in the adapters).

BenjaminRodenberg:
I'm done with my work on this PR. Nutils is still a todo and if somebody with more experience in nutils could contribute this part this would be helpful. There is also a possible performance gain if we increase the overall time step size. But this needs careful evaluation and we could also do this as a follow up.

MakisH:
Don't forget to update the `README.md` commenting on the difference between these.
Note that we would then need to re-generate the plots at the bottom.

perpendicular-flap/fluid-openfoam/system/controlDict
bevanwsjones: Hey Benjamin, sorry this was on my feed and I just thought I would put a quick comment. Hope you well otherwise :).

I have also encountered issues with inflow conditions (back when I was still building CFD codes). Esp, for incompressible solvers and large velocities (its basically water hammer). What I found, which worked really well, was to use a sigmoid function as the ramp, as the gradients are nicer. You have 'no' discontinuity at the start or end, so the 'size/magnitude' of the final velocity does not affect the size of the jump over the starting discontinuity (time step dependent though) and it 'gives' the flow some time to react locally. When I talk of discontinuity here its a C1 discontinuity.

Anyways just my 2 cents comment.

BenjaminRodenberg: Hi @bevanwsjones, thank you for the comment and the hint with the sigmoid function! That's very helpful. If we are already touching the case it would be nice to directly use a solution that works well (even for higher-order).

I also had some trouble with C1 discontinuous inflow velocities when I tried to reach 2nd order in time for the results presented [here](https://mediatum.ub.tum.de/1462603). The fix I found was using a -cos(t)+1 instead of a sin(t) as inflow condition for the elastic-tube-1d example. I assume this is a similar case and time-stepping order should also profit from higher-order continuity here.

uekerman:
Side remark: In the QNWI paper, we used a Stokes solution as initial guess in Nutils to circumvent the continuity problem (required to reach higher order in time): https://gitlab.lrz.de/precice/ijnme2019-experiments/-/blob/master/Fluid-Structure%20Interaction/convergence/run-wi/Nutils/nsale.py?ref_type=heads#L173-176

MakisH:
@uekerman wanted to add a ramp for the Nutils case as well.

PR #631, opened 2025-03-18
Add a workflow for running the system tests nightly

### LINKED ISSUE #552: Specify default build arguments

MakisH:
This adds a workflow to run the system tests with the latest state of each involved component. The workflow is triggered every night at 4am, or manually.

I add this here and not in the preCICE repository, as this repository is the consumer of all and defines the related infrastructure. We still need some way to notify the maintainers in case of failure (this is a bit complicated, let's do it in a separate PR).

While we already have a fully-customizable, manual workflow, this allows to just click a button and test the latest state of everything. As such, we can say it closes #552.

Checklist:

### DIFF
--- a/.github/workflows/system-tests-latest-components.yml
+++ b/.github/workflows/system-tests-latest-components.yml
@@ -1,0 +1,101 @@
+name: System tests (latest components)
+
+on:
+  schedule:
+    - cron: "0 4 * * *"
+  workflow_dispatch:
+    
+jobs:
+  gather-refs:
+    name: Map Git branches to latest refs
+    if: ${{ github.event.label.name == 'trigger-system-tests' }}
+    runs-on: ubuntu-latest
+    outputs:
+      ref-precice: ${{ steps.ref-precice.outputs.shorthash }}
+      ref-python-bindings: ${{ steps.ref-python-bindings.outputs.shorthash }}
+      ref-calculix-adapter: ${{ steps.ref-calculix-adapter.outputs.shorthash }}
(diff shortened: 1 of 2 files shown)


MakisH:
@BenjaminRodenberg does this make sense to you? I have not yet tested it (because it would only work on the default branch), but it is largely copied from https://github.com/precice/precice/blob/develop/.github/workflows/system-tests.yaml

Do you think the names are reasonable?

BenjaminRodenberg (approved):
Looks good.

.github/workflows/system-tests-latest-components.yml
```diff
@@ -0,0 +1,101 @@
...
+
+on:
+  schedule:
+    - cron: "0 4 * * *"
+  workflow_dispatch:
+    
+jobs:
+  gather-refs:
+    name: Map Git branches to latest refs
+    if: ${{ github.event.label.name == 'trigger-system-tests' }}
```
MakisH, line 11: I removed these `if` conditions in https://github.com/precice/tutorials/commit/3bb4ab5d0ef6a1226390060cd62d7e488eec07ba, as the jobs were skipped. These indeed don't make sense in this context.

MakisH:
Triggered a manual job: https://github.com/precice/tutorials/actions/runs/13927405810

Let's see tomorrow if the scheduling also worked.

PR #633, opened 2025-03-24
Read initial pressure from end of time window

BenjaminRodenberg:
Potential fix for systemtests regression

Compare run: https://github.com/precice/tutorials/actions/runs/14035646677

Checklist:

### DIFF
--- a/elastic-tube-1d/solid-python/SolidSolver.py
+++ b/elastic-tube-1d/solid-python/SolidSolver.py
@@ -65,7 +65,8 @@ print("Solid: init precice...")
 interface.initialize()
 
 # Calculate initial crossSection and local pressure from initial received pressure
-pressure = interface.read_data(meshName, pressureName, vertexIDs, 0)
+precice_dt = interface.get_max_time_step_size()
+pressure = interface.read_data(meshName, pressureName, vertexIDs, precice_dt)
 
 crossSection0 = crossSection0(pressure.shape[0] - 1)
 pressure0 = p0 * np.ones_like(pressure)


BenjaminRodenberg:
@MakisH Can we run the systemtests on this branch?

I assume that due to the update of the default to `substeps="true"` there is now some interpolation going on in `interface.read_data(meshName, pressureName, vertexIDs, 0)`. 

Another potential fix is explicitly setting `substeps="false"` in the preCICE config of this case.

A long term solution would be trying to come up with better (?) initial conditions. I think at the moment it is not really clear if the initial conditions we are using here for the pressure are the correct ones. But this is a bigger topic and would be a feature for the next tutorials release.

MakisH:
Here is a custom run: https://github.com/precice/tutorials/actions/runs/14037236466

I will add the label later, I need to organize a bit a potential spaghetti that is starting to form.

MakisH:
The tests still fail, as we have now identified the QR3 PR as the starting point of the regression.

The change makes sense independently, but let's first understand or fix the QR3-related regression and then merge this one, to not have multiple moving targets.

BenjaminRodenberg:
@MakisH 100% agreement. Let's focus on https://github.com/precice/tutorials/pull/634. 

As you might have noticed: I'm always a big fan of improving initial conditions. However, if #634 fixes the regression there is not really a good argument why the solution I provide here is superior to the one we already have.

MakisH:
@BenjaminRodenberg given that #634 is now merged and fixes the regression, do you still want this one in?

BenjaminRodenberg:
@MakisH no. This PR is not required.

PR #634, opened 2025-03-24
Enforce recomputation of preconditioner to avoid regression

uekerman:
This could avoid the regression originatting from https://github.com/precice/precice/pull/1493.
Makes sense to for the perpendicular flap, but I don't yet understand the 1D elastic tube, as there, a preconditioner should make no difference (one data field in acc only). 

### DIFF
--- a/changelog-entries/634.md
+++ b/changelog-entries/634.md
@@ -1,0 +1,1 @@
+- Changed the IQN settings in the elastic-tube-1d and perpendicular flap tutorials to enforce recomputation of the preconditioner [#634](https://github.com/precice/tutorials/pull/634)
--- a/elastic-tube-1d/precice-config.xml
+++ b/elastic-tube-1d/precice-config.xml
@@ -63,6 +63,7 @@
     <acceleration:IQN-ILS>
       <data name="CrossSectionLength" mesh="Solid-Nodes-Mesh" />
       <initial-relaxation value="0.01" />
+      <preconditioner type="residual-sum" update-on-threshold="false" />
       <max-used-iterations value="50" />
       <time-windows-reused value="8" />
       <filter type="QR2" limit="1e-3" />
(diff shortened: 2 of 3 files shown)


MakisH:
Here you are: https://github.com/precice/tutorials/actions/runs/14039455941

This seems to also fix the elastic-tube-1d.

uekerman:
> This seems to also fix the elastic-tube-1d.

@Fujikawas Any idea why? Any side effect I am overseeing?

@MakisH Feel free to merge here already to move things forward.

MakisH (approved):
The system tests now pass.
I added a changelog entry.

@uekerman @Fujikawas let's revert this in case it becomes not necessary anymore in the future, to keep the settings simpler.

uekerman:
> let's revert this in case it becomes not necessary anymore in the future, to keep the settings simpler.

yes, we should change to the default everywhere actually

```xml
<acceleration:IQN-ILS>
  <data name="CrossSectionLength" mesh="Solid-Nodes-Mesh" />
</acceleration:IQN-ILS>
```

davidscn:
I don't understand the rationale here: disabling a feature we added to preCICE (for performance reasons in this case) and then disabling it manually to make the tests happy?! Updating the tests would be more convenient in my opinion.

MakisH:
I guess what we don't yet understand is how different the results are and whether we accept them or not. To evaluate the extent of the regression, you might find the `precice-exports/diff_*.vtu` files in the tests archive useful.

If the new default is ok, I would also prefer updating the reference results than making every tutorial more complicated.

uekerman:
My idea was to do this in two steps:
1. test and release preCICE. easier without regressions.
2. update tutorials and reference data

MakisH:
I understand that, since https://github.com/precice/precice/pull/2254, we can revert this. Edit: actually, no: different issue.

One extra reason to remove these tweaks: this made our tutorials non-backwards compatible. Example from v3.1.2:

```
---[precice] ERROR: The tag <preconditioner> in the configuration contains an unknown attribute "update-on-threshold". Expected attributes are type, freeze-after.
```

PR #638, opened 2025-03-25
Enable running the system tests in forks

### LINKED ISSUE #580: Checking out from forks in the system tests
We currently require that the commits/tags we are testing are all in the original, specified repositories. This is clearly restricting, and we could extend the system tests infrastructure to cover that as well.

MakisH:
Forwards the PR number to the `systemtests.py`, which then forwards it to the `Dockerfile`, allowing a `git fetch origin pull/<pr>/head`.

In the `Systemtest.py` class, I had to remove a strict check (that every argument without a default value is provided), to keep the implementation relatively straight-forward.

Closes #580.

Checklist:

### DIFF
--- a/.github/workflows/system-tests-pr.yml
+++ b/.github/workflows/system-tests-pr.yml
@@ -109,6 +109,7 @@ jobs:
           echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
           echo "### Tutorials" >> $GITHUB_STEP_SUMMARY
           echo "Reference (pull request): \`${{ github.event.pull_request.head.sha }}\`" >> $GITHUB_STEP_SUMMARY
+          echo "Pull request number: [${{ github.event.number }}](https://github.com/precice/tutorials/pull/${{ github.event.number }})" >> $GITHUB_STEP_SUMMARY
 
   run-system-tests:
     name: Trigger system tests
@@ -127,6 +128,7 @@ jobs:
         OPENFOAM_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-openfoam-adapter }},\
         SU2_VERSION:7.5.1,\
         SU2_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-su2-adapter }},\
+        TUTORIALS_PR:${{ github.event.number }},\
         TUTORIALS_REF:${{ github.event.pull_request.head.sha }}"
-      system_tests_branch: develop
-      log_level: "INFO"
+      system_tests_branch: test-forks
+      log_level: "DEBUG"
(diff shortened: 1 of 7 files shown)


.github/workflows/system-tests-pr.yml
```diff
@@ -127,6 +128,7 @@ jobs:
         OPENFOAM_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-openfoam-adapter }},\
         SU2_VERSION:7.5.1,\
         SU2_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-su2-adapter }},\
+        TUTORIALS_PR:${{ github.event.number }},\
         TUTORIALS_REF:${{ github.event.pull_request.head.sha }}"
-      system_tests_branch: develop
-      log_level: "INFO"
\ No newline at end of file
+      system_tests_branch: test-forks
+      log_level: "DEBUG"
```
MakisH: Remember to revert these before merging.

.github/workflows/system-tests-pr.yml
```diff
@@ -127,6 +128,7 @@ jobs:
         OPENFOAM_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-openfoam-adapter }},\
         SU2_VERSION:7.5.1,\
         SU2_ADAPTER_REF:${{ needs.gather-refs.outputs.ref-su2-adapter }},\
+        TUTORIALS_PR:${{ github.event.number }},\
```
MakisH, line 131: This line needs to be ported to the workflows of other repositories (replacing the `TUTORIALS_PR` with, e.g., `PRECICE_PR`.

tools/tests/components.yaml
MakisH, line 1: The changes in this file are needed because we currently have a filtering step that restricts the `build_args` only to the components actually used. I am not sure if this filtering step helps or complicates the overall implementation.

tools/tests/dockerfiles/ubuntu_2404/Dockerfile
```diff
@@ -45,12 +45,14 @@ USER precice
...
 FROM precice_dependecies AS precice
 # Install & build precice into /home/precice/precice
+ARG PRECICE_PR
 ARG PRECICE_REF
 ARG PRECICE_PRESET
 USER precice
 WORKDIR /home/precice
 RUN git clone https://github.com/precice/precice.git precice && \
     cd precice && \
+    if [ -n "${PRECICE_PR}" ]; then git fetch origin pull/${PRECICE_PR}/head; fi && \
```
MakisH, line 55: The idea here is that we always use the same Dockerfile, but if a `PRECICE_PR` is provided (i.e., it is a string of non-zero length), then we first fetch.

tools/tests/components.yaml
MakisH: I removed the exception from the filtering step, which was checking if all needed arguments were provided. This allows to more easily pass additional arguments, without having to provide defaults for them. If we want such a strict check for some reason later on, we should rework the assumptions.

MakisH:
As I noticed some regressions when trying to run the system tests on this PR, here is a recent run that does not include these changes:

https://github.com/precice/tutorials/actions/runs/14088294869/job/39458261691

MakisH:
Succeeding job with the latest changes: https://github.com/precice/tutorials/actions/runs/14112641301/job/39537745120

I will now revert the debugging-related changes in the workflow.

PR #640, opened 2025-03-28
Revert #634 and update reference results

MakisH:
After releasing preCICE v3.2.0, we should update the reference results for the system tests and revert the changes introduced by #634. Besides the performance improvements we are missing, the configuration parameter `<preconditioner update-on-threshold="<bool>" />` is not available pre-v3.2.0.

Checklist:

- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`. -> Removed the one from #634.
- ~~Update the reference results in the same PR.~~ -> will do separately


uekerman:
If we touch this, I would suggest to directly switch to the default everywhere, e.g.

```xml
<acceleration:IQN-ILS>
  <data name="Displacement" mesh="Solid-Mesh" />
  <data name="Force" mesh="Solid-Mesh" />
</acceleration:IQN-ILS>
```

MakisH:
Since we have now released v3.2.0, we can merge this.
I can update the reference results and check for more instances.

MakisH:
> If we touch this, I would suggest to directly switch to the default everywhere, e.g.
> 
> ```
> <acceleration:IQN-ILS>
>   <data name="Displacement" mesh="Solid-Mesh" />
>   <data name="Force" mesh="Solid-Mesh" />
> </acceleration:IQN-ILS>
> ```

Reading the discussion again, I understand that with `everywhere` you mean in all tags (not all tutorials), so:

```diff
    <acceleration:IQN-ILS>
      <data name="Displacement" mesh="Solid-Mesh" />
      <data name="Force" mesh="Solid-Mesh" />
-      <preconditioner type="residual-sum" update-on-threshold="false" />
-      <filter type="QR2" limit="1e-2" />
-      <initial-relaxation value="0.5" />
-      <max-used-iterations value="100" />
-      <time-windows-reused value="15" />
    </acceleration:IQN-ILS>
```

I would rather do this in a separate PR and consider it for more tutorials. I opened an issue for that: https://github.com/precice/tutorials/issues/765

MakisH:
Status: I cannot update the reference results in this branch since it is in my fork. There are, anyway, a few more PRs that will need generating some new reference results, so I will bundle them all together.

PR #641, opened 2025-03-28
Exports: Migrate from vtk to vtu

MakisH:
Prevents the following error when running in parallel:

```
---[precice] ERROR:  You attempted to use the legacy VTK exporter with the parallel participant Solid, which isn't supported.Migrate to another exporter, such as the VTU exporter by specifying "<export:vtu ... />"  instead of "<export:vtk ... />".
```

Not sure if we can already change this in the system tests Docker Compose file as well. It currently just works.


fsimonis (approved):
Yes, let's make this future-proof for potentially parallel cases.

Does this affect the Quickstart documentation?

MakisH:
> Does this affect the Quickstart documentation?

It currently states `Load these VTK files in ParaView`, and `.vtu` is a VTK format, so no. But good suggestion :+1:

PR #642, opened 2025-04-02
OpenFOAM: Strictly require the preciceAdapterFunctionObject

MakisH:
Asks OpenFOAM to abort in case an error is raised inside the adapter, or in case the adapter cannot be loaded (most common source of confusion). Previously, the simulation would continue without the adapter/coupling. Now:

```
Starting time loop

forces forces:
    rho: rhoInf
    Freestream density (rhoInf) set to 10
    Not including porosity effects

--> FOAM Warning : 
    Could not load "libpreciceAdapterFunctionObject.so"
libpreciceAdapterFunctionObject.so: cannot open shared object file: No such file or directory


--> FOAM FATAL ERROR: (openfoam-2412)
Unknown function type preciceAdapterFunctionObject

Valid function types :

15
(
...
)



    From static Foam::autoPtr<Foam::functionObject> Foam::functionObject::New(const Foam::word&, const Foam::Time&, const Foam::dictionary&)
    in file db/functionObjects/functionObject/functionObject.C at line 129.

FOAM exiting
```

**Future reader:** In case you get this error, check for issues with the OpenFOAM adapter / preCICE / dependencies installation. Search or ask in the [forum](https://precice.discourse.group/).

Option [available since v2012](https://www.openfoam.com/news/main-news/openfoam-v20-12/post-processing#post-processing-function-object-error-handling), already long enough. Related discussion in the [OpenFOAM GitLab](https://develop.openfoam.com/Development/openfoam/-/issues/1779#note_57407).

I have added a comment, so that users of other versions can get a hint to remove it.

Pending PR to the adapter, to update the documentation and any error-handling workarounds.

Thanks to @fsimonis for asking again for a solution to this.

@vidulejs: Do you see any reason not to merge this?

Checklist:

### DIFF
--- a/breaking-dam-2d/fluid-openfoam/system/controlDict
+++ b/breaking-dam-2d/fluid-openfoam/system/controlDict
@@ -43,12 +43,12 @@ maxCo           2;
 maxAlphaCo      1;
 maxDeltaT       1;
 
+libs ("libpreciceAdapterFunctionObject.so");
 functions
 {
     preCICE_Adapter
     {
         type preciceAdapterFunctionObject;
-        libs ("libpreciceAdapterFunctionObject.so");
         errors strict; // Available since OpenFOAM v2012
     }
 }
--- a/breaking-dam-2d/fluid-openfoam/system/controlDict
+++ b/breaking-dam-2d/fluid-openfoam/system/controlDict
@@ -49,5 +49,6 @@ functions
     {
         type preciceAdapterFunctionObject;
         libs ("libpreciceAdapterFunctionObject.so");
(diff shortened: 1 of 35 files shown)


MakisH:
I now also moved the `libs(...)` list outside the `functions{...}` dictionary. This is required in some cases (for custom boundary conditions in the partitioned flow cases), but we should probably be consistent everywhere.

fsimonis (approved):

vidulejs (approved):

vidulejs (approved):

PR #643, opened 2025-04-04
FSI + kOmegaSST turbulence model

vidulejs:
As we agreed, this is the pull request where we can discuss the issue with the Poisson wall distance calculation and eventually create a new turbulent FSI tutorial. This case would have to be modified to have a higher Re number.

@MakisH 
@KariimAhmed 

Checklist:


MakisH:
Thanks for pushing this! @KariimAhmed could you please try to adjust this case to reproduce your issue? It would then be much easier for us to fix.

After we have a fix, we should indeed convert this to a new `perpendicular-flap-turbulent` tutorial.

@vidulejs I know this was meant to be a complete draft. I just went ahead changed some formatting things to make this already easier to review and transform into a tutorial, so that you can focus on the actual issue. Please pull from your branch.

vidulejs:
Thanks for fixing the formatting @MakisH! 

Just to re-iterate from our conversation, we want to create a fully turbulent FSI test case to see if an adapter-related issue appears. We should start with increasing the velocity, to achieve a higher Reynolds number. The numerical schemes will probably need updating, because the turbulent energy dissipation rate `omega` was exploding. 

I noticed I accidentally pushed the wrong `wallDist` method, but during our meeting I had the `Poisson` method and the case was running succesfully.

KariimAhmed:
Hello Daniels , Is there any updates regarding the test case ?

vidulejs:
Hi Karim!

I've updated the test case and it's working now. I increased the velocity so that the turbulence is more pronounced. K and Omega were unstable so I refined the mesh near the flap and added relaxation factors 0.4. You can see all my changes in the commits.

image

Let's try to replicate your error!

To start, you could post your latest error logs here. 

We had a guess that the error is related to the `Poisson` `wallDist` method. Currently the test case is running succesfully with this method.  

To replicate your issue, we need to match the test case conditions to your actual case. What do you see is missing from this test case that is in your case? Please suggest changes. You can pull this branch and make edits, but you will not be able to push them to this branch.

MakisH:
This PR is only meant to investigate an issue that @KariimAhmed faced. Current status: we cannot reproduce the issue, we need more input.

MakisH:
Since this PR has not been going anywhere since a while, I am closing it. @vidulejs feel free to transfer any lessons learned to other PRs.

PR #645, opened 2025-04-11
Add a cleanup workflow for the system tests runner

MakisH:
We run the system tests on the same runner to take advantage of the intrinsic caching of Docker, but we also often get issues because of this, as we have no cleanup step. These issues are mainly:

- Running out of networks
- Running out of storage space

The manual solution so far is that I log into the runner and run a `docker system prune -a` every time this happens. This PR adds a workflow that performs this step, so that:

- Others can easily trigger it, without having to connect to the runner
- It runs periodically, unsupervised

After each run, the caching effects are gone, so jobs should take longer. I set it to run once per month (1st day of the month, at 3am, so that it is before any scheduled run of the system tests). We might need to make it more frequent, but this felt like an easy to remember frequency.

Related to https://github.com/precice/infrastructure/issues/11.

### DIFF
--- a/.github/workflows/system-tests-cleanup.yml
+++ b/.github/workflows/system-tests-cleanup.yml
@@ -10,6 +10,12 @@ jobs:
     name: Cleanup system tests runner
     runs-on: [self-hosted, linux, x64, precice-tests-vm]
     steps:
+      - name: Report disk usage before cleanup
+        run: |
+          df -h | grep /var
       - name: Run docker system prune
         run: |
           docker system prune --all --force
+      - name: Report disk usage after cleanup
+        run: |
+          df -h | grep /var
--- a/.github/workflows/system-tests-cleanup.yml
+++ b/.github/workflows/system-tests-cleanup.yml
@@ -12,10 +12,10 @@ jobs:
     steps:
       - name: Report disk usage before cleanup
         run: |
-          df -h | grep /var
+          df -h | grep -E '(Filesystem|/var)'
       - name: Run docker system prune
         run: |
           docker system prune --all --force
       - name: Report disk usage after cleanup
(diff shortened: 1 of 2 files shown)


.github/workflows/system-tests-cleanup.yml
```diff
@@ -0,0 +1,21 @@
...
+ cleanup-system-tests-runner:
+    name: Cleanup system tests runner
+    runs-on: [self-hosted, linux, x64, precice-tests-vm]
+    steps:
+      - name: Report disk usage before cleanup
+        run: |
+          df -h | grep -E '(Filesystem|/var)'
+      - name: Run docker system prune
+        run: |
+          docker system prune --all --force
```
fsimonis, line 18: You could also keep the latest n images with something like this:

```
docker images --format "{{.ID}} {{.CreatedAt}}" | sort -k2,3 | head -n-5 | cut -f1 -d\ | xargs -r docker rmi
```

1. Get ID and creation date of images
2. Sort by date and time
3. Select all images except the 5 newest ones
4. Select only the ID
5. Pass to docker rmi if there is at least one input `-r`

MakisH: That's a good suggestion as well, to have more fine-grained cleanup. We would then also need to cleanup networks and volumes individually.

I wonder if it is actually good to have cache invalidation every now and then. I think in the past we had some regressions (due to dependencies) that mainly showed up once I cleared the caches.

I am a bit hesitant opening a box for a more elaborate solution without having a simpler solution first. How about getting this merged first, try it out, and then improve?

MakisH:
Merging to have a first working solution, we can explore alternative approaches later.

PR #648, opened 2025-04-25
Add Nutils participants for turek-hron-fsi3

gertjanvanzwieten:
This PR adds two Nutils participants for the turek-hron-fsi3 tutorial. The scripts are intentionally structured similar to our monolithic example so they can be compared side by side with https://examples.nutils.org/official-turek/.

Some notes:
- The scripts are only tested in mutual coupling, so I don't know how they hold up with other participants. In particular I hope that traction is exchanged with the correct sign.
- The script requires Nutils 9, which is the currently stable version.
- The time step is set equal to the max_time_step_size, rather than the script having its own notion of a preferred time step. I'm not sure if that is idiomatic.

I hope you will find this a useful addition, let me know if anything needs changing!

### DIFF
--- a/turek-hron-fsi3/README.md
+++ b/turek-hron-fsi3/README.md
@@ -59,19 +59,22 @@ In the first few timesteps, many coupling iterations are required for convergenc
 
 ## Post-processing
 
-You can visualize the results of the coupled simulation using e.g. ParaView. OpenFOAM uses the specific format `fluid-openfoam.foam`, which you directly load in Paraview or convert to VTK with `foamToVTK`. deal.II writes VTK files. Both Nutils solvers currently don not write VTK files (but their own in-situ visualization), but this can be added easily similarly to the [perpendicular flap solvers](https://github.com/precice/tutorials/blob/98a78fe2dc2f6c5d84b2b30d35d00352782236f8/perpendicular-flap/fluid-nutils/fluid.py#L227).
+You can visualize the results of the coupled simulation using e.g. ParaView. OpenFOAM uses an OpenFOAM-specific format, and you can directly load the (empty) file `fluid-openfoam.foam` in Paraview or convert the results to VTK with `foamToVTK`. deal.II writes VTK files. Both Nutils solvers currently do not write VTK files (but use their own in-situ visualization), but this can be easily added similarly to the [perpendicular flap solvers](https://github.com/precice/tutorials/blob/98a78fe2dc2f6c5d84b2b30d35d00352782236f8/perpendicular-flap/fluid-nutils/fluid.py#L227).
 
-If you want to visualize both domains with ParaView, keep in mind that the different solvers may write results with different output frequencies.
+If you want to visualize both domains with ParaView, keep in mind that the different solvers may write results with different output frequencies, which you might want to [synchronize](https://precice.org/tutorials-visualization.html#synchronizing-results).
 
-There is an [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. At the end of `run.sh`, we call `openfoam_remove_empty_dirs` (provided by `tools/openfoam-remove-empty-dirs`) to delete the additional files before importing the results in ParaView.
+There is a [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. At the end of `run.sh`, we call `openfoam_remove_empty_dirs` (provided by `tools/openfoam-remove-empty-dirs`) to delete the additional files before importing the results in ParaView.
 
-Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`. You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-dealii`. The resulting graph shows the vertical (y) displacement of the tip of the flap.
+Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`, which expects the directory of the selected solid participant as a command line argument. For example:
+
+ ```shell
+ plot-displacement.sh solid-dealii
 
 ![FSI3 watchpoint](images/tutorials-turek-hron-fsi3-tip-plot.png)
 
 Before running the simulation again, you may want to cleanup any result files using the script `clean-tutorial.sh`.
 
-## Mesh refinement for OpenFOAM
+## Mesh refinement
 
 In `fluid-openfoam/system/`, we provide three different fluid meshes:
 
--- a/turek-hron-fsi3/README.md
+++ b/turek-hron-fsi3/README.md
@@ -1,8 +1,8 @@
 ---
 title: Turek Hron FSI3
 permalink: tutorials-turek-hron-fsi3.html
-keywords: OpenFOAM, deal.II, verification
-summary: The Turek-Hron FSI cases are well-established numerical benchmarks and, therefore, well suited for verification of preCICE itself and the used adapters. In this tutorial, we focus on the FSI3 case, which presents the most challenging case in terms of added mass. Please note that the meshes of this case are significantly finer than for other tutorials. Running the simulation might take a few hours. We do not recommend to run this tutorials as your first preCICE tutorial.  
+keywords: OpenFOAM, deal.II, Nutils, verification
+summary: The Turek-Hron FSI cases are well-established numerical benchmarks and, therefore, well suited for verification of preCICE itself and the used adapters. In this tutorial, we focus on the FSI3 case, which presents the most challenging case in terms of added mass. Please note that the meshes of this case are significantly finer than for other tutorials. Running the simulation might take a few hours. We do not recommend to run this tutorials as your first preCICE tutorial.
 ---
 
 {% note %}
@@ -28,14 +28,16 @@ preCICE configuration (image generated using the [precice-config-visualizer](htt
 Fluid participant:
 
 * OpenFOAM (pimpleFoam). In case you are using a very old OpenFOAM version, you will need to adjust the solver to `pimpleDyMFoam` in the `Fluid/system/controlDict` file. For more information, have a look at the [OpenFOAM adapter documentation](https://precice.org/adapter-openfoam-overview.html).
+* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v9.0. This case takes significantly longer to run than the OpenFOAM case, see [related issue](https://github.com/precice/tutorials/issues/506).
 
 Solid participant:
 
 * deal.II. For more information, have a look at the [deal.II adapter documentation](https://precice.org/adapter-dealii-overview.html). This tutorial requires the nonlinear solid solver. Please copy the nonlinear solver executable to the `solid-dealii` folder or make it discoverable at runtime and update the `solid-dealii/run.sh` script.
+* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v9.0.
 
 ## Running the Simulation
 
-Open two separate terminals and start each participant by calling the respective run script.
+Open two separate terminals and start each participant by calling the respective run script. For example:
 
 ```bash
 cd fluid-openfoam
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
 
 ## Post-processing
 
-You can visualize the results of the coupled simulation using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `fluid-openfoam.foam` file. Solid results are in VTK format.
+You can visualize the results of the coupled simulation using e.g. ParaView. OpenFOAM uses the specific format `fluid-openfoam.foam`, which you directly load in Paraview or convert to VTK with `foamToVTK`. deal.II writes VTK files. Both Nutils solvers currently don not write VTK files (but their own in-situ visualization), but this can be added easily similarly to the [perpendicular flap solvers](https://github.com/precice/tutorials/blob/98a78fe2dc2f6c5d84b2b30d35d00352782236f8/perpendicular-flap/fluid-nutils/fluid.py#L227).
 
-If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`).
+If you want to visualize both domains with ParaView, keep in mind that the different solvers may write results with different output frequencies.
 
 There is an [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. At the end of `run.sh`, we call `openfoam_remove_empty_dirs` (provided by `tools/openfoam-remove-empty-dirs`) to delete the additional files before importing the results in ParaView.
 
-Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`.  The resulting graph shows the vertical (y) displacement of the tip of the flap.
+Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`. You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-dealii`. The resulting graph shows the vertical (y) displacement of the tip of the flap.
 
 ![FSI3 watchpoint](images/tutorials-turek-hron-fsi3-tip-plot.png)
 
 Before running the simulation again, you may want to cleanup any result files using the script `clean-tutorial.sh`.
 
-## Mesh refinement
+## Mesh refinement for OpenFOAM
 
 In `fluid-openfoam/system/`, we provide three different fluid meshes:
 
@@ -83,15 +85,6 @@ If you want to use one of the two refined meshes, simply swap the `blockMeshDict
 mv blockMeshDict blockMeshDict_original
 mv blockMeshDict_refined blockMeshDict
 ```
-
-For the double-refined mesh, it is wisely to use local basis functions in the RBF data mapping method instead of global ones. You can use:
-
-```xml
-<mapping:rbf-compact-tps-c2 direction="read" from="Fluid-Mesh-Centers" to="Solid-Mesh"
-                            support-radius="0.011" constraint="consistent" />
-```
-
-You can find more information on RBF data mapping in the [documentation](https://precice.org/configuration-mapping.html#radial-basis-function-mapping).
 
 ## References
 
--- a/turek-hron-fsi3/README.md
+++ b/turek-hron-fsi3/README.md
@@ -89,6 +89,11 @@ mv blockMeshDict blockMeshDict_original
 mv blockMeshDict_refined blockMeshDict
 ```
 
+## Acknowledgements
+
+Thanks to the Technical University of Eindhoven for funding the development of
+the Nutils participants for this tutorial.
+
 ## References
 
 [1]  S. Turek, J. Hron, M. Madlik, M. Razzaq, H. Wobker, and J. Acker. Numerical simulation and benchmarking of a monolithic multigrid solver for fluid-structure interaction problems with application to hemodynamics. In H.-J. Bungartz, M. Mehl, and M. Schäfer, editors, Fluid Structure Interaction II: Modelling, Simulation, Optimization, page 432. Springer Berlin Heidelberg, 2010.
--- a/turek-hron-fsi3/fluid-nutils/fluid.py
+++ b/turek-hron-fsi3/fluid-nutils/fluid.py
@@ -217,16 +217,16 @@ def main(domain: Domain = Domain(), fluid: Fluid = Fluid(), dynamic: Dynamic = D
     sys_up = System(res, trial='u,p', test='utest,ptest')
 
     # t = -σ·n => ∀q: ∮ -q·t = ∮ q·σ·n = ∫ div(q·σ) = ∫ (∇q:σ + q·div σ) = ∫ (∇q:σ + ρf q·DuDt)
-    ns.t = topo.field('t', btype='std', degree=2, shape=[
-                      2]) * (fluid.viscosity * fluid.velocity / domain.cylinder_radius)
+    ns.traction = topo.field('traction', btype='std', degree=2, shape=[
+        2]) * (fluid.viscosity * fluid.velocity / domain.cylinder_radius)
     res += topo.boundary['cylinder,structure'].integral(
-        'utest_i t_i dS' @ ns, degree=4)
-    sys_t = System(res, trial='t', test='utest')
+        'utest_i traction_i dS' @ ns, degree=4)
+    sys_t = System(res, trial='traction', test='utest')
     sqr = topo.boundary['cylinder,structure'].integral(
-        't_i t_i' @ ns, degree=4) / 'Pa2'
-    cons['t'] = numpy.isnan(
-        System(sqr, trial='t').solve_constraints(droptol=1e-10)['t'])
-    F = topo.boundary['cylinder,structure'].integral('t_i dS' @ ns, degree=4)
+        'traction_i traction_i' @ ns, degree=4) / 'Pa2'
+    cons['traction'] = numpy.isnan(
+        System(sqr, trial='traction').solve_constraints(droptol=1e-10)['traction'])
+    F = topo.boundary['cylinder,structure'].integral('traction_i dS' @ ns, degree=4)
 
     # mesh continuation with jacobian based stiffness
     sqr = topo.integral('C_kk - 2 log(J)' @ ns, degree=4)
@@ -268,7 +268,7 @@ def main(domain: Domain = Domain(), fluid: Fluid = Fluid(), dynamic: Dynamic = D
                                tol=1e-10)  # project traction
 
             participant.write_data(
-                w_name, 'Stress', w_ids, w_sample.eval(ns.t, args) / precice_stress)
+                w_name, 'Stress', w_ids, w_sample.eval(ns.traction, args) / precice_stress)
             participant.advance(timestep)
 
             if participant.requires_reading_checkpoint():
--- a/turek-hron-fsi3/fluid-nutils/fluid.py
+++ b/turek-hron-fsi3/fluid-nutils/fluid.py
@@ -72,7 +72,7 @@ class Domain:
 class Fluid:
     density: Density = Density('1kg/L')
     viscosity: Viscosity = Viscosity('1Pa*s')
-    velocity: Velocity = Velocity('1m/s')
+    velocity: Velocity = Velocity('2m/s')
 
     def reynolds(self, reference_length):
         return self.density * self.velocity * reference_length / self.viscosity
--- a/turek-hron-fsi3/fluid-nutils/fluid.py
+++ b/turek-hron-fsi3/fluid-nutils/fluid.py
@@ -1,0 +1,321 @@
+# Turek benchmark fluid solver, modified from https://examples.nutils.org/official-turek/
+
+from precice import Participant
+from nutils import cli, export, function
+from nutils.mesh import gmsh
+from nutils.solver import System
+from nutils.SI import Length, Density, Viscosity, Velocity, Time, Stress, Acceleration
+from nutils.expression_v2 import Namespace
+from collections import defaultdict, deque
+from dataclasses import dataclass
+from typing import Optional
+from pathlib import Path
+import treelog as log
+import numpy
+
+
+# reference lenghts used in communication
+precice_length = Length('m')
+precice_time = Time('s')
+precice_stress = Stress('Pa')
+
+
+@dataclass
+class Domain:
+    channel_length: Length = Length('2.5m')
+    channel_height: Length = Length('.41m')
+    x_center: Length = Length('.2m')
+    y_center: Length = Length('.2m')
+    cylinder_radius: Length = Length('5cm')
+    structure_length: Length = Length('35cm')
+    structure_thickness: Length = Length('2cm')
+    min_elemsize: Length = Length('4mm')
+    max_elemsize: Length = Length('4cm')
+
+    def generate_mesh(self):
+        u = Length('m')  # temporary reference length
+
+        topo, geom = gmsh(Path(__file__).parent / 'fluid.geo', dimension=2, order=2, numbers={
+            'channel_length': self.channel_length / u,
+            'channel_height': self.channel_height / u,
+            'x_center': self.x_center / u,
+            'y_center': self.y_center / u,
+            'cylinder_radius': self.cylinder_radius / u,
+            'structure_length': self.structure_length / u,
+            'structure_thickness': self.structure_thickness / u,
+            'min_elemsize': self.min_elemsize / u,
+            'max_elemsize': self.max_elemsize / u})
+
+        # consistency check
+        # zeros = topo.boundary['inlet,wall,outlet,cylinder,structure'].integrate(function.normal(geom) * function.J(geom, 1), degree=2)
+        # numpy.testing.assert_allclose(zeros, 0, atol=1e-14, err_msg='boundaries do not form a watertight hull')
+
+        bezier = topo.sample('bezier', 2)
+        bezier_structure = topo['fluid'].boundary['structure'].sample(
+            'bezier', 3)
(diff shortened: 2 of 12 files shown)


gertjanvanzwieten:
I ran both scripts through autopep8 -i but the style check is still failing. Please advise :)

MakisH:
> I ran both scripts through autopep8 -i but the style check is still failing. Please advise :)

Thanks for the PR, @gertjanvanzwieten!
Maybe this is a version issue. We are using a [pre-commit hook](https://github.com/precice/tutorials/blob/develop/.pre-commit-config.yaml), which you could install to get the same version. This should not mess with the rest of your system.

gertjanvanzwieten:
For some reason actions fail to kick in but I think the scripts are compliant now, changes are generated with pre-commit run.

turek-hron-fsi3/fluid-nutils/fluid.py
```diff
@@ -0,0 +1,321 @@
...
+    with log.iter.plain('timestep', dynamic.times) as times:
+        t = next(times)
+
+        while participant.is_coupling_ongoing():
+
+            if participant.requires_reading_checkpoint():
+                args = checkpoint
+
+            cons['d'][r_where] = participant.read_data(
+                r_name, 'Displacement', r_ids, timestep)
```
uekerman, line 280: Just to confirm: You are using Newmark-beta for time integration. This means, you want to read at the end of the time step size, right? This is what is done right now.

turek-hron-fsi3/solid-nutils/solid.py
```diff
@@ -0,0 +1,269 @@
...
+        return dict(args, d=d + uδt + .5 * aδt2, d0=d, u0δt=uδt, a0δt2=aδt2)
+
+    def newmark_defo(self, d):
+        D = self.newmark_defo_args(
+            d, *[function.replace_arguments(d, [('d', t)]) for t in ('d0', 'u0δt', 'a0δt2')])
+        return D['u0δt'] / self.timestep, D['a0δt2'] / self.timestep**2
+
+
+@dataclass
+class DummyParticipant:
```
uekerman: Did you use this mock for testing? Is it still required?

turek-hron-fsi3/solid-nutils/solid.py
```diff
@@ -0,0 +1,269 @@
...
+        t = next(times)
+
+        while participant.is_coupling_ongoing():
+
+            if participant.requires_reading_checkpoint():
+                args = checkpoint
+
+            args = dynamic.newmark_defo_args(**args)
+            args['traction'] = participant.read_data(
+                rw_name, 'Stress', rw_ids, timestep)
```
uekerman, line 204: Again reading at the end of the time step. Is this what you want here?

uekerman (commented):
Thanks @gertjanvanzwieten for this nice and clean contribution :heart: and sorry for the very late reply.

Could you please tick the box in the PR to allow edits from maintainers?

> The scripts are only tested in mutual coupling, so I don't know how they hold up with other participants

For how long did you run the simulation. Only the first few timesteps or till the end? If the latter, did you get identical results compared to the monolithic one?

I was able to start Nutils-Nutils and the first few time steps converged. Now I am running OpenFOAM-Nutils for longer to check physical results.

TODO (for us):
- [x] add participants to README

uekerman:
`fluid-openfoam` coupled to `solid-nutils` gives correct results and runs fast enough.
Screenshot from 2025-07-30 09-21-39

uekerman:
@gertjanvanzwieten For the `fluid-nutils` solver, you start from zero initial values, but directly apply the full inflow velocity, correct? In OpenFOAM, we use a cos ramp over the first two seconds, `1 - cos((pi / 2) * t)) / 2` for the inflow velocity. Normally, sth like this is needed to get the partitioned simulation to run. Currently, `fluid-nutils` coupled to `solid-dealii` crashes.
Could you please add such a ramp?

turek-hron-fsi3/solid-nutils/solid.py
gertjanvanzwieten: That is not required anymore, apologies for the sloppiness (I thought I had removed it). I used it during development to validate the individual participants against the CFD and CSM tests of the original turek paper.

turek-hron-fsi3/fluid-nutils/fluid.py
gertjanvanzwieten: Indeed, everything is solved for the end of the time step. That's what I have come to really like about Newmark. Is it a poor fit for precice?

turek-hron-fsi3/solid-nutils/solid.py
gertjanvanzwieten: See my answer above.

gertjanvanzwieten:
> @gertjanvanzwieten For the `fluid-nutils` solver, you start from zero initial values, but directly apply the full inflow velocity, correct? In OpenFOAM, we use a cos ramp over the first two seconds, `1 - cos((pi / 2) * t)) / 2` for the inflow velocity. Normally, sth like this is needed to get the partitioned simulation to run. Currently, `fluid-nutils` coupled to `solid-dealii` crashes. Could you please add such a ramp?

The nutils solver applies the same ramp, see Dynamic.ramp_up on line 103 and its application on line 274. Does the solver fail in the very first time step? Could you try setting Dynamic.gamma to 1.0 and Dynamic.beta to 0.5 to make the fluid solver fully implicit?

uekerman:
True, there is a ramp in the Nutils fluid solver. Currently, deal.II crashes in the second iteration, so in the first iteration it receives non-zero values. I still have to take a closer look.

Changing to fully implicit time stepping does not change the behavior.

~Minor: It would be good to use the exact same ramp in Nutils than in OpenFOAM, so changing Nutils also to ramp up over 2 seconds. But this should not be the problem I am currently seeing.~
(It is exactly the same ramp)

turek-hron-fsi3/fluid-nutils/fluid.py
uekerman: No, perfect like this. `timestep` above means that you read values at the end of the time step. You could also use `timestep / 2` for a midpoint rule or similar.

turek-hron-fsi3/solid-nutils/solid.py
uekerman: Removed it. Maybe interesting to know: we are starting to develop such a general mock for developing and testing.

uekerman:
Interestingly, I get the same problem (deal.II is reporting a not-finite number) when running OpenFOAM-deal.II. And also currently for the perpendicular flap, which should be tested as part of the system tests, so I suspect a problem on my end. Btw, I am using the latest develop of the deal.II adapter and deal.II 9.2.0.

@MakisH Could you please test whether fluid-nutils coupled to solid-dealii works for you? At least the first few time steps.

I am running both Nutils solvers coupled to validate that also the fluid solver does the correct thing.

turek-hron-fsi3/solid-nutils/solid.py
gertjanvanzwieten: I thought of suggesting something like that, very useful indeed!

gertjanvanzwieten:
> For how long did you run the simulation. Only the first few timesteps or till the end? If the latter, did you get identical results compared to the monolithic one?

Just noticed I didn't reply to this yet. It's been a while since submitting this, but I'm pretty certain it ran out the entire 15 seconds time frame.

gertjanvanzwieten:
And I remember something else: I did not give either script its own timestep parameter, instead I just use the one I receive from precice. Is that acceptable for a tutorial?

uekerman:
> And I remember something else: I did not give either script its own timestep parameter, instead I just use the one I receive from precice. Is that acceptable for a tutorial?

Yes, this is valid usage. Especially for such a benchmark, where classically both solvers used the same fixed time step size.

turek-hron-fsi3/README.md
```diff
@@ -28,14 +28,16 @@ preCICE configuration (image generated using the [precice-config-visualizer](htt
 Fluid participant:
 
 * OpenFOAM (pimpleFoam). In case you are using a very old OpenFOAM version, you will need to adjust the solver to `pimpleDyMFoam` in the `Fluid/system/controlDict` file. For more information, have a look at the [OpenFOAM adapter documentation](https://precice.org/adapter-openfoam-overview.html).
+* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v9.0. This case takes significantly longer to run than the OpenFOAM case, see [related issue](https://github.com/precice/tutorials/issues/506).
```
MakisH, line 31: Self-reminder: We will also need to add this in the general overview (website repository) https://precice.org/tutorials.html

turek-hron-fsi3/README.md
```diff
@@ -28,14 +28,16 @@ preCICE configuration (image generated using the [precice-config-visualizer](htt
...
 
 Solid participant:
 
 * deal.II. For more information, have a look at the [deal.II adapter documentation](https://precice.org/adapter-dealii-overview.html). This tutorial requires the nonlinear solid solver. Please copy the nonlinear solver executable to the `solid-dealii` folder or make it discoverable at runtime and update the `solid-dealii/run.sh` script.
+* Nutils. For more information, have a look at the [Nutils adapter documentation](https://precice.org/adapter-nutils.html). This Nutils solver requires at least Nutils v9.0.
```
MakisH, line 36: Do we need to document the version here, if it is already documented in the `requirements.txt`? We now also always install in a virtual environment.

I would rather remove these two instances, as we will most definitely forget to update them later.

turek-hron-fsi3/README.md
```diff
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
 
 ## Post-processing
 
-You can visualize the results of the coupled simulation using e.g. ParaView. Fluid results are in the OpenFOAM format and you may load the `fluid-openfoam.foam` file. Solid results are in VTK format.
+You can visualize the results of the coupled simulation using e.g. ParaView. OpenFOAM uses the specific format `fluid-openfoam.foam`, which you directly load in Paraview or convert to VTK with `foamToVTK`. deal.II writes VTK files. Both Nutils solvers currently don not write VTK files (but their own in-situ visualization), but this can be added easily similarly to the [perpendicular flap solvers](https://github.com/precice/tutorials/blob/98a78fe2dc2f6c5d84b2b30d35d00352782236f8/perpendicular-flap/fluid-nutils/fluid.py#L227).
```
MakisH:
```suggestion
You can visualize the results of the coupled simulation using e.g. ParaView. OpenFOAM uses an OpenFOAM-specific format, and you can directly load the (empty) file `fluid-openfoam.foam` in Paraview or convert the results to VTK with `foamToVTK`. deal.II writes VTK files. Both Nutils solvers currently do not write VTK files (but use their own in-situ visualization), but this can be easily added similarly to the [perpendicular flap solvers](https://github.com/precice/tutorials/blob/98a78fe2dc2f6c5d84b2b30d35d00352782236f8/perpendicular-flap/fluid-nutils/fluid.py#L227).
```

turek-hron-fsi3/README.md
```diff
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
...
 
-If you want to visualize both domains with ParaView, keep in mind that the deal.II solver writes results every few timesteps, while the OpenFOAM solver writes in reference to simulated time. For this reason, make sure that you use compatible write intervals. You may also need to convert the OpenFOAM results to VTK (with the command `foamToVTK`).
+If you want to visualize both domains with ParaView, keep in mind that the different solvers may write results with different output frequencies.
```
MakisH:
```suggestion
If you want to visualize both domains with ParaView, keep in mind that the different solvers may write results with different output frequencies, which you might want to [synchronize](https://precice.org/tutorials-visualization.html#synchronizing-results).
```

turek-hron-fsi3/README.md
```diff
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
...
 
 There is an [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. At the end of `run.sh`, we call `openfoam_remove_empty_dirs` (provided by `tools/openfoam-remove-empty-dirs`) to delete the additional files before importing the results in ParaView.
```
MakisH:
```suggestion
There is a [known issue](https://github.com/precice/openfoam-adapter/issues/26) that leads to additional "empty" result directories when running with some OpenFOAM versions, leading to inconveniences during post-processing. At the end of `run.sh`, we call `openfoam_remove_empty_dirs` (provided by `tools/openfoam-remove-empty-dirs`) to delete the additional files before importing the results in ParaView.
```

turek-hron-fsi3/README.md
```diff
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
...
 
-Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`.  The resulting graph shows the vertical (y) displacement of the tip of the flap.
+Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`. You need to specify the directory of the selected solid participant as a command line argument, so that the script can pick-up the desired watchpoint file, e.g. `plot-displacement.sh solid-dealii`. The resulting graph shows the vertical (y) displacement of the tip of the flap.
```
MakisH: (rendering on GitHub is broken, but maybe you can manually commit what is in the unrendered version)

```suggestion
Moreover, as we defined a watchpoint at the flap tip (see `precice-config.xml`), we can plot it with gnuplot using the script `plot-displacement.sh`, which expects the directory of the selected solid participant as a command line argument. For example:

 ```shell
 plot-displacement.sh solid-dealii
 ```
 
The resulting graph shows the vertical (y) displacement of the tip of the flap.
```

turek-hron-fsi3/README.md
```diff
@@ -57,19 +59,19 @@ In the first few timesteps, many coupling iterations are required for convergenc
...
 
 ![FSI3 watchpoint](images/tutorials-turek-hron-fsi3-tip-plot.png)
 
 Before running the simulation again, you may want to cleanup any result files using the script `clean-tutorial.sh`.
 
-## Mesh refinement
+## Mesh refinement for OpenFOAM
 
 In `fluid-openfoam/system/`, we provide three different fluid meshes:
```
MakisH: Since we don't provide another case, I would leave it as-it.

```suggestion
## Mesh refinement

In `fluid-openfoam/system/`, we provide three different fluid meshes:
```

turek-hron-fsi3/fluid-nutils/fluid.py
```diff
@@ -0,0 +1,321 @@
...
+    x_bz = bezier.bind(ns.x)
+    u_bz = bezier.bind(ns.u)
+    p_bz = bezier.bind(ns.p) - \
+        topo.points['B'].sample('gauss', 1).bind(ns.p)[0]
+
+    bbezier = topo['fluid'].boundary['cylinder,structure'].sample('bezier', 3)
+    x_bbz = bbezier.bind(ns.x)
+
+    assert participant.requires_writing_checkpoint()
+    checkpoint = args
```
MakisH: Is this meant to be the final version?
Strange to add this into an assertion. A comment would be helpful here.

Same for the reading.

I also understand that `args` is the complete state of the simulation, correct?

turek-hron-fsi3/fluid-nutils/fluid.py
```diff
@@ -0,0 +1,321 @@
...
+                log.info(f'lift: {L:N/m}')
+                log.info(f'drag: {D:N/m}')
+                with export.mplfigure('force.jpg', dpi=150) as fig:
+                    dynamic.add_and_plot(
+                        'lift [N/m]', t / 's', L / 'N/m', ax=fig.add_subplot(211))
+                    dynamic.add_and_plot(
+                        'drag [N/m]', t / 's', D / 'N/m', ax=fig.add_subplot(212, xlabel='time [s]'))
+
+                t = next(times)
+
```
MakisH, line 331: `precice.finalize()` is missing.
Even if it is implicitly called by the destructor, it would be nice to have it explicit, due to the educational purpose of the tutorials.

turek-hron-fsi3/fluid-nutils/fluid.py
```diff
@@ -0,0 +1,321 @@
...
+            args = sys_d.solve(arguments=args, constrain=cons,
+                               tol=1e-10)  # mesh extension
+            # Navier-Stokes time step
+            args = sys_up.solve(arguments=args, constrain=cons, tol=1e-10)
+            args = sys_t.solve(arguments=args, constrain=cons,
+                               tol=1e-10)  # project traction
+
+            participant.write_data(
+                w_name, 'Stress', w_ids, w_sample.eval(ns.t, args) / precice_stress)
+            participant.advance(timestep)
```
MakisH, line 295: Adjusting the time step size is missing:

https://github.com/precice/tutorials/blob/develop/perpendicular-flap/fluid-nutils/fluid.py#L165-L166

You already have it after initialization, but it is needed in every time step.

MakisH (commented):
Thanks again for the contribution! Some mostly minor comments to consider from my side. I have not really looked in detail into the Python scripts, but I think I found a couple of issues.

MakisH:
> Interestingly, I get the same problem (deal.II is reporting a not-finite number) when running OpenFOAM-deal.II. And also currently for the perpendicular flap, which should be tested as part of the system tests, so I suspect a problem on my end. Btw, I am using the latest develop of the deal.II adapter and deal.II 9.2.0.

The OpenFOAM-deal.II perpendicular flap indeed works as expected.

> @MakisH Could you please test whether fluid-nutils coupled to solid-dealii works for you? At least the first few time steps.

It does not, but for me Nutils is the one that crashes:

```
---[precice]  iteration: 1 of 100 (min 1), time-window: 1, time: 0 of 15, time-window-size: 0.001, max-time-step-size: 0.001, ongoing: yes, time-window-complete: no, write-iteration-checkpoint 
writing log to: file:///home/gc/public_html/fluid/log.html
NUTILS 9.0 "Jook-Sing"
arguments > domain=channel_length=2.5m,channel_height=.41m,x_center=.2m,y_center=.2m,cylinder_radius=5cm,structure_length=35cm,structure_thickness=2cm,min_elemsize=4mm,max_elemsize=4cm
arguments > fluid=density=1kg/L,viscosity=1Pa*s,velocity=2m/s
arguments > dynamic=init=2s,window=1s,gamma=0.8,beta=0.4
start 2025-08-01 14:58:39
Re: 199.99999999999997
gmsh > Running '/home/gc/repos/precice/tutorials/turek-hron-fsi3/fluid-nutils/.venv/bin/gmsh /home/gc/repos/precice/tutorials/turek-hron-fsi3/fluid-nutils/fluid.geo -o /tmp/tmpfhie3670.msh -order 2 -2 -bin -setnumber channel_length 2.5 -setnumber channel_height 0.41 -setnumber x_center 0.2 -setnumber y_center 0.2 -setnumber cylinder_radius 0.05 -setnumber structure_length 0.35000000000000003 -setnumber structure_thickness 0.02 -setnumber min_elemsize 0.004 -setnumber max_elemsize 0.04' [Gmsh 4.14.0, 1 node, max. 1 thread]
gmsh > Started on Fri Aug  1 14:58:39 2025
gmsh > Reading '/home/gc/repos/precice/tutorials/turek-hron-fsi3/fluid-nutils/fluid.geo'...
gmsh > [  0%] Difference                                                                                  
gmsh > [ 10%] Difference                                                                                  
gmsh > [ 20%] Difference                                                                                  
gmsh > [ 30%] Difference                                                                                  
gmsh > [ 40%] Difference                                                                                  
gmsh > [ 50%] Difference                                                                                  
gmsh > [ 70%] Difference - Performing intersection of shapes                                                                                
gmsh > [ 80%] Difference - Making faces                                                                                
gmsh > [ 90%] Difference                                                                                  
gmsh >                                                                                 
gmsh > [  0%] Difference                                                                                  
gmsh > [ 10%] Difference                                                                                  
gmsh > [ 20%] Difference                                                                                  
gmsh > [ 30%] Difference                                                                                  
gmsh > [ 40%] Difference                                                                                  
gmsh > [ 50%] Difference                                                                                  
gmsh > [ 60%] Difference                                                                                  
gmsh > [ 70%] Difference - Filling splits of edges                                                                                
gmsh > [ 80%] Difference - Adding holes                                                                                
gmsh > [ 90%] Difference - Making faces                                                                                
gmsh > [  0%] Fragments                                                                                  
gmsh > [ 10%] Fragments                                                                                  
gmsh > [ 20%] Fragments                                                                                  
gmsh > [ 30%] Fragments                                                                                  
gmsh > [ 40%] Fragments                                                                                  
gmsh > [ 50%] Fragments                                                                                  
gmsh > [ 60%] Fragments                                                                                  
gmsh > [ 70%] Fragments                                                                                  
gmsh > [ 80%] Fragments - Splitting faces                                                                                
gmsh >                                                                                 
gmsh > Done reading '/home/gc/repos/precice/tutorials/turek-hron-fsi3/fluid-nutils/fluid.geo'
gmsh > Meshing 1D...
gmsh > [  0%] Meshing curve 1 (Line)
gmsh > [ 10%] Meshing curve 2 (Line)
gmsh > [ 20%] Meshing curve 3 (Line)
gmsh > [ 20%] Meshing curve 4 (Line)
gmsh > [ 30%] Meshing curve 5 (Line)
gmsh > [ 40%] Meshing curve 6 (Line)
gmsh > [ 40%] Meshing curve 7 (Line)
gmsh > [ 50%] Meshing curve 8 (Line)
gmsh > [ 60%] Meshing curve 9 (Ellipse)
gmsh > [ 60%] Meshing curve 10 (Ellipse)
gmsh > [ 70%] Meshing curve 11 (Ellipse)
gmsh > [ 70%] Meshing curve 12 (Ellipse)
gmsh > [ 80%] Meshing curve 13 (Line)
gmsh > [ 90%] Meshing curve 14 (Line)
gmsh > [ 90%] Meshing curve 15 (Line)
gmsh > [100%] Meshing curve 16 (Line)
gmsh > Done meshing 1D (Wall 0.142933s, CPU 0.139479s)
gmsh > Meshing 2D...
gmsh > [  0%] Meshing surface 1 (Plane, Frontal-Delaunay)
gmsh > [ 30%] Meshing surface 2 (Plane, Frontal-Delaunay)
gmsh > [ 50%] Meshing surface 3 (Plane, Frontal-Delaunay)
gmsh > [ 70%] Meshing surface 4 (Plane, Frontal-Delaunay)
gmsh > [ 90%] Meshing surface 5 (Plane, Frontal-Delaunay)
gmsh > Done meshing 2D (Wall 0.218753s, CPU 0.218473s)
gmsh > Meshing order 2 (curvilinear on)...
gmsh > [  0%] Meshing curve 1 order 2
gmsh > [ 10%] Meshing curve 2 order 2
gmsh > [ 10%] Meshing curve 3 order 2
gmsh > [ 20%] Meshing curve 4 order 2
gmsh > [ 20%] Meshing curve 5 order 2
gmsh > [ 30%] Meshing curve 6 order 2
gmsh > [ 30%] Meshing curve 7 order 2
gmsh > [ 40%] Meshing curve 8 order 2
gmsh > [ 40%] Meshing curve 9 order 2
gmsh > [ 50%] Meshing curve 10 order 2
gmsh > [ 50%] Meshing curve 11 order 2
gmsh > [ 60%] Meshing curve 12 order 2
gmsh > [ 60%] Meshing curve 13 order 2
gmsh > [ 70%] Meshing curve 14 order 2
gmsh > [ 70%] Meshing curve 15 order 2
gmsh > [ 80%] Meshing curve 16 order 2
gmsh > [ 80%] Meshing surface 1 order 2
gmsh > [ 90%] Meshing surface 2 order 2
gmsh > [ 90%] Meshing surface 3 order 2
gmsh > [100%] Meshing surface 4 order 2
gmsh > [100%] Meshing surface 5 order 2
gmsh > Surface mesh: worst distortion = 0.965309 (0 elements in ]0, 0.2]); worst gamma = 0.63475
gmsh > Done meshing order 2 (Wall 0.0498959s, CPU 0.047879s)
gmsh > 28851 nodes 14912 elements
gmsh > Writing '/tmp/tmpfhie3670.msh'...
gmsh > Done writing '/tmp/tmpfhie3670.msh'
gmsh > Stopped on Fri Aug  1 14:58:39 2025 (From start: Wall 0.463014s, CPU 0.711052s)

gmsh > loaded 2d gmsh topology consisting of #4869 elements
- volume groups: fluid #4869
- boundary groups: inlet #12, outlet #11, wall #129, cylinder #71, structure #154
- point groups: A #1, B #1
gmsh > 7492 vertices are not used
mesh.jpg
solve_constraints > optimizing for argument d (2623,2) with drop tolerance 1e-10
solve_constraints > residual norm: 2.0e-15
solve_constraints > optimal value: 1.4e-14
solve_constraints > constrained 310 degrees of freedom of d
solve_constraints > optimizing for argument u (10115,2) and d (2623,2) with drop tolerance 1e-10
solve_constraints > residual norm: 5.2e-17
solve_constraints > optimal value: 2.2e-16
solve_constraints > constrained 1487 degrees of freedom of u
solve_constraints > constrained 448 degrees of freedom of d
solve_constraints > optimizing for argument t (10115,2) with drop tolerance 1e-10
solve_constraints > residual norm: 0.0e+00
solve_constraints > optimal value: 0.0e+00
solve_constraints > constrained 900 degrees of freedom of t
timestep 1 > solve > optimizing for argument d (2623,2) using newton method
timestep 1 > solve > iter 0 > residual norm: 1.7e-12
timestep 1 > solve > optimal value: 4.9e+03
timestep 1 > solve > solving for argument u (10115,2) and p (2623) using newton method
timestep 1 > solve > iter 0 > residual norm: 1.9e-04
nutils.matrix.MatrixError: reordering problem
```

Packages:

```
appdirs==1.4.4
contourpy==1.3.2
cycler==0.12.1
Cython==3.1.2
fonttools==4.59.0
gmsh==4.14.0
kiwisolver==1.4.8
markdown-it-py==3.0.0
matplotlib==3.10.5
mdurl==0.1.2
meshio==5.3.5
mpi4py==4.1.0
numpy==1.26.4
nutils==9.0
nutils-poly==1.0.1
packaging==25.0
pillow==11.3.0
Pygments==2.19.2
pyparsing==3.2.3
pyprecice==3.2.1
python-dateutil==2.9.0.post0
rich==14.1.0
six==1.17.0
stringly==1.0b3
treelog==1.0
typing_extensions==4.14.1
```

and Python 3.10.12 on Ubuntu 22.04. You might want to change the minimum required/recommended Python version of Nutils 9, because with 3.12, it goes further:

```
---[precice] ERROR:  The required actions write-iteration-checkpoint are not fulfilled. Did you forget to call "requiresReadingCheckpoint()" or "requiresWritingCheckpoint()"?
```

which might be due to the unusual way of calling `requires_writing_checkpoint()`.

turek-hron-fsi3/README.md
uekerman: I copied from other Nutils descriptions. Let's merge here and then remove everywhere?

turek-hron-fsi3/fluid-nutils/fluid.py
uekerman: Discussed here: https://github.com/precice/tutorials/pull/648#issuecomment-3148311046

uekerman:
Nutils-Nutils gives correct results:

Screenshot from 2025-08-05 14-03-52

turek-hron-fsi3/fluid-nutils/fluid.py
gertjanvanzwieten: But this not just about following precice's time step size, but about accounting for the fact that precice may (apparently?) change the time step size at every moment. I just started looking into how to modify the code for this.

gertjanvanzwieten:
> timestep 1 > solve > optimizing for argument d (2623,2) using newton method
> timestep 1 > solve > iter 0 > residual norm: 1.7e-12
> timestep 1 > solve > optimal value: 4.9e+03
> timestep 1 > solve > solving for argument u (10115,2) and p (2623) using newton method
> timestep 1 > solve > iter 0 > residual norm: 1.9e-04
> nutils.matrix.MatrixError: reordering problem

This is an MKL Pardiso error that I personally never encountered before, but that is most likely caused by the matrix being problematic, singular or otherwise. You could try switching matrix backends by setting the `NUTILS_MATRIX=scipy` environment -- I doubt it will make the problem go away but Scipy may have something more insightful to say in the error.

gertjanvanzwieten:
> which might be due to the unusual way of calling `requires_writing_checkpoint()`.

Possible, I'm in the process of changing that.

gertjanvanzwieten:
I pushed some modifications based on @MakisH feedback: fix order of operations, request time step every iterations, and end with finalize. I'm still running the coupled scripts to verify that results are still correct but so far so good. I'm curious if these changes do anything towards fixing the crashes you guys have been observing.

gertjanvanzwieten:
Question: I notice (and noticed before) that my results do not remain identical from one run to the next. I'm not aware of any source of randomness in my code, but if there is that would certainly need fixing. However I seem to recall that my one sided "dummy participant" tests were indeed reproducible. Could it be that Precice introduces randomness through any of its operations, for instance by depending on order of communication or something of the kind?

gertjanvanzwieten:
PS I don't know why actions are complaining about autopep8 again, pre-commit passes just fine locally.

MakisH:
> PS I don't know why actions are complaining about autopep8 again, pre-commit passes just fine locally.

This also fails for `develop` already due to a dependency update. I suggest ignoring this on this branch and we fix it on `develop`.

MakisH:
> You could try switching matrix backends by setting the `NUTILS_MATRIX=scipy` environment -- I doubt it will make the problem go away but Scipy may have something more insightful to say in the error.

Same, not much additional info: [log-4.html.txt](https://github.com/user-attachments/files/21623109/log-4.html.txt)

gertjanvanzwieten:
I just looked at the results of the updated scripts and I don't think I'm getting the correct frequencies, so I may have messed up the Newmark modifications. I'll run more tests and report back in a few days.

gertjanvanzwieten:
> I just looked at the results of the updated scripts and I don't think I'm getting the correct frequencies, so I may have messed up the Newmark modifications. I'll run more tests and report back in a few days.

Actually scratch that, I had forgotten that we're doing FSI3, not FSI2. The results look correct.

solid
fluid

MakisH:
> > You could try switching matrix backends by setting the `NUTILS_MATRIX=scipy` environment -- I doubt it will make the problem go away but Scipy may have something more insightful to say in the error.
> 
> Same, not much additional info: [log-4.html.txt](https://github.com/user-attachments/files/21623109/log-4.html.txt)

Tried again, after removing a user-wide installation of Nutils 9.

- With the default, I still get the same issue.
- With scipy, it now works (at least for three time windows).

MakisH (approved):
The structure and code now look fine to me.

If I use scipy, I can run the case for a few (three) time windows, but it still crashes in the beginning with the default backend. If it works for @uekerman, then that's fine.

I have not executed the complete simulation (too long) and I have not compared the listed results to the reference (I assume this is already ok).

As I wrote above, we can ignore the failing style check, as currently our autopep8 check is broken.

gertjanvanzwieten:
> Tried again, after removing a user-wide installation of Nutils 9.
> 
>     * With the default, I still get the same issue.
> 
>     * With scipy, it now works (at least for three time windows).

Pardiso has been known to sometimes stumble over matrices that Scipy (SuperLU) happily solves. I have never been able to figure out the reason, nor have I managed to solve it by tweaking any of Pardiso's myriad settings. Unfortunately Scipy is not much of a fallback as it is very slow, as I'm sure you have noticed.

Other than adding support for another performant matrix back-end (a long term goal on the Nutils roadmap) I am not sure what to do with this situation.

gertjanvanzwieten:
One thing that I forgot to bring up before: the development of these scripts has been funded by TU Eindhoven, and I would like to give them an attribution for it. What would be the best place for that?

MakisH:
> One thing that I forgot to bring up before: the development of these scripts has been funded by TU Eindhoven, and I would like to give them an attribution for it. What would be the best place for that?

Feel free to add a comment in the scripts.
Potentially also in the `README.md`, but we don't do this consistently at the moment. One example would be https://precice.org/tutorials-flow-around-controlled-moving-cylinder.html#acknowledgements

MakisH:
Regarding the formatting issues, try merging develop back to your branch.
Check that after reformatting, your code still runs, as we had some issues with other scripts already with the latest autopep8 version.

gertjanvanzwieten:
I rebased to develop, squashed the "Update turek-hron-fsi3/README.md" commits into e3d28a3, and added an acknowledgement in the readme f6f8d14.

MakisH:
Looks good, but why not add a comment in these or the original scripts as well? Would be easier to keep track of the attribution over the years.

gertjanvanzwieten:
Makes sense, done.

MakisH (approved):
To summarize once again for @uekerman: all good from my side, just waiting for a confirmation that you can run the scripts with the default setting (because I can only run them with scipy).

uekerman:
I can run both Nutils participants coupled without modifying the envs or the run scripts and without setting any Nutils-related environment variables.

uekerman (approved):

turek-hron-fsi3/README.md
MakisH: Added in https://github.com/precice/precice.github.io/pull/583

PR #654, opened 2025-07-04
Make the Quickstart compatible with foam-extend

hoehnp:
I added a modified version of the quickstart tutorial for foam-extend with the purpose of testing the foam-extend adapter.


quickstart/fluid-foam-extend/system/controlDict
```diff
@@ -0,0 +1,82 @@
...
+\*---------------------------------------------------------------------------*/
+FoamFile
+{
+    version     2.0;
+    format      ascii;
+    class       dictionary;
+    location    "system";
+    object      controlDict;
+}
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
```
MakisH: We don't include these headers in our tutorial files, as they are often outdated (also, this file is not part of foam-extend):

```suggestion
FoamFile
{
    version     2.0;
    format      ascii;
    class       dictionary;
    location    "system";
    object      controlDict;
}
```

quickstart/fluid-foam-extend/system/controlDict
MakisH, line 1: For easier maintenance, it would be good if this file followed the same order and formatting as the one in `fluid-openfoam`, so that we can easily take a diff.

quickstart/fluid-foam-extend/system/controlDict
```diff
@@ -0,0 +1,82 @@
...
+
+writePrecision  10;
+
+writeCompression uncompressed;
+
+timeFormat      general;
+
+timePrecision   8;
+
+runTimeModifiable true;
```
MakisH: This should not be needed.

quickstart/fluid-foam-extend/system/controlDict
```diff
@@ -0,0 +1,82 @@
...
+
+adjustTimeStep  no;
+
+maxCo           0.9;
```
MakisH: Is this explicitly needed?

MakisH (commented):
Thanks for the contribution! A first quick review of the changes, without having tried to run the code.

We generally have alternatives for the perpendicular-flap tutorial (which is meant to be something to copy from), not for the Quickstart (which is meant to be a specific first case to run). With that said, I do understand that this is a minimal case for you.

Since the changes are very minimal, I wonder if we could make the original case compatible with foam-extend, even if we just applied a small patch automatically. Some of the changes for foam-extend should still work with OpenFOAM (e.g., the `noSlip`/`zeroGradient` or the path of the preCICE config) or not be that relevant (e.g., the forces function object).

___

List of differences compared to `fluid-openfoam`:

`0/U`: 

```diff
- type            noSlip;
+ type            zeroGradient;
```

`dynamicMeshDict`:

```diff
- motionSolverLibs ("libfvMotionSolvers.so");
+ motionSolverLibs ("libfvMotionSolver.so");
```
```diff
+ diffusivity uniform;
```

`controlDict`:

```diff
  libs
  (
+    "libforces.so"
+    "liblduSolvers.so"
     "libpreciceAdapterFunctionObject.so"
  );
```

+ `forces` function object configuration (do we actually need that?).

`fvSchemes`:

```diff
-     div((nuEff*dev2(T(grad(U))))) Gauss linear;
+     div((nuEff*dev(T(grad(U))))) Gauss linear;
```

`fvSolution`:

```diff
     p
     {
         solver           GAMG;
         tolerance        1e-6;
         relTol           1e-4;
         smoother         DICGaussSeidel;
+        agglomerator    faceAreaPair;
+        nCellsInCoarsestLevel 10;
+        mergeLevels 1;
     }
```

`preciceDict`:

```diff
- preciceConfig "../precice-config.xml";
+ preciceConfig ".."/"precice-config.xml";
```

quickstart/fluid-foam-extend/system/controlDict
hoehnp: yes, without it, it was not running.

hoehnp: I start to wonder if this only applies to the debug version for some reason.

MakisH: But is this even a valid combination? If `adjustTimeStep` is disabled, what effect should `maxCo` have? We could still set one with a comment that it is not used and only there for compatibility reasons.

quickstart/fluid-openfoam/system/controlDict
```diff
@@ -36,14 +36,23 @@ timeFormat          general;
 
 timePrecision       8;
 
-libs ("libpreciceAdapterFunctionObject.so");
+libs
```
MakisH, line 43: Input by @hoehnp: Formatting is important here, so that the run script can add more lines.

Let's add a comment.

quickstart/fluid-foam-extend/system/controlDict
MakisH: Not needed anymore.

quickstart/fluid-openfoam/0/U
```diff
@@ -22,11 +22,13 @@ boundaryField
     top
     {
         type            noSlip;
+        value           uniform (0 0 0);
```
MakisH, line 25: We need a comment that this is important for foam-extend.

quickstart/fluid-openfoam/constant/dynamicMeshDict
```diff
@@ -14,6 +14,8 @@ motionSolverLibs ("libfvMotionSolvers.so");
 solver      displacementLaplacian;
 // OpenFOAM9 or newer: rename "solver" to "motionSolver"
 
+diffusivity uniform;
```
MakisH: Add a comment that this is (only) needed for foam-extend.

quickstart/fluid-openfoam/system/controlDict
```diff
@@ -36,14 +36,23 @@ timeFormat          general;
...
+(
+    "libpreciceAdapterFunctionObject.so"
+);
 functions
 {
     forces
     {
         type                forces;
         libs                ( "libforces.so" );
+        functionObjectLibs  ( "libforces.so" );
```
MakisH, line 58: Let's move the foam-extend related things to one group, and add a comment.

tools/run-foam-extend.sh
MakisH, line 1: It might be a good idea to take a backup of the original folders/files before starting to apply changes on them, so that if someone accidentally changes their own (untracked) case, they can easily revert.

Let's start with the script, and if this gets tricky to maintain in the long run, we can also switch back to a separate folder.

**Edit 1:** It's been a while since the two reviews, and I only now remember the motivation for the script: Since this is the Quickstart, we do not want any other version of the two solvers.

**Edit 2:** I also see now that this is in the `tools/` directory, meant to be a general solution for all tutorials. This is great!

MakisH (commented):
Looking good, just needs some comments.

MakisH (approved):
Thank you once more for the contribution. I have added some documentation, and I think that this is now ready to merge.

Note that this tutorial is not meant to provide alternatives (and this is why we should not have an alternative folder - I missed the context when we talked in person). A better tutorial to have a foam-extend alternative for would be the perpendicular-flap. However, since foam-extend can be seen as an alternative version of the same solver, and since this is a test case for you, I see the value and I hold back my general objection to modifying this tutorial.

I also notice now that the "porting" script is actually in the `tools/`, so it is rather reusable across tutorials. This is a great step in the right direction. As a user, I would expect an overview of needed configuration changes in the adapter documentation (currently the best place would be https://precice.org/adapter-openfoam-support.html, which is sourced from the openfoam-adapter repository).

MakisH:
I think this was only open because it was missing some hint for it in the OpenFOAM adapter documentation. I will add that separately, no need for this to stay open any longer.

PR #658, opened 2025-07-25
Add julia example in resonant circuit

marinlauber:
This PR adds Julia participants to the `resonant-circuit`.


resonant-circuit/capacitor-julia/run.sh
MakisH, line 1: In our run scripts, we have some more features that we would need here for consistency as well:

- shell is bash, set via env
- we `set -e -u` to make the script more robust
- we capture all logging into a file

See the [python variant](https://github.com/precice/tutorials/blob/develop/resonant-circuit/capacitor-python/run.sh) for an example.

It would be good if the script also installed the Julia bindings, in case they are not already installed. Similarly to the Python example.

resonant-circuit/README.md
```diff
@@ -33,6 +33,7 @@ preCICE configuration (image generated using the [precice-config-visualizer](htt
 * *MATLAB* A solver using the [MATLAB bindings](https://precice.org/installation-bindings-matlab.html).
  Before running this tutorial, follow the [instructions](https://precice.org/installation-bindings-matlab.html) to correctly install the MATLAB bindings.
 * *Python* A solver using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html).
+* *Julia* A solver using the preCICE [Julia bindings](https://precice.org/installation-bindings-julia.html).
```
MakisH, line 35: Self-reminder: We will need also an edit in https://precice.org/tutorials.html

resonant-circuit/capacitor-julia/capacitor.jl
```diff
@@ -0,0 +1,90 @@
...
+    phi = 0                       # Phase of the signal
+
+    w0 = 1 / sqrt(L * C)          # Resonant frequency
+    U0 = -w0 * L * Io * sin(phi)  # Initial condition for U
+
+    function f_U(u, p, t)
+        (dt, C, mesh_name, read_data_name, vertex_ids) = p
+        I = PreCICE.readData(mesh_name, read_data_name, vertex_ids, dt)
+        return -I[1] / C      # Time derivative of U
+    end
```
MakisH, line 34: I find it strange to have the `readData()` call inside such a function and not at the same level as `writeData()`.

I see that you are getting the `dt` from `p` (`parameters`? how is that mapped to `p`? I guess this is Julia-specific), but this is also known outside.

What complications am I missing?

resonant-circuit/README.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Resonant Circuit
 permalink: tutorials-resonant-circuit.html
-keywords: MATLAB, Python
+keywords: MATLAB, Python, Julia
```
MakisH, line 4: The rest of the page seems to be a bit focused on MATLAB. This is a good chance to change this (I can do).

MakisH (commented):
Thank you for the contribution, @marinlauber!

I still have not managed to run it (first time I try out Julia and our Julia bindings), but I added some first comments that you could already address if you want.

resonant-circuit/capacitor-julia/run.sh
marinlauber: I've addressed most of those, for the Julia binding installation I just added a line in `run.sh` that initialises the environment using the packages specified in the `Project.toml` file (this file is similar to setting up a Python environment to run the project).
Not sure if that's why you had in mind.

resonant-circuit/capacitor-julia/capacitor.jl
marinlauber: In my understanding, there `readData` is called at every iteration of the ODE solver because waveform iteration is used, this interpolated (cubic) the vaiable `I` inside the time step instead of having a constant value.

I copied what was done in the Python version.

marinlauber: `dt` is packed into a `params` tuple that is passed to the`ODEProblem` [here](https://github.com/precice/tutorials/pull/658/files#diff-89005e6f2e1975bbd7bc22fe928826b3ed1c222f2bfc3c8dfc3569340806f650R57-R63)
This is the passed internally to the `func(u, param, t)` to solve, see [here](https://docs.sciml.ai/DiffEqDocs/stable/getting_started/)

marinlauber:
@MakisH, I've modified the example to be more in line with the other examples, the `run.sh` script now also generates the `Project.toml` file and installs the `PreCICE.jl` binding (with the ODE solver lib).

resonant-circuit/capacitor-julia/run.sh
MakisH: Thanks for the quick response. Yes, this is exactly what I had in mind.

I now still need to test this (currently https://github.com/precice/PreCICE.jl/issues/70 is blocking me).

resonant-circuit/capacitor-julia/capacitor.jl
MakisH: > I copied what was done in the Python version.

Ah, I see. Then all good!

marinlauber:
@MakisH is there anything from my side that needs to be done?

MakisH:
> @MakisH is there anything from my side that needs to be done?

Not yet, at least not from my side. I would like to manage to run this first, currently blocked by https://github.com/precice/PreCICE.jl/issues/70.

I will come back to you, but I cannot promise that this will be before the workshop.

MakisH (approved):
Thanks again for your patience, @marinlauber! I could now run the case, and I directly applied a few additional modifications that were needed (some to the other participants of the same tutorial, since we are touching this).

PR #659, opened 2025-08-02
Fix domain definition in documentation

BenjaminRodenberg:


BenjaminRodenberg:
Forget about this PR. I was thinking about the coordinates from the perspective of the corner points of the squares. But actually it is a Cartesian product. I guess I've spent too much time with TikZ lately.

PR #663, opened 2025-08-21
Add implementation to enable parallel execution of macro-dumux

mathiskelm:
Handle parallel execution of the participant macro-dumux in the two-scale heat conduction tutorial. Currently the communication of coupling data between ranks is handled in the application. In the future, it would be better to handle this in the DuMuX-preCICE adapter.

Checklist:

### DIFF
--- a/two-scale-heat-conduction/macro-dumux/appl/main.cc
+++ b/two-scale-heat-conduction/macro-dumux/appl/main.cc
@@ -278,6 +278,8 @@ int main(int argc, char **argv)
                                                       dt);
       couplingParticipant.readQuantityFromOtherSolver(meshName,
                                                       readDataPorosity, dt);
+      // store coupling data in problem
+      problem->spatialParams().updateCouplingData();
     }
     std::cout << "Solver starts" << std::endl;
 
--- a/two-scale-heat-conduction/macro-dumux/appl/main.cc
+++ b/two-scale-heat-conduction/macro-dumux/appl/main.cc
@@ -24,6 +24,7 @@
 #include <dune/common/timer.hh>
 #include <dune/grid/io/file/vtk.hh>
 #include <dune/istl/io.hh>
+#include <dune/grid/common/gridenums.hh>
 
 #include "properties.hh"
 
@@ -114,6 +115,8 @@ int main(int argc, char **argv)
   // get mesh coordinates
   std::vector<double> coords;
   std::vector<int>    coupledElementIdxs;
+  std::vector<double> overlapCoords;
+  std::vector<int>    overlapElementIdxs;
 
   // coordinate loop (created vectors are 1D)
   // these positions of cell centers are later communicated to precice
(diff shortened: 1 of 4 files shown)


two-scale-heat-conduction/macro-dumux/appl/main.cc
```diff
@@ -119,6 +120,8 @@ int main(int argc, char **argv)
   // these positions of cell centers are later communicated to precice
   std::cout << "Coordinates: " << std::endl;
   for (const auto &element : elements(leafGridView)) {
+    if (element.partitionType() == Dune::OverlapEntity)
```
IshaanDesai: Which part of the code is now actually under this `if` condition? The `continue` statement would mean that the code below will be skipped, right?

two-scale-heat-conduction/macro-dumux/appl/spatialparams.hh
```diff
@@ -104,8 +103,42 @@ public:
     return K;
   }
 
+  void updateCouplingData()
+  {
+    for (const auto &element : elements(this->gridGeometry().gridView())) {
+      if (element.partitionType() == Dune::OverlapEntity)
```
IshaanDesai: Same comment as above.

two-scale-heat-conduction/macro-dumux/appl/spatialparams.hh
```diff
@@ -104,8 +103,42 @@ public:
...
+            couplingParticipant_.getScalarQuantityOnFace("macro-mesh", "k_00", elementIdx);
+        couplingData_[elementIdx][2] =
+            couplingParticipant_.getScalarQuantityOnFace("macro-mesh", "k_01", elementIdx);
+        couplingData_[elementIdx][3] =
+            couplingParticipant_.getScalarQuantityOnFace("macro-mesh", "k_10", elementIdx);
+        couplingData_[elementIdx][4] =
+            couplingParticipant_.getScalarQuantityOnFace("macro-mesh", "k_11", elementIdx);
+      }
+    }
+    // exchange coupling data with other ranks
```
IshaanDesai:
```suggestion
    // Trigger exchange of coupling data between neighboring ranks, if the domain is partitioned
```

IshaanDesai (approved):
Thanks a lot for this! I just have one minor question for my understanding, and one suggestion. Once these are resolved, I can merge.

two-scale-heat-conduction/macro-dumux/appl/main.cc
mathiskelm: The `if` condition itself only applies to the next statement but as this is a `continue`, the condition skips the entire loop content below for this element and proceeds to the next. The condition should rather be `!= Dune::InteriorEntity` to also exclude ghost cells (if the grid has them), and we should iterate over interior elements only in the first place with `elements(leafGridView, Dune::Partitions::interior)`. That should make this clearer.

mathiskelm: Note: For discretization methods with coupling points on faces/vertices of the grid we would have to ensure they are registered exactly once on the "border" partition (border of interior to ovelap/ghost partition). `Dune::Partitions::interiorBorder` would include such entities in the loop (elements can't lie on the border so it doesn't matter in this case) but we would still need to register only once.

PR #664, opened 2025-08-25
Fix negotiation of dt value in two-scale-heat-conduction/macro-dumux

IshaanDesai:
The negotiation of the time step value from the time step recommended by preCICE and the time step recommended by DuMuX is correctly handled. The incorrect implementation was noticed when the case was scaled up, leading to DuMuX suggesting a time step value lesser than the one recommended by preCICE.

Checklist:

### DIFF
--- a/two-scale-heat-conduction/macro-dumux/appl/main.cc
+++ b/two-scale-heat-conduction/macro-dumux/appl/main.cc
@@ -223,14 +223,18 @@ int main(int argc, char **argv)
   // initialize
   couplingParticipant.initialize();
 
-  // get some time loop parameters
+  // time loop parameters
   const auto tEnd      = getParam<Scalar>("TimeLoop.TEnd");
   double     preciceDt = couplingParticipant.getMaxTimeStepSize();
+  double     solverDt;
   double     dt;
-  if (getParam<bool>("Precice.RunWithCoupling") == true)
-    dt = preciceDt;
-  else
+
+  if (getParam<bool>("Precice.RunWithCoupling") == true) {
+    solverDt = getParam<Scalar>("TimeLoop.InitialDt");
+    dt       = std::min(preciceDt, solverDt);
+  } else {
     dt = getParam<Scalar>("TimeLoop.InitialDt");
+  }
 
   // instantiate time loop
   auto timeLoop = std::make_shared<TimeLoop<Scalar>>(0.0, dt, tEnd);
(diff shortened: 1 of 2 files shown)


two-scale-heat-conduction/macro-dumux/appl/main.cc
```diff
@@ -265,6 +269,10 @@ int main(int argc, char **argv)
         xOld = x;
       }
 
+      preciceDt = couplingParticipant.getMaxTimeStepSize();
+      solverDt  = std::min(nonLinearSolver.suggestTimeStepSize(timeLoop->timeStepSize()), getParam<Scalar>("TimeLoop.MaxDt"));
```
mathiskelm:
```suggestion
      solverDt  = std::min(nonLinearSolver.suggestTimeStepSize(timeLoop->timeStepSize()), timeLoop->maxTimeStepSize());
```
together with ```timeLoop->setMaxTimeStepSize(getParam<Scalar>("TimeLoop.MaxDt"));``` when creating the time loop.

two-scale-heat-conduction/macro-dumux/appl/main.cc
```diff
@@ -297,22 +305,17 @@ int main(int argc, char **argv)
     // advance precice
     if (getParam<bool>("Precice.RunWithCoupling") == true) {
       couplingParticipant.advance(dt);
```
mathiskelm: `nonLinearSolver.solve(x, *timeLoopo);` may solve for a smaller time-step size than requested if the first attempt fails. The following fix does not account for the coupling data being different for this new time, follow-up required to read appropriate data in newton solver.
```suggestion
      couplingParticipant.advance(timeLoop->timeStepSize());
```

mathiskelm (changes requested):
- Accounting for the initial time-step size set for the DuMux solver is nice.
- Shifting the negotiation of time-step size from "after" the "read - solve -write - advance" block to before it, doesn't really change the behavior, except for the non-obvious side effect of increasing the time-step size suggested by DuMux by calling `nonLinearSolver.suggestTimeStepSize` before any attempts at solving the nonlinear system.
- The core issue of preCICE advancing with a larger time-step size is caused by the `nonLinearSolver` silently reducing the time-step size for the current time step if a first attempt fails (Newton doesn't converge). This is resolved by using `couplingParticipant.advance(timeLoop->timeStepSize())`, but the solve used the coupling data read for the original time-step size. Follow-up in the adapter to re-read coupling data in the newton solver.

mathiskelm (approved):

PR #665, opened 2025-08-26
[two-scale-heat-conduction] Fix parallel macro: Write data only for interior elements

mathiskelm:
Coupling data was still written from the solution vector, not limited to interior elements / coupling points.

Checklist:


IshaanDesai (approved):
I tried out this patch and checked the results and they are now fixed.

PR #666, opened 2025-08-31
Fix subcycling / checkpointing for two-scale-heat-conduction participant macro-dumux

mathiskelm:
Checkpointing was mixed up with normal DuMux time-step advances, breaking the macro-dumux participant in the case of subcycling. Fixed checkpointing and subcycling and cleaned up the time loop.

Checklist:


IshaanDesai:
The results look as what we would expect them to. I would suggest leaving the smaller initial time step value for `macro-dumux` as it is now, which induces subcycling. That makes this tutorial a stronger test case. Changing the end time of the simulation is also not problematic in my opinion.

IshaanDesai (approved):

PR #673, opened 2025-10-09
Changing wording of Calculix case in perpendicular-flap

IshaanDesai:
The reason why the result for the perpendicular-flap tutorial using CalculiX as a solid solver is different than when FEniCS or deal.II is used, is misleading. It was pointed out by a user in [this post](https://precice.discourse.group/t/recommended-solid-solver-with-3d-functionality/1094/5?u=ishaandesai). This PR changes the comment to clarify the disparity.

Checklist:

- [ ] ~~I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.~~

### DIFF
--- a/perpendicular-flap/README.md
+++ b/perpendicular-flap/README.md
@@ -89,7 +89,7 @@ You should get results similar to this one:
 
 Reasons for the differences:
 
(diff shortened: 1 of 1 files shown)


MakisH (approved):
This is adding more detail and removes the "only supports", which looks like it can only help. Thanks!

PR #680, opened 2025-11-12
Python tutorials: Allow skipping venv setup

### LINKED ISSUE #679: Option to skip installing Python dependencies in run scripts
While automatically installing the dependencies for Python-based tutorials is convenient for some use cases (new user runs such a tutorial on their system), it is inconvenient when using a system already prepared. This includes:

1. training systems (USB/VM), where we provide a virtual environment already prepared, and which should work offline.
2. system tests, where we want to test specific versions (see also https://github.com/precice/tutorials/issues/584).

We could provide an option to these scripts to skip installing the dependencies (e.g., `--skip-setup`).

This would simplify a few aspects, and would be an easy solution for #584.

Also related to #547 and #526.

I want to work on this soon.

MakisH:
closes #679

Checklist:


fsimonis:
The amount of extra shellscript for this is pretty daunting. 

I commented on the issue that an environment variable would be way easier to integrate.
If you want to use shellscript for this, then it would be very nice to refactor this into a single file and include it somehow.

MakisH:
Very good point, an environment variable would make everything easier and would not require input from the user. I will adapt.

MakisH:
Switched to:

```bash
if [ -z "${PRECICE_TUTORIALS_NO_VENV:-}" ]
then
    python3 -m venv .venv
    . .venv/bin/activate
    pip install -r requirements.txt && pip freeze > pip-installed-packages.log
fi
```

(we already use this form just fine in the [OpenFOAM adapter Allwmake](https://github.com/precice/openfoam-adapter/blob/develop/Allwmake))

Examples:

- `./run.sh`: Installs and runs
- `PRECICE_TUTORIALS_NO_VENV="" ./run.sh`: Installs and runs
- `PRECICE_TUTORIALS_NO_VENV="true" ./run.sh`: Skips install, but runs

Notes:

- Because we set `-u`, we cannot use the `-v` option without additional handling. Calling `./run.sh` would give `PRECICE_TUTORIALS_NO_VENV: unbound variable`. A workaround based on a default value (`:-`) would defeat the purpose.
- Kept the built-in `[` test instead of the bash `[[` test for consistency with the rest of the scripts (not needed here anyway).

changelog-entries/680.md
```diff
@@ -0,0 +1 @@
+- Added the `PRECICE_TUTORIALS_NO_VENV` environment variable to skip the setup of a Python venv in the run.sh scripts of Python-related tutorials [#680](https://github.com/precice/tutorials/pull/680)
```
fsimonis: There should be an explanation on how to use this env.

channel-transport-reaction/chemical-fenics/run.sh
```diff
@@ -4,9 +4,12 @@ set -e -u
 . ../../tools/log.sh
 exec > >(tee --append "$LOGFILE") 2>&1
 
-python3 -m venv --system-site-package .venv
-. .venv/bin/activate
-pip install -r requirements.txt && pip freeze > pip-installed-packages.log
+if [ -z "${PRECICE_TUTORIALS_NO_VENV:-}" ]
```
fsimonis: This is confusing when setting the env to an empty value.

```console
$ export PRECICE_TUTORIALS_NO_VENV
$ [ -z "${PRECICE_TUTORIALS_NO_VENV:-}" ] && echo "Y" || echo "N" 
Y
$ unset PRECICE_TUTORIALS_NO_VENV
$ [ -z "${PRECICE_TUTORIALS_NO_VENV:-}" ] && echo "Y" || echo "N" 
Y
$ export PRECICE_TUTORIALS_NO_VENV
$ [ -z "${PRECICE_TUTORIALS_NO_VENV:-}" ] && echo "Y" || echo "N" 
N
```

Why not use `-v`? Then you also don't need a default and to wrap it in "".

MakisH: Because

```bash
#!/usr/bin/env bash
set -e -u

if [ -v "${PRECICE_TUTORIALS_NO_VENV}" ]
then
    echo "foo"
fi
```

gives:

```
./run.sh: line 4: PRECICE_TUTORIALS_NO_VENV: unbound variable
```

and we want the `set -e -u` for further side effects.

fsimonis: `-v` checks if the pass env name exists. The invocation is:

```bash
if [ -v PRECICE_TUTORIALS_NO_VENV ]
then
    echo "foo"
fi
```

MakisH: Right... I switched it to that.

That would mean using it as `set PRECICE_TUTORIALS_NO_VENV`. If one misused it as `PRECICE_TUTORIALS_NO_VENV="false" ./run.sh` (which is allowed and equivalent to `PRECICE_TUTORIALS_NO_VENV="anything" ./run.sh`), it would be doing something unexpected, but a double negation would be strange in the first place.

Anyway, this is something meant mainly for internal use.

fsimonis (approved):

PR #684, opened 2025-12-01
Add micro-dumux-surrogate participant to two-scale heat conduction

IshaanDesai:
This PR adds a surrogate variant to the existing participant micro-dumux, called the micro-dumux-surrogate. The surrogate is created using the tool [BayesValidRox](https://git.iws.uni-stuttgart.de/inversemodeling/bayesvalidrox/-/tree/master?ref_type=heads). The surrogate is built using PCE (Polynomial Chaos Expansion). This participant is built to highlight the functionality of model adaptivity in the Micro Manager: https://github.com/precice/micro-manager/pull/198

Checklist:


changelog-entries/684.md
```diff
@@ -0,0 +1 @@
+- Added PCE-base surrogate for micro-dumux in two-scale heat conduction [#684](https://github.com/precice/tutorials/pull/684)
```
MakisH, line 1: The README.md file of the tutorial also needs to document this alternative and what kind of surrogate it is.

In other cases (e.g., `perpendicular-flap`), we had participants named `-fake`? Is this case doing something similar?

two-scale-heat-conduction/micro-dumux-mada/mada_switcher.py
MakisH, line 1: What is this directory, and how is it related to the `micro-dumux-surrogate/`?
What does `mada` stand for?
Why is there no `run.sh` and `clean.sh`, if this is a case directory?

two-scale-heat-conduction/micro-dumux-mada/micro_sim_sur.py
```diff
@@ -0,0 +1,56 @@
+"""
+Micro simulation Surrogate, requrie previous computation of surrogate model
```
MakisH:
```suggestion
Micro simulation Surrogate, requires previous computation of surrogate model
```

?

two-scale-heat-conduction/micro-dumux-surrogate/run-surrogate-workflow.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e -u
+
+python3 -m venv .venv
+. .venv/bin/activate
+
+pip install -r requirements.txt
```
MakisH, line 7: See https://github.com/precice/tutorials/pull/680 for some updates (or directly copy from other `run.sh` in the same tutorial).

MakisH (commented):
I have not yet tried to run it, but here are some first, easy-to-address comments.

IshaanDesai:
Thanks for the comments. I am still thinking about how to structure the folders so that only the full-order model and the surrogate model are exposed to the user. By adding a surrogate model at the micro scale, we have a unique scenario in which the Micro Manager runs not one but two solvers simultaneously. Technically, the solvers are not participants of the coupling; the Micro Manager is. But having the solvers in individual folders makes sense because it highlights which physics is being used and which software is being run.

uekerman:
> Thanks for the comments. I am still thinking about how to structure the folders so that only the full-order model and the surrogate model are exposed to the user. By adding a surrogate model at the micro scale, we have a unique scenario in which the Micro Manager runs not one but two solvers simultaneously. Technically, the solvers are not participants of the coupling; the Micro Manager is. But having the solvers in individual folders makes sense because it highlights which physics is being used and which software is being run.

Makes sense to me. The only real alternative would be a subfolder `micro` or `micro-manager` and in there multiple folders with the different models.

two-scale-heat-conduction/micro-dumux-surrogate/requirements.txt
```diff
@@ -0,0 +1,3 @@
+numpy
+bayesvalidrox
+micro-manager-precice
```
MakisH, line 4: Recent update: Please also add `pyprecice` here, even if implied by the `micro-manager-precice`.

two-scale-heat-conduction/micro-dumux-surrogate/run.sh
MakisH, line 1: Recent: This will need some updates regarding the venv setup. See how the other tutorials now (consistently) handle that.

MakisH (commented):
Since there is now activity on this PR, notice some recent changes in the rest of the tutorials, regarding:

- [ ] venv
- [ ] `requirements.txt`
- [ ] `metadata.yaml`

You could also directly try to update the `tests.yaml` to add a test case.

Don't forget to update the tutorial `README.md` file to add the cases to the overview.

Issue #674, opened 2025-10-09
Systemtest summary not properly escaped


fsimonis:
The systemtest workflows add commit messages to the summaries.
These commit messages are not escaped though.
This leads to backticks in the commit messages being interpreted as subshells, leading to various problems.

https://github.com/precice/precice/actions/runs/18374238279/job/52345314741

fsimonis:
Using single quotes seems to work. https://github.com/precice/precice/pull/2373/commits/59f064429fb26eb8221bc6d249983ed709f21bf0

MakisH:
I removed the respective feature in #842, thus this is currently not relevant, as only URLs are given. I might re-add it in the future, in a simpler and safer way.

PR #694, opened 2026-01-14
System tests: Escape strings in commit info

MakisH:
~~Replaces double quotes with single quotes in the Markdown summary tables, as suggested in #674.~~

The main table (with `printf '...'`) is still failing with the `'` in `Eigen's` in https://github.com/precice/precice/commit/40a98c3f0cf5543eae2352d0dd09b9424c486462

Error: https://github.com/precice/tutorials/actions/runs/20990221387/job/60333606849?pr=694

Tried with `printf '%q\n' "message"`, but then this does not handle backticks, and it messes up with the Markdown rendering. It looks like we cannot easily handle both backticks and single quotes.

### DIFF
--- a/.github/workflows/system-tests-latest-components.yml
+++ b/.github/workflows/system-tests-latest-components.yml
@@ -80,49 +80,49 @@ jobs:
       - id: summary
         name: Prepare Markdown summary
         run: |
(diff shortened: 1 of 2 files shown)


MakisH:
Not relevant since #842.

PR #697, opened 2026-01-19
Update reference results

### LINKED ISSUE #2432: Regression in RBF mapping
The system tests:

- fail since #2389 ([logs](https://github.com/precice/tutorials/actions/runs/20994882470))
- succeed until that point (commit 4c55add, [logs](https://github.com/precice/tutorials/actions/runs/20999477197))

This does not mean that the new state is worse than the previous one, but we need to confirm what is going on. If we accept the new state, we need to update the reference results, before more regressions accumulate.

Related:

- [system tests documentation](https://precice.org/dev-docs-system-tests.html#understanding-what-went-wrong)
- [issue on the numerical threshold we use for the comparisons](https://github.com/precice/tutorials/issues/393) (currently 3e-7)

The workflow to update the reference results is to trigger a workflow on a branch, typically the same PR that brings the changes. For that reason, it is easier if every non-trivial PR passes the system tests before merging ((re-)add the `trigger-system-tests` label).

I would suggest a closer look into the states before and after #2389. If the differences are as intended, we can update the reference results.

MakisH:
To the state after merging https://github.com/precice/precice/pull/2389

Discussion in and closes https://github.com/precice/precice/issues/2432

Workflow triggered: https://github.com/precice/tutorials/actions/runs/21145585207/job/60810248331

### DIFF
--- a/elastic-tube-1d/reference-results/fluid-cpp_solid-cpp.tar.gz
+++ b/elastic-tube-1d/reference-results/fluid-cpp_solid-cpp.tar.gz
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:45966fe65edaa117964c3977632e8c4589645f86de3cf592a64560225ae997f6
-size 706971
+oid sha256:9b66094c7cf9cc8e58892e31474b31eab1dc2b6efd29444c8c7d6801e3f6ef3c
+size 706900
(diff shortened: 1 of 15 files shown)


MakisH:
Merging as the discussion in https://github.com/precice/precice/issues/2432#issuecomment-3773296571 converged.

Other repositories that use the system tests in their workflows will need to change the version of preCICE used:

```
PRECICE_REF: "40a98c3" # develop, after https://github.com/precice/precice/pull/2389
```

PR #696, opened 2026-01-16
Add tutorials with DuMux and Micro-Manager into system tests

Fujikawas:
This PR adds dumux, dumux-adapter and Micro Manager into the docker image for system tests and adds dumux-only combinations of the tutorial [free-flow-over-porous-media](https://precice.org/tutorials-free-flow-over-porous-media-2d.html) and [two-scale-heat-conduction](https://precice.org/tutorials-two-scale-heat-conduction.html) into system test as test suite `dumux_test`

Checklist:

- [ ] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.
- [ ] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #708, opened 2026-02-06
Add flow-over-heated-plate-two-meshes test case

MakisH:
Our [flow-over-heated-plate-two-meshes tutorial](https://precice.org/tutorials-flow-over-heated-plate-two-meshes.html) has been broken since a few months. This PR adds it to the system tests (follow-up of https://github.com/precice/calculix-adapter/pull/148). It serves as a test case for the CalculiX adapter with one data field/direction exchanged per mesh.

@Fujikawas @IshaanDesai since you are currently working on #696, you might be interested in reviewing this PR.

Checklist:

- [x] Triggered the generation of reference results: https://github.com/precice/tutorials/actions/runs/21760188808
- [x] Waiting for the system tests to add reference results.

### DIFF
--- a/tools/tests/tests.yaml
+++ b/tools/tests/tests.yaml
@@ -30,6 +30,11 @@ test_suites:
           - fluid-openfoam
           - solid-openfoam
         reference_result: ./flow-over-heated-plate-nearest-projection/reference-results/fluid-openfoam_solid-openfoam.tar.gz
+      - path: flow-over-heated-plate-two-meshes
+        case_combination:
+          - fluid-openfoam
+          - solid-calculix
+        reference_result: ./flow-over-heated-plate-two-meshes/reference-results/fluid-openfoam_solid-openfoam.tar.gz
   openfoam_adapter_release:
     tutorials:
       - path: flow-over-heated-plate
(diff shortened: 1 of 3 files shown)


tools/tests/tests.yaml
```diff
@@ -18,6 +18,11 @@ test_suites:
           - fluid-openfoam
           - solid-openfoam
         reference_result: ./flow-over-heated-plate/reference-results/fluid-openfoam_solid-openfoam.tar.gz
+      - path: perpendicular-flap
+        case_combination:
+          - fluid-openfoam
+          - solid-calculix
+        reference_result: ./perpendicular-flap/reference-results/fluid-openfoam_solid-calculix.tar.gz
```
MakisH, line 54: This should have been here already, I just took the chance to add it (I previously added the new case here as well, but then decided to remove it as it does not make any difference to the normal flow-over-heated-plate for the OpenFOAM adapter).

Fujikawas (approved):
Why does this PR also result in changes of some other reference results?

MakisH:
> Why does this PR also result in changes of some other reference results?

The respective workflow at the moment always updates all results. You are right that this looks strange and is unnecessary, I opened an issue: https://github.com/precice/tutorials/issues/711

IshaanDesai (approved):
Had a brief look. Aside from the confusing part of other tutorials' reference results being updated, the additions to the tutorial in question make sense.

MakisH:
I cleaned up this branch and removed the previous reference results, but I now have an issue checking out the respective Git ref in the runner to generate the new reference results. I will add them directly on develop.

PR #709, opened 2026-02-09
Update the links to the source page for free-flow-over-porous-media case

Fujikawas:
Checklist:

### DIFF
--- a/free-flow-over-porous-media/README.md
+++ b/free-flow-over-porous-media/README.md
@@ -6,7 +6,7 @@ summary: Flow-flow coupling example with porous media field and free flow field.
 ---
 
 {% note %}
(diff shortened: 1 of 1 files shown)


MakisH (approved):
Good catch. There is an inconsistency between the directory name and the webpage URL here. I guess your are planning a 3D case as well, correct?

Fujikawas:
> Good catch. There is an inconsistency between the directory name and the webpage URL here. I guess your are planning a 3D case as well, correct?

Not really, the related research about complex interface processes is also in 2D.

free-flow-over-porous-media/README.md
MakisH: Shall we maybe also change the title and URL to remove the 2D, then?
If yes, I can update the website as well.

```suggestion
title: Free flow over porous media
permalink: tutorials-free-flow-over-porous-media.html
```

MakisH (approved):

PR #713, opened 2026-02-16
Use configuration reader in dumux-adapter

Fujikawas:
With the changes in https://github.com/precice/dumux-adapter/pull/68, it's not allowed to hardcode dumux-adapter configuration and the config-reader should be used. This PR applies this change to the relevant tutorial solvers.

Checklist:


IshaanDesai (approved):
We tested both tutorial cases using https://github.com/precice/dumux-adapter/pull/68, and the results were as expected.

PR #714, opened 2026-02-19
Add 3D case for partitioned heat conduction

NiklasVin:
- compared to existing heat conduction examples, it is in 3D and helps to validate vector valued and 3D coupling
- what is necessary to run this new case: FEniCSx v0.10, preCICE v.3.3.1, preCICE python bindings v.3.3.1 and FEniCSx-preCICE adapter (currently the most recent state on the develop branch)
- Output at t=1:
Screenshot from 2026-02-19 21-21-36

Checklist:


IshaanDesai (commented):
The case setup and folder structure look good. I think the tutorial is valuable as we currently do not have any other way of testing the FEniCSx adapter in 3D.

I remember there were some issues with data-mapping accuracy. Can you please briefly describe the issues, which mapping settings you tried, and what effect each had?

NiklasVin:
I dont remember every mapping I used but I ran a few setups similar to them I tried in the past. I figured out that the mapping defined for the Dirichlet participant (i.e. read from Neumann Mesh) primarily influenced the accuracy of the result (I suppose the reason is that the Dirichlet participant requires BC function of degree 2, and the Neumann only degree 1):
* `nearest-neigbor`: Does not really work (error after first time step `3.8e-1`)
* `rbf-pum` with `compact-polynomial-c2`: Better but the error is also rather large after the first time step (magnitude of about `e-3`)
* `rbf-pum` with `compact-polynomial-c8`: The best I could find (the c4, c6 are also better than the c2 option)
* For the `rbf-pum`, I noticed that you need to tremendously increase `relative-overlap`, `vertices-per-cluster` and `support-radius` from the default values to get somewhat accurate results.

Generally I am not very happy with the results because setting the dofs directly from the preCICE data yields better results (I think the error was within the magnitude of `e-7`). As far as I can assess this situation the solver code is correct but the mapping & interpolation steps break the accuracy...

NiklasVin:
I ran more tests to further investigate how the mapping configuration changes the accuracy (only to `t=0.1`):
| Mapping Neumann → Dirichlet | Mapping Dirichlet → Neumann | L2-error (Dirichlet)|
|:---------:|:---------:|:---:|
| analytical BCs  | nearest-neighbor | 9.92E-15 / Neumann: 1.16E-14
| nearest-neighbor  | " | 0.0423
| nearest-projection  | " | not supported by adapter
| thin-plate-splines*  | "| 0.000464
| compact-tps-c2*  | "| >1e-4
| multiquadratics (sup.-rad. 1)*  | "|1.8E-06
| multiquadratics (sup.-rad. 1.25)*  | "|4.16E-07
| inv. multiquadratics (sup.-rad. 1)*  | "|6.13E-06
| inv. multiquadratics (sup.-rad. 1.2)*  | "|1.68E-06
| compact-polynomial-c2 (sup.-rad. 2.5)*  | "|0.000305
| compact-polynomial-c4 (sup.-rad. 2.5)*  | "|9.7E-05
| compact-polynomial-c6 (sup.-rad. 2.5)*  | "|4.69E-05
| compact-polynomial-c8 (sup.-rad. 2.5)*  | "|2.81E-05
| compact-polynomial-c8 (sup.-rad. 1)*  | "|0.000475
| compact-polynomial-c8 (sup.-rad. 10)*  | "|2.93E-07

All results are computed with `consistent` constraint. Entries marked with a * are computed with the `rbf-global-direct` mapping.

The current mapping configuration committed on GitHub uses `rbf-pum-direct`, with incredibly large `relative-overlap`, `vertices-per-cluster` and `support-radius`, which is I guess a sign for bad configuration.

@davidscn Since I’m not very familiar with finding the most sensible mapping, could you maybe take a look at what could be changed in the mapping configuration to reduce the error?

davidscn:
I am not sure I understand the entire discussion.  A few points that come to my mind

You replace the Neumann participant by an analytical function with matching meshes and a nearest neighbor mapping, you get something in the order of 1e-14, i.e., the first row, correct? Or do you replace both participants? What does analytical mean and how does the coupling look like? Which participants do you replace? Is it unidirectional? To isolate the mapping error, it would be useful to have an analytical function at nonmatching points.
In all other rows: do you replace the analytical function by nonmatching participants and use a mapping in preCICE. How do you make the setup nonmatching then?

- which accuracy do you expect from the RBF mapping and why? 
- if you want to tune the accuracy of the mapping, I would pay more attention to the support-radius configuration of a given RBF, not so much the RBF itself. Direct RBF vs PUM does not matter all that much at the moment. I would plot the error over the support radius config, and this yields generally an optimum curve.
- I have a very similar 2D experiment, which you may want to look at to get a feeling for what to expect.

NiklasVin:
Thanks for your reply:)
In the first row, the setup is basically unidirectional. There is no data mapping from Neumann to Dirichlet but only from Dirichlet to Neumann. And on the Dirichlet side, instead of the preCICE data, the manufactured solution (or "analytical solution") is used as boundary values at the coupling interface. Then, you can see that both the solvers are correctly implemented and the issue is not the nearest-neighbor mapping from Dirichlet to Neumann.
The Neumann mesh is basically a subset of the Dirichlet mesh, because the Neumann mesh is defined by the vertices of the triangles located on the coupling surface whereas the Dirichlet mesh additionally defines grid points in the middle between two vertices as well. So the direction Neumann → Dirichlet is nonmatching by default.
Since this is similar to the 2D case of the partitioned heat equation and results within machine precision can be obtained for this case, I expected that this can also be achieved here?

Edit: Maybe for a better understanding, the errors of the Dirichlet mesh are the following:
image

image

On top, the error more or less looks identical for all different compact polynomials with different support radii.
And on the bottom the error for nearest neighbor mapping, as expected, the error is nearly zero on the vertices but gets larger in the middle of each edge.

uekerman:
Trying to summarize a bit:
- Neumann is the inner domain, Dirichlet the outer.
- The Neumann mesh is a subset of the Dirichlet mesh.
- The non-machine precision comes from the Neumann -> Dirichlet mapping

> Since this is similar to the 2D case of the partitioned heat equation and results within machine precision can be obtained for this case, I expected that this can also be achieved here?

In general, I would not expect machine precision here; 1e-7 seems already rather small. To which case are you referring exactly? The manufactured solution has only up to second-order terms in space, right? IIRC, we explicitly fitted a quadratic function at some point. Is this maybe still used in the 2D case? Then, you get to machine precision. But only with RBFs + linear polynomial, I don't see how.

What exactly do you want to show with this case? Matching meshes or non-matching meshes?

What are the space dimensions of the domain? (This info could also be good for the README)

NiklasVin:
The primary goal of this tutorial is to verify that the FEniCSx adapter can deal with scalar- & vector-valued coupling in 3D. Since we wanted to add this feature to the software update paper, it would have been a perfect argument that the feature works as expected if the error was within machine precision.

My confusion originated from the fact that the 2D FEniCSx-FEniCSx partitioned heat equation case with nearest-neighbor mapping and no quadratic fit of a function, resulted in a solution within machine precision even though the Neumann mesh is a subset of the Dirichlet mesh, similar to the 3D case. Therefore, I thought my mapping configuration of this case is just chosen badly. But I realised just now that I silently defined matching meshes for the 2D case - which in the 3D case is not easily doable - and now, with the non-matching mesh errors of the 2D partitioned heat equation case, the errors of the 3D case make more sense to me.
It is good to hear that 1e-7 is already small.
I am sorry for my confusion caused by myself 😅

Just for completeness, I tested another domain decomposition:
Screenshot from 2026-05-26 17-33-46
Screenshot from 2026-05-26 17-47-50

IshaanDesai (approved):
Now that there is clarity about the error, we can proceed with finalizing this PR.

changelog-entries/714.md
```diff
@@ -0,0 +1 @@
+- add FEniCSx-based solver for the 3D partitioned heat conduction tutorial
```
MakisH:
```suggestion
- Added a FEniCSx-based solver for the 3D partitioned heat conduction tutorial [#714](https://github.com/precice/tutorials/pull/714)
```

- Changelogs are written from the perspective of "what happened in that release".
- A link to the PR helps find out more.

However, since this is a new tutorial:


```suggestion
- Added a 3D partitioned heat conduction tutorial with FEniCSx [#714](https://github.com/precice/tutorials/pull/714)
```

partitioned-heat-conduction-3d/neumann-fenicsx/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -e -u
+
+python3 -m venv --system-site-packages .venv
```
MakisH: There have been updates in how Python-based tutorials handle venvs. See, e.g., https://github.com/precice/tutorials/blob/develop/elastic-tube-1d/fluid-python/run.sh

partitioned-heat-conduction-3d/precice-config.xml
MakisH, line 1: A `metadata.yaml` file is now also expected from every tutorial. See an example: https://github.com/precice/tutorials/blob/develop/partitioned-heat-conduction/metadata.yaml

FEniCSx remains commented-out until we implement the respective component in the system tests.

partitioned-heat-conduction-3d/README.md
MakisH, line 1: Remember to also prepare the respective changes to the website. See the latest PR template: https://github.com/precice/tutorials/blob/develop/.github/pull_request_template.md

MakisH (commented):
Hi there! I added some first comments on what has very recently changed; they are all easy to implement.

I have not yet tried to run the code, and the `run.sh` scripts are expected to change.

I am optimistic that we can get this merged in the next couple of days.

partitioned-heat-conduction-3d/precice-config.xml
```diff
@@ -0,0 +1,80 @@
...
+      constraint="consistent" />
+  </participant>
+
+  <m2n:sockets acceptor="Dirichlet" connector="Neumann" exchange-directory=".." />
+
+  <coupling-scheme:serial-implicit>
+    <participants first="Dirichlet" second="Neumann" />
+    <max-time value="1.0" />
+    <time-window-size value="0.1" />
+    <max-iterations value="200" />
```
MakisH: The 2D `partitioned-heat-conduction` tutorials set 100 max-iterations. Why 200 here?


```suggestion
    <max-iterations value="100" />
```

partitioned-heat-conduction-3d/precice-config.xml
MakisH: Added.

partitioned-heat-conduction-3d/neumann-fenicsx/run.sh
MakisH: Updated.

partitioned-heat-conduction-3d/solver-fenicsx/problem_setup.py
```diff
@@ -0,0 +1,131 @@
+"""
+Problem setup for partitioned-heat-conduction/fenicsx tutorial
+"""
+from dolfinx.io import gmsh as gmshio
+from dolfinx.mesh import DiagonalType, create_box
+import dolfinx.mesh
+from my_enums import DomainPart
+import numpy as np
+from mpi4py import MPI
```
MakisH, line 9: MPI is not used. Remove?


```suggestion
```

partitioned-heat-conduction-3d/solver-fenicsx/problem_setup.py
```diff
@@ -0,0 +1,131 @@
+"""
+Problem setup for partitioned-heat-conduction/fenicsx tutorial
+"""
+from dolfinx.io import gmsh as gmshio
+from dolfinx.mesh import DiagonalType, create_box
+import dolfinx.mesh
```
MakisH, line 6: `mesh` itself is not used. Remove?


```suggestion
```

partitioned-heat-conduction-3d/solver-fenicsx/requirements.txt
```diff
@@ -0,0 +1,4 @@
+numpy >1, <2
+fenicsxprecice
+scipy
+sympy
```
MakisH, line 4: Since `problem_setup.py` calls `import gmsh`:

```suggestion
sympy
gmsh
```

partitioned-heat-conduction-3d/precice-config.xml
MakisH: It looks like the iterations are always 88, so I think we can indeed restore this.

partitioned-heat-conduction-3d/README.md
MakisH: Started a PR: https://github.com/precice/precice.github.io/pull/929

partitioned-heat-conduction-3d/solver-fenicsx/problem_setup.py
MakisH: Opened an issue: https://github.com/precice/tutorials/issues/833

MakisH (approved):
I directly applied some feedback and further fixes, and I integrated the tutorial into the system tests, implementing the respective component. Since this branch is in a fork, I will add the reference results directly in develop after merging. I have checked that this works.

Nice to have a 3D case for FEniCSx!

PR #717, opened 2026-02-22
Handle missing reference results gracefully in Systemtest field comparison

### LINKED ISSUE #404: Generate error message instead of exception when reference results are not found
Currently, when running the tests without reference results, the user gets an exception about file not found. While this is clear enough, and since this will be a common case, it would be nice if that would be an actual error message.

ZELTROX1:
Summary:
This change improves the user experience when reference result archives are missing or invalid during system test execution.

Previously, missing reference archives caused a raw exception (for example FileNotFoundError).
Now the test fails cleanly with a clear error message.

Fixes #404 

Checklist:

- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/404.md`.


MakisH:
Thank you @ZELTROX1. I have not yet looked in detail, but it makes sense overall.

Could you please demonstrate with some logs?

I am also asking Copilot for a review. Please ignore anything that doesn't make sense to you, incorporate anything that you think makes sense.

ZELTROX1:
This is recreated issue without code change
image
This is after the change
image
I've added the screenshot of the logs as requested. Let me know if this helps clarify things, or if you need the raw text output to dig into it further!"

There are some helpful point that the copilot mentioned on docstring and  stderr.log/CI output i will make the change accordingly and raise a pr

ZELTROX1:
Requesting a review after the highlighted issue fixes

MakisH (approved):
Thank you for following up! Everything looks good, and I tested this locally.

Issue #617, opened 2025-02-17
Tutorial needed for Python Actions


MakisH:
We already have an example for the elastic-tube-1d in the documentation: https://precice.org/configuration-action.html#python-callback-interface

We could convert this into a proper tutorial.

kishansinghifs1:
Hiii @MakisH can you PTAL : https://github.com/precice/tutorials/pull/719

MakisH:
See discussion in https://github.com/precice/tutorials/pull/719#issuecomment-4420622518

PR #719, opened 2026-02-23
Add Python Actions tutorial based on elastic-tube-1d

kishansinghifs1:
Add Python callback interface tutorial and related scripts for elastic-tube-1d #617 

1. Added a complete Python callback interface workflow (precice-config-callback.xml, solver wrappers, and run scripts) alongside the baseline elastic tube case without modifying existing functionality. 
2. Created pressureRampAction.py with a performAction(...) function that applies a time-dependent pressure ramp (t < 0.2) via the preCICE Python callback interface.

Checklist:

### DIFF
--- a/elastic-tube-1d/README.md
+++ b/elastic-tube-1d/README.md
@@ -43,6 +43,61 @@ Both fluid and solid participant are supported in:
 - *Python*: example solvers using the preCICE [Python bindings](https://precice.org/installation-bindings-python.html). The run script installs these automatically via pip in a virtual environment.
 - *Rust*: example solvers using the preCICE [Rust bindings](https://precice.org/installation-bindings-rust.html). They need `cargo` to be installed.
 
+## Python callback interface tutorial track
+
+This case now includes a runnable callback variant matching the Python callback interface from the preCICE documentation:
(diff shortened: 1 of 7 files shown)


fsimonis:
The Python actions is a feature that is on the slippery slope towards undefined behavior.
There definitely is an issue when [mixing time-window sizes](https://github.com/precice/precice/issues/1865).

We at least need to verify the results of this implementation.

A more strategic question is whether we even want to promote this feature.

@uekerman @MakisH

uekerman:
Verifying the results for his PR should be easy enough. I am rather sure that the Python actions still work here.

That the possibilities of the actions are vast is clear. I agree that we should eventually restrict these or even remove the feature completely. But here is probably not the best place for this discussion.

Despite all this, the solution in this PR seems too complicated IMO. 
A cleaner solution could be to even simply overwrite the elastic tube to always do the ramp?

MakisH (changes requested):
I also think that this tutorial modification is rather complicated (duplicates the `precice-config.xml` and `run.sh`) and that it would need some motivation in the physics.

I do still believe we need a tutorial demonstrating callbacks, and I would also prefer having a tutorial always using a callback.

Since there has not been any further reaction on this, I am closing it. Happy to review any follow-ups.

PR #720, opened 2026-02-23
Fix symlinks of two clean-tutorial.sh scripts

MakisH:
The `flow-around-controlled-moving-cylinder` and `channel-transport-reaction` cases were copying the `clean-tutorial-base.sh` script instead of just linking to it, without any clear reason (the file was essentially the same).

This PR replaces the files with symlinks, as expected.

Interestingly, GitHub thinks that this is a conflict. :thinking:

### DIFF
--- a/channel-transport-reaction/clean-tutorial.sh
+++ b/channel-transport-reaction/clean-tutorial.sh
@@ -1,11 +1,1 @@
-#!/bin/sh
-set -e -u
-
-# shellcheck disable=SC1091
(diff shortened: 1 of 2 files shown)


MakisH:
Fixed directly on develop https://github.com/precice/tutorials/commit/863bb4df872baaf8f0a838f9106d6be7c53cd2e1

(and it doesn't complain about conflicts anymore. Strange.)

PR #723, opened 2026-02-24
Make the get-latest-components workflow conditional.

### LINKED ISSUE #682: Latest components workflow hits API usage limits
[This workflow](https://github.com/precice/tutorials/blob/develop/.github/workflows/system-tests-latest-components.yml) sometimes fails with:

```
API rate limit exceeded for 52.159.245.176. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

We need to set a `token` option to the workflow ([documentation](https://github.com/nmbgeek/github-action-get-latest-commit?tab=readme-ov-file#configuration)).

kishansinghifs1:
Fix #682  - Part 1
1. **Gate the gather-refs job in system-tests-pr.yml** with the trigger-system-tests label check so the workflow skips entirely for unlabeled PRs
   - Added conditional: `if: ${{ github.event.label.name == 'trigger-system-tests' }}`
   - This prevents unnecessary API calls when the workflow isn't explicitly triggered

Checklist:

### DIFF
--- a/.github/workflows/system-tests-latest-components.yml
+++ b/.github/workflows/system-tests-latest-components.yml
@@ -25,6 +25,7 @@ jobs:
           owner: precice
           repo: precice
           branch: develop
+          token: ${{ secrets.GITHUB_TOKEN }}
       - id: ref-python-bindings
         name: Get Python bindings ref
         uses: nmbgeek/github-action-get-latest-commit@main
@@ -32,6 +33,7 @@ jobs:
           owner: precice
           repo: python-bindings
           branch: develop
+          token: ${{ secrets.GITHUB_TOKEN }}
       - id: ref-calculix-adapter
         name: Get CalculiX adapter ref
         uses: nmbgeek/github-action-get-latest-commit@main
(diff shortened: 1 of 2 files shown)


MakisH (commented):
Thank you for the contribution.

The conditional check already worked in this PR.

I am not sure if passing the `GITHUB_TOKEN` would be the right approach here. In https://github.com/precice/tutorials/issues/682, the suggestion was for an app id, which we would need to set up.

Let's split this into two. Please remove the authentication aspects, and we see if that's enough. I think it will be.

Btw, I renamed the PR, please rename again after removing the authentication aspects.

kishansinghifs1:
Hiii @MakisH I was initially thinking of the same but now as you clarified it would be more helpful to me.I'll apply those changes and let you know soon. Thanks for the review!!!!!!!!

kishansinghifs1:
Hi @MakisH I  removed the authentication spec and now you can cleanly merge  this . Thank you !!!!!

kishansinghifs1:
Already committed cleanly #733 we can close this PR as duplicate.

MakisH (approved):
Thanks!

Issue #707, opened 2026-02-06
pre-commit picks up local files not tracked by git


MakisH:
See comment by @davidscn in https://github.com/precice/tutorials/issues/705#issuecomment-3859114780

pre-commit just executes this script, which does not distinguish if the files are staged or not: https://github.com/precice/tutorials/blob/develop/tools/check-image-names.py

The rest of the custom scripts we use should show the same behavior.

We could either guard the scripts, or maybe there are options in pre-commit.

davidscn:
> We could either guard the scripts, or maybe there are options in pre-commit.

I think we could just ask git to give us a list of files.

PR #725, opened 2026-02-25
Fix precommit hooks

PranjalManhgaye:
this PR addresses #707 and adjusts how our validation scripts behave during pre-commit checks.

# Summary

at the moment, scripts like `check-image-names.py`, `check.sh`, and `check-size.sh` also pick up untracked local files. This can cause pre-commit to fail even when those files aren’t part of the repository.

With this change, the scripts now only operate on files that are actually tracked by Git. Untracked local files are ignored.

Changes

* `check-image-names.py` now determines tracked files using `git ls-files` and skips anything untracked.
* `check.sh` and `check-size.sh` have been updated to validate only Git-tracked files.

This keeps the naming and size checks strict for repository content, while avoiding unnecessary failures caused by local development files.

## Impact

* pre-commit no longer fails because of untracked local files.
* Naming and size rules are still enforced for all tracked content.
* The workflow becomes less frustrating for contributors without reducing repository standards.

## Checklist

* [ ] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.
* [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

### DIFF
--- a/tools/check-image-names.py
+++ b/tools/check-image-names.py
@@ -2,9 +2,30 @@
 
 import sys
 import os
+import subprocess
+
+
+def is_tracked(path: str) -> bool:
+    """
+    Return True if the given path is tracked by git, False otherwise.
+
+    This guards the hook against local, untracked files that pre-commit
+    might pick up when running outside of the normal commit workflow.
+    """
+    result = subprocess.run(
+        ["git", "ls-files", "--error-unmatch", path],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    return result.returncode == 0
+
 
 problems = False
 for file in sys.argv[1:]:
+    # Skip files that are not tracked by git
+    if not is_tracked(file):
+        continue
+
     parts = file.split(os.sep)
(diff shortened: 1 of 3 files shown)


MakisH:
@PranjalManhgaye thanks for the contribution! It looks like you included some commits from another branch that were not meant to be here. Could you please sort out the history and push again?

PranjalManhgaye:
sure , thanks @MakisH  for pointing that out. I accidentally branched fix-precommit-hooks off fix-clean-scripts, so the pr included unrelated systemtest/cleaning commits.

MakisH (approved):
I checked locally for:

- markdown issues on untracked (and tracked) files
- image names on untracked (and tracked) images
- image size on untracked (and tracked) images

all work as expected. Thanks!

PranjalManhgaye:
@MakisH looks like the pre-commit check failed due to markdownlint npm install (403). let me know if I should open a fix or re-run,
is it a dependency download issue ?

MakisH:
All good, triggered it again, worked.

PR #728, opened 2026-02-27
Add --suites filter to generate_reference_results

### LINKED ISSUE #711: Workflow to generate reference results always updates all cases
Even if only one new tutorial is added (see #708), all reference results archives are updated, even if this is not needed. Also, since what is updated is binaries, we cannot just check and skip updating identical files.

One easy work-around would be an input argument in the [respective workflow](https://github.com/precice/tutorials/blob/develop/.github/workflows/generate_reference_results_manual.yml) with a file or test case filter.

Geethapranay1:
Closes #711

### Changes

1. **`tools/tests/generate_reference_results.py`** — added `--suites` argument (comma-separated, optional) to filter which test suites to generate reference results for. When not provided, all suites run (backward compatible).
2. **`generate_reference_results_workflow.yml`** — added optional `suites` input. When provided, passes `--suites` to the script and uses selective `git add --update` to only commit modified files. When empty, behavior is unchanged.
3. **`generate_reference_results_manual.yml`** — added `suites` input field for the manual trigger UI.

### Testing

**1. full docker test passed for `--suites` filter**
Screenshot From 2026-02-26 17-56-26

only 3 elastic tube tests ran not all 10

**2. error handling for invalid suite name**
image

**3. backward compatible default behavior (no `--suites`)**
Screenshot From 2026-02-26 18-09-55

**4. `--help` shows a new argument**
image

Checklist:

- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<711>.md`.


Geethapranay1:
hi @MakisH 
this pr solves the #711 issue.
check once and let me know if any further requirements needed in the pr.

tools/tests/generate_reference_results.py
```diff
@@ -99,7 +102,25 @@ def main():
...
+        test_suites = []
+        for name in test_suites_requested:
+            found = all_test_suites.get_by_name(name)
+            if not found:
+                logging.error(f"Did not find the testsuite with name {name}")
+            else:
+                test_suites.append(found)
+        if not test_suites:
+            raise RuntimeError(
+                f"No matching test suites with names {test_suites_requested} found. "
```
MakisH: This gives some info output with `ERROR:` in front of it, but it does not actually exit the script, so it fails silently.

MakisH (commented):
Thanks for contributing! This is useful, especially since I would now restructure our `test.yaml` based on more granular test cases (#767).

I tested it locally, and it works as expected. Just one comment on an error message, which should be easy to fix (I can also just fix it directly).

I am also requesting a Copilot review mostly out of curiosity.

tools/tests/generate_reference_results.py
MakisH: This is actually a broader issue in the whole script. Let's ignore it here.

MakisH (approved):

PR #730, opened 2026-02-27
Make clean.sh scripts executable from any directory

### LINKED ISSUE #667: Cleaning scripts are not callable from the root directory
To clean a case, one has the following scripts available:

- `tutorials/clean-all.sh`
- `tutorials/tutorial/clean-tutorial.sh`
- `tutorials/tutorial/case/clean.sh`

These scripts, however, are only doing the right thing if they are called from the directory in which they are located.

A common scenario is that a user is inside `tutorials/tutorial/case/` but needs to call `tutorials/tutorial/clean-tutorial.sh`. This means that they first need to go to the parent directory, which is rather unergonomic. It would be great to be able to call `../clean-tutorial.sh`.

Similarly, calling `./case/clean.sh` would also be nice.

Feedback from the 2025 workshop.

Eshaan-byte:
## Summary

Extends the fix from #716 to the remaining case-level `clean.sh` scripts and one non-symlinked `clean-tutorial.sh`.

- Added `SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"` and `cd "$SCRIPT_DIR"` to all 111 case-level `clean.sh` files
- Applied the same fix to `partitioned-heat-conduction-direct/clean-tutorial.sh` (real file, not a symlink, so #716 didn't cover it)
- Added missing `set -e -u` to `flow-around-controlled-moving-cylinder/solid-python/clean.sh` for consistency

Closes #667

## Test plan

- Run `../clean-tutorial.sh` from inside a case directory (e.g. `quickstart/fluid-openfoam/`)
- Run `./fluid-openfoam/clean.sh` from a tutorial directory (e.g. `quickstart/`)
- Run `clean-all.sh` from the repository root
- Verify all three work without errors

## Checklist


MakisH:
Hi @Eshaan-byte,

Similar comment as in #716: We want to keep these files as slim as possible. Would it be possible to move this (or probably a similar hack) to `../tools/cleaning-tools.sh`?

Also: No need for an issue before a PR, if the PR is anyway on the way.

Eshaan-byte:
Hi @MakisH, thanks for the review!

The challenge is that clean.sh sources cleaning-tools.sh via a relative path (../../tools/cleaning-tools.sh), which fails before cleaning-tools.sh even loads when called from a different directory.

A possible approach to keep clean.sh slim: change the source line to resolve from $0 instead of $PWD, and let cleaning-tools.sh handle the cd.

clean.sh (only the source line changes):
#!/usr/bin/env sh
set -e -u

. "$(dirname "$0")/../../tools/cleaning-tools.sh"

clean_openfoam .
cleaning-tools.sh (add at the top):

cd "$(cd "$(dirname "$0")" && pwd)"
This way no extra lines are added to clean.sh — it stays the same size. Would this approach work for you?

PranjalManhgaye:
Hi @Eshaan-byte , thanks for the suggestion ,i agree that resolving the source path using dirname "$0" fixes the issue locally.

In the final merged PR #716, i applied a similar idea but made it explicit in each script so they reliably determine their own directory before sourcing anything. This ensures all cleaning scripts are truly callable from any directory, not just when sourcing a helper file. That’s the pattern that got approved  — scripts use:
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
cd "$SCRIPT_DIR"

idea was to make the files slim,
This avoids depending solely on the sourced file to set up the correct working directory and makes each script self-contained and robust across environments.
thanks you pointed ,

MakisH (approved):
Even though the main usability issue was always with the `clean-tutorial.sh` (since one is typically inside a case directory, e.g., `quickstart/fluid-openfoam` and needs to call `../clean-tutorial.sh`), this is still a welcome usability improvement.

I always wanted to have these scripts as slim as possible, but this is not a large addition either.

I tested calling the scripts from various locations, and I did not observe any issue.

I also ported the remaining/newer cases, and I will do further polishing in another PR.

Thanks again for contributing!

PR #732, opened 2026-03-01
Source test cases from other repositories

### LINKED ISSUE #538: Source test cases from other repositories
In the [tests.yaml](https://github.com/precice/tutorials/blob/develop/tools/tests/tests.yaml#L4), we currently assume that everything is inside the tutorials repository:

https://github.com/precice/tutorials/blob/develop/tools/tests/tests.yaml#L4

We still need a way to incorporate cases from other repositories (or even getting plain archives from publications or data repositories).

@valentin-seitz am I overlooking some functionality here that is already implemented?

PranjalManhgaye:
## What this is about 
(closes #538)

right now, every entry in `tests.yaml` assumes the tutorial folder is already in this repo. you give a `path` like `flow-over-heated-plate`, and the runner finds it under the local `precice/tutorials` checkout.

Issue #538 asks for a way to run cases that live somewhere else — another git repo, a fork, or from a direct download link (e.g. a .tar.gz ), without copying the files into this repo first.

This PR adds an optional `source` block in `tests.yaml` for that.

## What happens at runtime

I have not touched the local behaviour : if there is no `source` field, behaviour is the same as today.

If `source` is set:

1. The runner fetches the tutorial (git shallow clone or archive download) into a cache directory (`~/.cache/precice-tutorials` by default, or `PRECICE_EXTERNAL_CACHE_DIR` if you set it).
2. It loads `metadata.yaml` from that folder, same as for in-repo tutorials.
3. It copies the tutorial into the usual run directory and continues with Docker build, `docker compose up`, and fieldcompare.

`TUTORIALS_REF` / `TUTORIALS_PR` still only affect **local** tutorials. For external ones, the version is whatever you put in `source.ref`.

## Example

Git:

```yaml
  source:
    type: git
    url: https://github.com/precice/tutorials.git
    ref: develop
    subdir: .
  case_combination:
    - fluid-openfoam
    - solid-openfoam
  reference_result: ./flow-over-heated-plate/reference-results/fluid-openfoam_solid-openfoam.tar.gz
```

Archive:

```yaml
- path: some-tutorial
  source:
    type: archive
    url: https://example.org/tutorial-bundle.tar.gz
  case_combination: [...]
  reference_result: ./some-tutorial/reference-results/...
```

`type` can be `git` or `archive`. If you omit `source`, it stays local.

For git, if `clone --branch ref` fails, there is a small fallback that tries fetching without a branch and then checks a few common default branch names.

## How it works

```mermaid
flowchart LR
  A[tests.yaml entry] --> B{Has source?}
  B -->|No| C[Use local tutorial]
  B -->|Yes| D[Fetch git or archive]
  D --> E[Resolve tutorial folder]
  E --> F[Copy into run directory]
  F --> G[Docker build and run]
  G --> H[fieldcompare vs reference]
```
## Files touched

- `tools/tests/systemtests/sources.py` — new: fetch + cache + `resolve_tutorial_root`
- `tools/tests/systemtests/TestSuite.py` — read `source`, load external metadata when needed
- `tools/tests/systemtests/Systemtest.py` — copy from resolved path; ref/PR checkout only for local
- `tools/tests/metadata_parser/metdata.py` — `Tutorial.source`, `ReferenceResult.base_dir`
- `tools/tests/README.md` — short how-to
- `changelog-entries/732.md`

I did not add a real `source:` entry to `tests.yaml` yet, so existing suites should behave as before until someone opts in.

Rebased onto current `develop` and fixed the merge conflicts in README / `Systemtest.py` / `TestSuite.py`.

## testing

I ran the usual checks on external-source-fix after rebasing onto develop:
`python3 -m compileall` on tools/tests/systemtests and metadata_parser — no issues.

Loaded all 13 suites from `tests.yaml `— unchanged for entries without source.

For git, I used a smoke test that drops quickstart from the local index and points at precice/tutorials on develop. The tutorial was fetched into PRECICE_EXTERNAL_CACHE_DIR, source.type was git, and the quickstart systemtest (fluid-openfoam + solid-cpp) passed end to end. No tutorials_ref symlink in the run directory.

For archive, I built a quickstart-develop.tar.gz from the same develop checkout (symlinks resolved so scripts are real files), served it via a file:// URL, and ran the same smoke path with source.type: archive. Download, extract, cache, copy, Docker build/run, and fieldcompare all passed. Again, no tutorials_ref.

I also ran the normal quickstart entry from tests.yaml with TUTORIALS_REF:develop so local tutorials still behave as today when the path exists in the repo.

Docker images were already cached, so both external runs were on the order of ~15–20s each; the steps were still the full pipeline.

## Checklist

- [x] Changelog entry in `changelog-entries/732.md`


PranjalManhgaye:
@makish , sir happy to update anything needed before review .

MakisH:
@PranjalManhgaye this seems quite long. Is it still intended for review in its current form?

In the PR description, I would expect some explanation of the whole concept.

Please also resolve the conflicts.

PranjalManhgaye:
thanks @makish , i will take an another look of this pr and will inform you if its indeed necessary to review it .

PranjalManhgaye:
@MakisH , i  updated the description, rebased on develop, resolved the conflicts, and retested everything locally ,Existing suites are unchanged and both external git + archive quickstart paths are passing now. i think everything should be good now, but happy to change anything if you spot something ,,

MakisH (commented):
Thanks for including a configuration example and a Mermaid graph, this helps a lot!

Looking at the example, I think that defining a `source:` in `tests.yaml` makes sense. I also like that it works both for repositories and for archives.

Could you please update your branch to the latest state? I am through with the larger changes I wanted to do for now, and it would be a good time to review this in the next days.

PranjalManhgaye:
@makish i have merged latest develop into the branch , also fixed the conflicts and retested locally , it is ready for a review when you get time in next days ,,

tools/tests/README.md
```diff
@@ -327,6 +327,31 @@ This template defines:
 - `volumes`: Directories mapped between the host and the container. Apart from directories relating to the users and groups, this also defines where to run the cases.
 - `command`: How to run a case depending on this component, including how and where to redirect any screen output.
 
+#### External tutorial sources
```
MakisH: Looks like this section is in the wrong place: Why under `Implementation details / Component templates` and not under `Extending`? I would expect the following:

```
## Extending

### Adding new tests

#### External test sources

### Adding new components

### Adding new repositories that can trigger the tests
```

The content of the sections looks fine and helpful.

tools/tests/README.md
```diff
@@ -327,6 +327,31 @@ This template defines:
...
+
+By default, every `path` must exist in the local `precice/tutorials` checkout. For tutorials maintained elsewhere, add an optional `source` block to the `tests.yaml` entry:
+
+```yaml
+- path: flow-over-heated-plate
```
MakisH: The example should show a complete test suite definition. We currently have:

```yaml
test_suites:
  breaking-dam-2d:
    tutorials:
      - ...
```

Would I write this external test suite still with the `tutorials:` in the front? The original provision was to replace `tutorials:` with some external source.

It is also a bit strange to have the `path:` before the `source:` is defined.

With the current implementation, the easiest would be to replace `tutorials` with `external` for external sources.

An alternative would be to remove the `tutorials` layer and assume that the `source` is the tutorials repository by default.

MakisH (commented):
I tried this locally with a concrete example (from #845) and it seems to work. 🎉 

Still, I think the configuration looks a bit strange/unintuitive at the moment. Let's refine the configuration based on that example.

I have not yet tried generating reference results, which might be an issue, as it would not be easy/always possible for the runner to push to the external repository (some sources are even static archives). We should also not store those reference results in the tutorials repository, as they have nothing to do with the tutorials, and might not be public. For the specific test, I could push reference results to the same repository/branch and use a more relaxed tolerance. Other than that, I would just expect to skip the comparison now (both related to #847).

tools/tests/README.md
PranjalManhgaye: moved to extending => adding new tests => external test sources, thanks

tools/tests/README.md
PranjalManhgaye: updated to use external: key with source before path and a full suite example => should match what you had in mind

PranjalManhgaye:
@makish addressed the review points => tested locally, all existing suites still load fine and the external example parses correctly , for reference results on external tests i'll wire up skip_compare from #847 once that's in rather than storing refs here , let me know if the config shape looks better now ,,

PranjalManhgaye:
@makish I have merged latest develop and pushed - now it should address the conflict and the external `skip_compare`  about which we talked earlier. Ready for another look when you have time.

tools/tests/metadata_parser/metdata.py
```diff
@@ -6,6 +6,9 @@
 import itertools
 from paths import PRECICE_TESTS_DIR, PRECICE_TUTORIAL_DIR
 
+# Import TutorialSource from systemtests.sources (used for external tutorial sources).
```
MakisH: This comment does not add much value, as it is mostly repeating what is already in code. I would remove it:


```suggestion
```

tools/tests/metadata_parser/metdata.py
```diff
@@ -283,13 +286,15 @@ def from_cases_tuple(cls, cases: Tuple[Case], tutorial: Tutorial):
 class ReferenceResult:
     path: Path
     case_combination: CaseCombination
+    base_dir: Optional[Path] = None
```
MakisH: Similarly to previous reviews, the `Optional` is not the current recommended practice. There are more, also in other files.

tools/tests/systemtests/TestSuite.py
```diff
@@ -62,81 +77,64 @@ def from_yaml(cls, path, parsed_tutorials: Tutorials):
...
-                        raise Exception(
-                            f"Could not find the case combination {tutorial_case['case_combination']} in the current metadata of tutorial {tutorial.name}, or it does not define all necessary participants.")
+                suite_def = test_suites_raw[test_suite_name]
+                local_cases = suite_def.get('tutorials', [])
+                external_cases = suite_def.get('external', [])
+                if local_cases and external_cases:
+                    raise ValueError(
+                        f"Test suite '{test_suite_name}' must use either 'tutorials' or "
+                        f"'external', not both."
+                    )
```
MakisH: Why are they not possible to combine? It would be convenient to be able to write:

```yaml
  release:
    tutorials:
      - *breaking-dam-2d_fluid-openfoam_solid-calculix
      - *channel-transport_fluid-openfoam_transport-nutils
      - ...
    external:
      - *partitioned-heat-conduction-robin_lrft-openfoam_right-openfoam
      - *partitioned-heat-conduction-mixedbc_left-openfoam_right-openfoam
```

which would also allow the external cases to be used in the `openfoam-adapter` test suite, for example.

tools/tests/systemtests/sources.py
```diff
@@ -0,0 +1,206 @@
...
+import subprocess
+import tarfile
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional
+
+# Cache directory for fetched tutorials. Can be overridden via PRECICE_EXTERNAL_CACHE_DIR env.
+_DEFAULT_CACHE = Path(os.environ.get("XDG_CACHE_HOME", Path.home() / ".cache")) / "precice-tutorials"
+PRECICE_EXTERNAL_CACHE_DIR = Path(os.environ.get("PRECICE_EXTERNAL_CACHE_DIR", _DEFAULT_CACHE))
```
MakisH, line 21: What is the motivation for this cache?
Similarly for the ` _cache_key` function.

tools/tests/systemtests/sources.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+@dataclass
+class TutorialSource:
+    """Describes where a tutorial is sourced from."""
```
MakisH: Technically, I would only call "tutorials" the test cases in the tutorials repository. To not rewrite the code (since the term is now used in many places), you could modify the comment a bit:


```suggestion
class TutorialSource:
    """Describes where a test case (tutorial) is sourced from (tutorials repository or external source)."""
```

tools/tests/systemtests/sources.py
```diff
@@ -0,0 +1,206 @@
...
+    key = _cache_key("git", url, ref, subdir)
+    checkout = cache_dir / key
+
+    if checkout.exists():
+        try:
+            subprocess.run(
+                ["git", "-C", str(checkout), "fetch", "origin", ref, "--depth", "1"],
+                check=True,
+                capture_output=True,
+                timeout=120,
```
MakisH: There is a short timeout defined globally somewhere else. Why not use that?
Similarly in different places.
No strong opinion on the one or the other approach, but it looks inconsistent.

tools/tests/systemtests/sources.py
```diff
@@ -0,0 +1,206 @@
...
+            text=True,
+            timeout=300,
+        )
+        if result.returncode != 0:
+            # Fallback: branch may not exist (e.g. repo uses develop/master instead of main)
+            # Clone without branch, then fetch and checkout ref (with common aliases)
+            shutil.rmtree(checkout, ignore_errors=True)
+            logging.debug(
+                f"git clone --branch {ref} failed ({result.stderr}), trying clone + fetch"
+            )
```
MakisH: If the `ref` does not exist, shouldn't the user get an error to choose a different `ref`?

In general, I think that parts of this function can be simplified. Better throw an error / let it crash and let the user fix the configuration, rather than trying to guess.

tools/tests/systemtests/sources.py
```diff
@@ -0,0 +1,206 @@
...
+
+        extract_dir.mkdir(parents=True, exist_ok=True)
+        if url.endswith(".tar.gz") or url.endswith(".tgz") or url.endswith(".tar"):
+            with tarfile.open(tmp_path, "r:*") as tf:
+                tf.extractall(extract_dir)
+        else:
+            import zipfile
+
+            with zipfile.ZipFile(tmp_path, "r") as zf:
+                zf.extractall(extract_dir)
```
MakisH, line 151: When downloading a zip file (https://github.com/vidulejs/tutorials/archive/refs/heads/partitioned-heat-conduction-mixed-robin.zip) I am getting a `permission denied` when trying to run the respective `run.sh` scripts in the system tests.

See the `Partitioned heat conduction mixed BC (left-openfoam, right-openfoam)` case (sourced from an archive), which reports `success`, even though the logs show this error: https://github.com/precice/tutorials/actions/runs/28495445035/attempts/5

However, if I just `unzip` and run the tutorial locally, it just works. So the ZIP file seems to have maintained the file permissions. This seems to be a common issue:

https://stackoverflow.com/q/76645832/2254346

https://www.w3reference.com/blog/python-zipfile-removes-execute-permissions-from-binaries/

MakisH (commented):
I had another look and test, and while it generally works, I think some aspects can be simplified. My main doubts were regarding the cache, the restriction to not mix `tutorials:` and `external:`, and some issue with ZIP archives.

Getting there!

tools/tests/systemtests/TestSuite.py
PranjalManhgaye: I have dropped the mutual-exclusion check so a suite can list both tutorials and external. External cases can be defined once with anchors and referenced from release (and other suites). Also, I have updated the README example accordingly.

tools/tests/systemtests/sources.py
PranjalManhgaye: I have added a short explanation in the README: the cache avoids re-cloning/re-downloading the same external source on every run and in CI; And it is keyed by URL, ref, and optional subdir. Override with `PRECICE_EXTERNAL_CACHE_DIR` if needed.

tools/tests/systemtests/sources.py
PranjalManhgaye: I have replaced the hardcoded fetch timeouts with `_FETCH_TIMEOUT`, driven by the same `PRECICE_SYSTEMTESTS_TIMEOUT` env var as `GLOBAL_TIMEOUT` (could not import `GLOBAL_TIMEOUT` directly without a circular import).

tools/tests/systemtests/sources.py
PranjalManhgaye: I have removed the main/develop/master fallback. If `git clone --branch` <ref> fails, we now raise a clear error so the config can be fixed.

tools/tests/systemtests/sources.py
PranjalManhgaye: After the `zip/tar` extract (and on cache hits), we restore execute bits on `*.sh files`. I have also reproduced the `mixed-BC archive `case locally; the external suite passes end-to-end with no permission denied on `run.sh`.

tools/tests/systemtests/sources.py
```diff
@@ -54,7 +57,13 @@ def _cache_key(prefix: str, url: str, ref: Optional[str] = None, subdir: Optiona
     return hashlib.sha256(raw.encode()).hexdigest()[:16]
 
 
-def fetch_git_repo(url: str, ref: str, cache_dir: Path, subdir: Optional[str] = None) -> Path:
+def _restore_shell_script_permissions(root: Path) -> None:
+    """Restore execute bits on shell scripts (zip extraction strips them)."""
+    for script in root.rglob("*.sh"):
+        script.chmod(script.stat().st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
```
MakisH: I understand that this does not restore the permissions (these are not known) - it marks every `.sh` file as executable.

While a good-enough solution for the particular use case, this could be dangerous in other use cases.

Let's adjust the comment.

tools/tests/systemtests/sources.py
MakisH: I understand what it does, but I am wondering why you wanted to implement that. Did you face any particular issue?

I think it is making the code unnecessarily more complex. Most sources will be cloned only once or a couple of times.

tools/tests/systemtests/sources.py
PranjalManhgaye: I have renamed it to `_ensure_shell_scripts_executable` and reworded the docstring to make it clear that we don't restore original permissions (those are not known).

tools/tests/systemtests/sources.py
MakisH: I see the point. I think it is a high-effort/low-benefit, but since the effort is already invested and the potential impact isolated in this file, let's keep it.

However, I think we need a log message when fetching an external case or getting it from the cache, to avoid surprises.

Ideally, also automatically prepend an `External:` in the name of the case in the logs.

tools/tests/systemtests/sources.py
MakisH: I [tested it again](https://github.com/precice/tutorials/actions/runs/28646023071) and it works. Thanks!

tools/tests/systemtests/TestSuite.py
MakisH: I [tested it again](https://github.com/precice/tutorials/actions/runs/28646023071), and it works. Thanks!

tools/tests/systemtests/sources.py
PranjalManhgaye: Thanks, I have added fetch/cache log messages and the External: prefix in case names.

MakisH: [Works](https://github.com/precice/tutorials/actions/runs/28658287093/job/84992583813), but it seems like the order of the logs is a bit wrong now: the caching log message appears in the logs of the previous test:

Screenshot from 2026-07-03 13-54-04

MakisH: The output [now](https://github.com/precice/tutorials/actions/runs/28663440399/job/85011468788) appears only once per source, at the top:

Image

We should eventually more clearly format and annotate this initial output, but this is good enough for now.

tools/tests/README.md
```diff
@@ -170,6 +170,42 @@ Note that you will need to define the `TUTORIALS_REF` in the file [`reference_ve
...
+        case_combination:
+          - left-openfoam
+          - right-openfoam
+        reference_result: ./path-to-test-case/reference_results/participant-combination.tar.gz
+
+  release:
+    tutorials:
+      - *quickstart_openfoam_cpp
+    external:
+      - *some-external-test-case_left-openfoam_right-openfoam
```
```suggestion
  some-external-test-case:
    external:
      - &some-external-test-case
        source:
          type: git
          url: https://github.com/some-user/tutorials.git
          ref: some-branch
          subdir: .
        path: path-to-test-case
        case_combination:
          ...

  release:
    tutorials:
      - *quickstart_openfoam_cpp
    external:
      - *some-external-test-case
```

MakisH (approved):
Looks good now, thank you!

PR #735, opened 2026-03-02
System tests: Add configurable simulation timeout per test case

### LINKED ISSUE #371: Add Timeout to kill the simulation when it "hangs"
Proposed syntax:
```
  fenics_test:
    tutorials:
        case_combination:
 
         - fluid-openfoam
         - solid-fenics
        reference_result: ./flow-over-heated-plate/reference-data/fluid-openfoam_solid-fenics.tar.gz
        timeout: 400s
```

AdityaGupta716:
Adds an optional `timeout` field to entries in `tests.yaml`, allowing per-test timeout configuration instead of a single hardcoded global value.

Closes #371

## Changes

- `systemtests/TestSuite.py` — parses optional `timeout` from `tests.yaml`; validates that the value is positive
- `systemtests/Systemtest.py` — solver run and fieldcompare now use the per-test configured timeout; sets the default timeout to 900
- `tools/tests/README.md` — documented the new optional field with an example

## Example usage in `tests.yaml`

```yaml
- path: heat-exchanger
  case_combination:
    - fluid-openfoam
    - solid-calculix
  reference_result: ./heat-exchanger/reference-results/...
  timeout: 3600
```

Backward compatible — existing entries without `timeout` use the previous default of 1200.

## Checklist
- [x] I added a summary of any user-facing changes in `changelog-entries/735.md`
- [x] I will remember to squash-and-merge


AdityaGupta716:
Hi @MakisH, the timeout is currently attached to ReferenceResult since that's where per-test configuration lives. Happy to move it to a dedicated field or separate dataclass if you prefer a different structure. Also kept BUILD_TIMEOUT fixed since build time feels infrastructure-dependent rather than simulation-dependent — let me know if you'd like that configurable too.

tools/tests/metadata_parser/metdata.py
```diff
@@ -279,7 +279,6 @@ def from_cases_tuple(cls, cases: Tuple[Case], tutorial: Tutorial):
 class ReferenceResult:
     path: Path
     case_combination: CaseCombination
-
```
MakisH, line 282: Unrelated change.

tools/tests/systemtests/Systemtest.py
```diff
@@ -19,7 +19,7 @@
 import os
 
 
-GLOBAL_TIMEOUT = 900
+BUILD_TIMEOUT = 900
```
MakisH: Why rename it?
How does this fit with the changes suggested/discussed in #731?

Let's discuss directly in the Changelog entry thread.

tools/tests/components.yaml
MakisH, line 1: Unrelated changes, see https://github.com/precice/tutorials/pull/731#discussion_r2993357870 and revert.

tools/tests/README.md
```diff
@@ -331,8 +331,11 @@ test_suites:
           - fluid-openfoam
           - solid-fenics
         reference_result: ./flow-over-heated-plate/reference-results/fluid-openfoam_solid-fenics.tar.gz
+        timeout: 3600
 ```
 
+The optional `timeout` field (in seconds, default: 600) sets the maximum time allowed for the solver run and fieldcompare phases. Use this for long-running tutorials such as `heat-exchanger` or `turek-hron-fsi3`.
```
MakisH: To be able to test this PR, you could set a timeout to one tutorial. We can remove it again before merging. For example, the Nutils cases are much slower to run than the rest, while the new DuMux cases take the longest to build.

Note that the default recently changed to 900 in https://github.com/precice/tutorials/pull/696

changelog-entries/735.md
```diff
@@ -0,0 +1 @@
+- Add optional `timeout` field to `tests.yaml` entries, allowing per-test timeout configuration. Defaults to 600s for backward compatibility. Applies to the solver run and fieldcompare phases.
```
MakisH: - I would not mention the default here (it recently changed to 900s in #696)
- Is there a design reason for only applying to solver run and fieldcompare, or is it just complicated to apply it to the build phase as well?

The total timeout now per case would be the build timeout + the case-specific timeout = 900s + 600s = 1500s = 25min by default. Is that intended? That would be much larger than what we currently have.

tools/tests/systemtests/TestSuite.py
```diff
@@ -10,6 +10,7 @@ class TestSuite:
     name: str
     cases_of_tutorial: Dict[Tutorial, List[CaseCombination]]
     reference_results: Dict[Tutorial, List[ReferenceResult]]
+    timeouts: Dict[Tutorial, List[int]] = field(default_factory=dict)
```
MakisH: Why is this a (dictionary with a) list per `Tutorial` and not a single variable per `CaseCombination`?

tools/tests/systemtests/Systemtest.py
```diff
@@ -134,6 +134,7 @@ class Systemtest:
     arguments: SystemtestArguments
     case_combination: CaseCombination
     reference_result: ReferenceResult
+    timeout: int = 600
```
MakisH: This number seems to now be hard-coded in different places.

tools/tests/systemtests.py
```diff
@@ -58,10 +58,11 @@ def main():
         for test_suite in test_suites_to_execute:
             tutorials = test_suite.cases_of_tutorial.keys()
             for tutorial in tutorials:
-                for case, reference_result in zip(
-                        test_suite.cases_of_tutorial[tutorial], test_suite.reference_results[tutorial]):
+                timeouts = test_suite.timeouts.get(tutorial, [600] * len(test_suite.cases_of_tutorial[tutorial]))
```
MakisH: Why hard-coding the `600` here? Same for the `generate_reference_results.py`.

MakisH (commented):
Thanks for the PR! This is going in a good direction, but needs some further refactoring and cleanup. See some comments.

tools/tests/metadata_parser/metdata.py
AdityaGupta716: removed it

tools/tests/systemtests/Systemtest.py
AdityaGupta716: No longer renaming, GLOBAL_TIMEOUT stays as is, now overridable via PRECICE_SYSTEMTEST_TIMEOUT environment variable as suggested .

tools/tests/components.yaml
AdityaGupta716: removed it

tools/tests/README.md
AdityaGupta716: Added timeout: 1200 to nutils_test. Default now references to GLOBAL_TIMEOUT

changelog-entries/735.md
AdityaGupta716: Removed the default mention from changelog. The per-test timeout only applies to solver run and fieldcompare because build time depends on infrastructure , not on the simulation itself. Default behavior is unchanged all three phases default to GLOBAL_TIMEOUT . The per-test timeout only takes effect when set in tests.yaml.

tools/tests/systemtests/TestSuite.py
AdityaGupta716: i followed it from cases_of_tutorial and reference_result thought should do the same here should i change it ?

tools/tests/systemtests/Systemtest.py
AdityaGupta716: fixed it , now timeout: int = GLOBAL_TIMEOUT

tools/tests/systemtests.py
AdityaGupta716: fixed it

flow-over-heated-plate/reference-results/fluid-openfoam_solid-fenics.tar.gz
MakisH, line 1: I did not mean to update these files, not sure where the changes came from. I will revert.

changelog-entries/735.md
MakisH: I see that having a per-suite build timeout is a bit more complicated with the current code structure. Let's leave it at that, which would already help with running longer simulations.

MakisH:
~~@copilot apply changes based on the comments in [this thread](https://github.com/precice/tutorials/pull/735#pullrequestreview-4275775101)~~ (fails for reasons related to our Git LFS config)

MakisH (approved):
I tested this locally, and it works as expected.

PR #736, opened 2026-03-02
Add heat-exchanger-simplified to system tests (GSoC entry task)

PranjalManhgaye:
**GSoC 2026 entry task (System Tests Improvements)**

**New test**
- Added heat_exchanger_simplified_test in tests.yaml for the heat-exchanger-simplified tutorial with case (fluid-top-

- openfoam, fluid-btm-openfoam, solid-calculix) and reference path ./heat-exchanger-simplified/reference-results/fluid-top-openfoam_fluid-btm-openfoam_solid-calculix.tar.gz.
- Created heat-exchanger-simplified/reference-results/ with .gitkeep only; the actual .tar.gz can be generated later via CI or generate_reference_results.py.

**Compatibility fixes (so the new suite runs locally)**
- Systemtest.py: Use either docker compose or docker-compose and pass the compose file with -f so it works on both.
- 
- ubuntu_2404 Dockerfile: Handle Ubuntu 24.04 default user (UID 1000) by removing it and creating precice; set ARG defaults for PYTHON_BINDINGS_REF and FENICS_ADAPTER_REF; add wget retries for CalculiX source download.
- 
- components.yaml: Add PYTHON_BINDINGS_REF to calculix-adapter build args so the shared Dockerfile gets a valid ref when building the python_bindings stage.

**Local testing**
- Ran python3 systemtests.py --suites=heat_exchanger_simplified_test --build_args=TUTORIALS_REF:develop from tools/tests. Build and tutorial run succeed; it only fails at field comparison because the new reference archive is missing (expected).
- 
- Tried generate_reference_results.py but hit the 600s Docker build timeout on the test; reference for heat-exchanger-simplified can be generated in CI or by a maintainer.
- **What I ran**
- `python3 print_case_combinations.py` — heat-exchanger-simplified and the new case appear as expected.
- `python3 systemtests.py --suites=heat_exchanger_simplified_test --build_args=TUTORIALS_REF:develop` — Docker images build, the coupled run completes; the run fails only at the field-comparison step with `FileNotFoundError` for the missing reference archive (expected until the .tar.gz is generated).

**Checklist:**

### DIFF
--- a/changelog-entries/736.md
+++ b/changelog-entries/736.md
@@ -1,0 +1,1 @@
+- Added heat-exchanger-simplified to system tests (new suite `heat_exchanger_simplified_test`). Improved Docker and Ubuntu 24.04 compatibility for local system test runs (Compose detection, precice user, build args, CalculiX download).
--- a/tools/tests/components.yaml
+++ b/tools/tests/components.yaml
@@ -117,6 +117,9 @@ calculix-adapter:
     TUTORIALS_REF:
       description: Tutorial git reference to use
       default: "master"
+    PYTHON_BINDINGS_REF:
+      description: Git ref of the Python bindings (required when shared Dockerfile builds python_bindings stage)
+      default: "master"
     CALCULIX_VERSION:
       description: Version of Calculix to use
       default: "2.20"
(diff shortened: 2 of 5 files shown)


MakisH:
Duplicate of https://github.com/precice/tutorials/pull/731, which was first and has since been merged.

Note that this PR addresses many points in one patch, and its description is a bit long and difficult to review.

PR #737, opened 2026-03-02
Add pybind11 to dockerfile for system test

Fujikawas:
According to https://github.com/precice/tutorials/pull/696#issuecomment-3983020818, the `micro-manager` participant in `two-scale-heat-conduction` case depends on `pybind11`.

Checklist:


IshaanDesai (commented):
pybind11 is technically a dependency of micro-dumux, and not of the adapter. Could we install it right before the participant micro-dumux is compiled?

Fujikawas:
> pybind11 is technically a dependency of micro-dumux, and not of the adapter. Could we install it right before the participant micro-dumux is compiled?

You are right. As pybind11 is also not an independent component that we would test etc. like the other adapters/MM, I'd tend to merge it into the installation of `Micro Manager` since it's often used between C++ case and MM. Also I would like to base the dumux-adapter on preCICE instead of MM.

IshaanDesai:
> You are right. As pybind11 is also not an independent component that we would test etc. like the other adapters/MM, I'd tend to merge it into the installation of `Micro Manager` since it's often used between C++ case and MM. Also I would like to base the dumux-adapter on preCICE instead of MM.

Makes sense! Let's trigger the system test from this PR once the changes are done.

two-scale-heat-conduction/metadata.yaml
```diff
@@ -17,4 +17,6 @@ cases:
     participant: Micro-Manager
     directory: ./micro-dumux
     run: ./run.sh -l /home/precice/dumux
-    component: dumux-adapter
+    component:
+      - dumux-adapter
+      - micro-manager
```
MakisH: I don't think this is what you need, see https://github.com/precice/tutorials/issues/739#issuecomment-4025953148

MakisH (commented):
Besides https://github.com/precice/tutorials/issues/739#issuecomment-4025953148, the rest of the changes make sense.

IshaanDesai:
Despite changing the `DUMUX_ADAPTER_REF` to a commit hash on its develop branch, the system test manual run still fails with an error that occurs because an older version of the adapter is used. This version was previously used in the system test. Could it be that the cache is not cleared?

MakisH:
Link to the job output?

Fujikawas:
> Link to the job output?

https://github.com/precice/tutorials/actions/runs/23255829787

IshaanDesai:
> Link to the job output?

https://github.com/precice/tutorials/actions/runs/23255829787

The error that the participants `two-scale-heat-conduction/macro-dumux`, `free-flow-dumux`, and `porous-media-dumux` appear because the DuMuX adapter v3.0.0 gets picked. This happens despite https://github.com/precice/tutorials/pull/737/changes/b4d1d58adf450ed99ba844b1af3960be5c49c418, in which Jun changed the adapter reference to a commit on its develop branch. The question is why, despite doing this, the old version of the adapter got picked.

MakisH:
from the output:

```
WARNING: No argument provided for needed parameter DUMUX_VERSION. Substituting with 3.7
WARNING: No argument provided for needed parameter DUMUX_ADAPTER_REF. Substituting with main
WARNING: No argument provided for needed parameter DUNE_VERSION. Substituting with 2.9
```

are the changes already in `main`?

IshaanDesai:
Good catch! These changes are not in `main`. @Fujikawas, this means that the `DUMUX_ADAPTER_REF` is not set correctly.

MakisH:
I had a brief conversation on Wednesday with @Fujikawas and I think I said something wrong: the default values are not been picked up from the `reference_versions.yaml`, but from the `components.yaml`.

Fujikawas:
Test locally on my laptop with `--build_args TUTORIALS_REF:develop,MICRO_MANAGER_VERSION:develop,DUMUX_ADAPTER_REF:develop` the heat conduction case finished normally and missed reference results.

IshaanDesai:
> the heat conduction case finished normally and missed reference results.

which is expected or is there a problem?

tools/tests/dockerfiles/ubuntu_2404/Dockerfile
```diff
@@ -202,8 +202,15 @@ RUN python3 -m venv /home/precice/venv && \
...
     git clone https://github.com/precice/micro-manager.git && \
     cd micro-manager && \
+    if [ -n "${MICRO_MANAGER_PR}" ]; then git fetch origin pull/${MICRO_MANAGER_PR}/head; fi && \
     git checkout ${MICRO_MANAGER_VERSION} && \
     pip3 install .
+WORKDIR /home/precice
+RUN python3 -m venv /home/precice/venv && \
+    . /home/precice/venv/bin/activate && \
+    pip3 install pybind11
+ENV pybind11_DIR = "/home/precice/venv/lib/python3.13/site-packages/pybind11/share/cmake/pybind11"
```
MakisH: - Why is `pybind11_DIR` needed?
- The `python3.13` is bound to the specific Python version. Is there a way to make this future-proof?

MakisH (approved):
Looks fine overall, just one question.

tools/tests/dockerfiles/ubuntu_2404/Dockerfile
Fujikawas: This shouldn't be necessary anymore with the venv activated. I will remove it.

PR #738, opened 2026-03-02
feat(systemtests): add per-test max_time override for precice-config closes#402

PranjalManhgaye:
This PR adds support for an optional max_time field per tutorial entry in tools/tests/tests.yaml.
TestSuite parses max_time and passes it to each Systemtest instance as max_time.
Before running a test, Systemtest now optionally overrides <max-time value="..."/> in the copied precice-config.xml, if max_time is set and <max-time> exists.
No existing tutorials or reference results are modified; the PR only adds the override mechanism, so behavior is unchanged unless a max_time entry is added in tests.yaml.
This is complementary to PR #734, which adds a wall‑clock --timeout for the systemtest subprocess; this PR implements the config‑level restriction requested in #402.

Checklist:


PranjalManhgaye:
this version adds validation for max_time and a regex that only touches <max-time> (not other attributes) , thanks

MakisH:
Thank you for contributing! This was in a difficult to review state (mainly because it includes a few unrelated changes). https://github.com/precice/tutorials/pull/744 was solving the same issue and it was easier to review. After some iterations, it has now been merged, so I have to close this one as duplicate, even though it came first. But there was so much traffic during the GSoC application period, that I lost the overview a bit.

Issue #601, opened 2024-11-30
date:unrecognized option '--rfc-email'


LaysonShen:
Hey guys, I actually hate it that commit an issue whenever running into a problem. But this time it really drives me mad is a bit weird.
To be short, I tried to test Flow-around-controlled-moving-cylinder and ran
```
 ./run.sh
``` 
separately in three terminals. However bash raised an error immediately
```
date: Unrecognized option '--rfc-email', use '--help' for more information.
```
 It seems that this error was raised not from precice, but I don't know how to deal with it, so I commit this issue.
- Tutorials state (last commit / release):release 202404.0
- Versions of solvers and adapters used: Openfoam 10, openfoam-adapter-Openfoam10, Python3.13, numpy1.26.4, pyprecice3.1.2, fmiprecice, Fmpy0.3.21
- preCICE version:3.1.2
- Log files:Only two lines  as mentioned above.

LaysonShen:
root_directory/tools/log.sh has two lines using --rfc-email. We can change them into '--rfc-2822'.

PR #741, opened 2026-03-04
Fix macOS compatibility in tools/log.sh

Kunal241207:
## Summary 

tools/log.sh used GNU-specific flags that are not available on macOS.

This change makes the script portable by:

- Replacing `date --rfc-email` with a helper that falls back to a compatible `date` format when the flag is unavailable.
- Replacing `tee --append` with the portable `tee -a`.

The behavior of the script remains the same, but it now works correctly on macOS systems.

---
Checklist:

### DIFF
--- a/tools/log.sh
+++ b/tools/log.sh
@@ -5,14 +5,18 @@ CASENAME="$(pwd | xargs basename)"
 LOGFILE="$CASENAME.log"
 export LOGFILE
 
-STARTDATE="$(date --rfc-email)"
+rfc_email_date() {
+    date --rfc-email 2>/dev/null || date -u '+%a, %d %b %Y %H:%M:%S %z'
+}
+
+STARTDATE="$(rfc_email_date)"
 STARTTIME="$(date +%s)"
 echo "Started on: $STARTDATE" | tee "$CASENAME.log" 2>&1
 
 close_log() {
-    echo "Started on:  $STARTDATE" | tee --append "$LOGFILE" 2>&1
-    ENDDATE="$(date --rfc-email)"
+    echo "Started on:  $STARTDATE" | tee -a "$LOGFILE" 2>&1
+    ENDDATE="$(rfc_email_date)"
     ENDTIME="$(date +%s)"
-    echo "Finished on: $ENDDATE" | tee --append "$LOGFILE" 2>&1
-    echo "Duration:    $((ENDTIME-STARTTIME)) seconds (wall-clock time, including time waiting for participants)" | tee --append "$LOGFILE" 2>&1
+    echo "Finished on: $ENDDATE" | tee -a "$LOGFILE" 2>&1
+    echo "Duration:    $((ENDTIME-STARTTIME)) seconds (wall-clock time, including time waiting for participants)" | tee -a "$LOGFILE" 2>&1
 }
--- a/tools/log.sh
+++ b/tools/log.sh
@@ -6,7 +6,7 @@ LOGFILE="$CASENAME.log"
 export LOGFILE
 
 rfc_email_date() {
-    date --rfc-email 2>/dev/null || date -u '+%a, %d %b %Y %H:%M:%S %z'
+    date --rfc-email 2>/dev/null || date '+%a, %d %b %Y %H:%M:%S %z'
 }
 
 STARTDATE="$(rfc_email_date)"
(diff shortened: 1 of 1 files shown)


PranjalManhgaye:
@Kunal241207  nice fix. two quick questions: (1) Was using date -u in the fallback intentional? on linux we get local time, so i was wondering if we’d want the same on macOS , (may be removing that -u) (2) One “Started on:” has two spaces after the colon and the other has one—was that on purpose? Just curious. thanks

Kunal241207:
@PranjalManhgaye Good catch, thanks!

1) The `-u` wasn’t intentional. You're right that `date --rfc-email` uses local time, so removing `-u` makes the fallback match the original behavior better.
2) The spacing difference wasn’t intentional either. I’ll normalize it to a single space.

I've pushed the update.

MakisH:
Ah, this has been in my list since longer, thanks!

Related issue: https://github.com/precice/tutorials/issues/601

Do you think this approach solves both?

Kunal241207:
This change makes the timestamp generation in tools/log.sh portable, so it avoids the `--rfc-email` failure reported in #601 when this script is used.
However, I haven't done a full sweep of the repository to check for other direct usages of `date --rfc-email`, so there may still be additional occurrences elsewhere.

tools/log.sh
```diff
@@ -5,14 +5,18 @@ CASENAME="$(pwd | xargs basename)"
 LOGFILE="$CASENAME.log"
 export LOGFILE
 
-STARTDATE="$(date --rfc-email)"
+rfc_email_date() {
+    date --rfc-email 2>/dev/null || date '+%a, %d %b %Y %H:%M:%S %z'
```
MakisH, line 9: Why the `2>/dev/null`? If there is an error, I would rather see it than get the script to fail silently.

tools/log.sh
```diff
@@ -5,14 +5,18 @@ CASENAME="$(pwd | xargs basename)"
 LOGFILE="$CASENAME.log"
 export LOGFILE
 
-STARTDATE="$(date --rfc-email)"
+rfc_email_date() {
```
MakisH, line 8: The `rfc_email_date` directly refers to the implementation options (`--rfc-email`), which is the non-portable part. A more general name would be better.

tools/log.sh
MakisH, line 9: If `--rfc-email` is not portable, why provide an explicit format as fallback, instead of making the explicit format the only option? That should work on every system.

tools/log.sh
```diff
@@ -5,14 +5,18 @@ CASENAME="$(pwd | xargs basename)"
...
 
 close_log() {
-    echo "Started on:  $STARTDATE" | tee --append "$LOGFILE" 2>&1
-    ENDDATE="$(date --rfc-email)"
+    echo "Started on: $STARTDATE" | tee -a "$LOGFILE" 2>&1
+    ENDDATE="$(rfc_email_date)"
     ENDTIME="$(date +%s)"
-    echo "Finished on: $ENDDATE" | tee --append "$LOGFILE" 2>&1
-    echo "Duration:    $((ENDTIME-STARTTIME)) seconds (wall-clock time, including time waiting for participants)" | tee --append "$LOGFILE" 2>&1
+    echo "Finished on: $ENDDATE" | tee -a "$LOGFILE" 2>&1
```
MakisH, line 20: What's wrong with `--apend`? I used the verbose option for readability.

I don't have a macOS system to test on, but [this page](https://amazingalgorithms.com/commands/tee-macos/), for example, suggests that both are valid.

tools/log.sh
```diff
@@ -5,14 +5,18 @@ CASENAME="$(pwd | xargs basename)"
...
+    echo "Duration:    $((ENDTIME-STARTTIME)) seconds (wall-clock time, including time waiting for participants)" | tee -a "$LOGFILE" 2>&1
```
MakisH, line 21: Could you post some example output from this script, to see the format and the computed time?

MakisH (changes requested):
I just had another look. Overall, this PR not there yet, but it can be improved.

There are no other places where this is used.

MakisH:
Closing due to inactivity, and since this PR is more complicated than needed (see https://github.com/precice/tutorials/pull/741#discussion_r2975601510). Thanks for the suggestion anyway!

Kunal241207:
> Closing due to inactivity, and since this PR is more complicated than needed (see [#741 (comment)](https://github.com/precice/tutorials/pull/741#discussion_r2975601510)). Thanks for the suggestion anyway!

Sorry, I was not able to help in this, I was super stuck at my academics schedule. I'll improve my skills and will contribute again.

MakisH:
No worries! Happy to review any further PRs in the future! I am just trying to sort out the remaining PRs from the past few months.

PR #742, opened 2026-03-07
System tests: Archive fieldcompare diff files on failure 

### LINKED ISSUE #441: Archive diff files in the reports
We can pass the `--diff` file in fieldcompare to compute differences between meshes: https://gitlab.com/dglaeser/fieldcompare/-/merge_requests/197

We could archive such a VTK file, and we could potentially also archive common visualizations that would give us an overview of where things are going wrong, or how much.

Related to #426

AdityaGupta716:
## What this PR does

When fieldcompare fails, the diff VTK files it generates (via the `--diff` flag already present in the compose template) are now copied to a `diff-results/` folder inside the system test directory. This means they get included in the CI artifact and can be downloaded for visual inspection of exactly where results diverged.

Closes #441

## Changes

- Added `__archive_diff_files()` to `Systemtest.py` — scans `precice-exports/` for any `*diff*.vtu`, `*diff*.vtk`, or `*diff*.vtp` files after a fieldcompare failure and copies them to `diff-results/`
- Called only when `fieldcompare_result.exit_code != 0` — no unnecessary copies on passing runs
- Updated `README.md` to mention that diff files are available in `diff-results/` for ParaView inspection

## Notes

The `--diff` flag is already passed to fieldcompare in `docker-compose.field_compare.template.yaml`, so the diff files are already being generated — they just weren't being saved anywhere accessible until now.

## Checklist
- [x] I added a summary of any user-facing changes in `changelog-entries/742.md`


AdityaGupta716:
hi @MakisH, plz review!  The `--diff` flag was already being passed to fieldcompare but the generated diff files were never saved anywhere accessible — this copies them to `diff-results/` on failure so they show up in the CI artifact and can be opened in ParaView to see exactly where results diverged.

MakisH:
This PR seems to be building upon your previous one https://github.com/precice/tutorials/pull/744, making it difficult to merge independently.

It is also very similar to https://github.com/precice/tutorials/pull/740, which was opened first. Why this one as well?

AdityaGupta716:
@MakisH This one searches recursively for diff files in subdirectories, matches all output formats fieldcompare produces not just VTK, has the safety check to avoid copy loops, and adds logging for the really common case where fieldcompare fails but doesn't end up writing any diff files at all.
yes the timeout and yaml changes don't belong here. I'll remove the unrelated stuff out of this PR right now if you'd prefer to go with this version.

MakisH:
Closing as duplicate of https://github.com/precice/tutorials/pull/740, which was opened first and was very similar.

Thanks for contributing, in any case!

PR #743, opened 2026-03-08
Archive and compare iterations.log for implicit coupling

### LINKED ISSUE #440: Compare iterations.log

PranjalManhgaye:
## Summary
Implicit coupling tutorials write precice-*-iterations.log files that show how many coupling iterations each time step needed. This PR adds system-test support to save those logs and check them against reference copies, so we notice when coupling behaviour changes unexpectedly. Closes #440.

## Changes

- **Systemtest.py**: After a tutorial run, collect `precice-*-iterations.log` files into `iterations-logs/` and compare them by SHA-256 hash against archived reference copies in a `*.iterations-logs/` directory next to each reference `.tar.gz`. A successful check is logged at INFO; a mismatch fails the test.
- **generate_reference_results.py**: When you regenerate reference results, the iterations logs are copied into that *.iterations-logs/ folder so future runs have something to compare against.
- **README.md**: Documents the behaviour.
- **changelog-entries/743.md**: Changelog entry for this PR.

## test (local)

image

## Notes

- Tests without reference iterations logs skip comparison (unchanged tutorials).
- Explicit coupling (no iterations logs) is unaffected.

## Checklist

- [x] Changelog entry in `changelog-entries/743.md`


MakisH:
@PranjalManhgaye could you please resolve the merge conflicts and prepare this PR for reviewing?

tools/tests/generate_reference_results.py
```diff
@@ -139,6 +140,16 @@ def main():
             raise RuntimeError(
                 f"Error executing: \n {systemtest} \n Could not find result folder {reference_result_folder}\n Probably the tutorial did not run through properly. Please check corresponding logs")
 
+        # Write iterations.log hashes sidecar for implicit-coupling regression checks (issue #440)
```
MakisH: Instead of archiving the hashes, why not archive the files themselves, similarly to the export files?

Computing the hashes is quick, and the files are small.

Having the complete reference files would be more useful to say whether the regression is significant or not.

tools/tests/systemtests/Systemtest.py
```diff
@@ -413,6 +417,88 @@ def _run_field_compare(self):
...
+            for log_file in case_dir.glob("precice-*-iterations.log"):
+                if log_file.is_file():
+                    rel = f"{Path(case.path).name}/{log_file.name}"
+                    collected.append((rel, log_file))
+        return collected
+
+    def __archive_iterations_logs(self):
+        """
+        Copy precice-*-iterations.log from case dirs into iterations-logs/
+        so they are available in CI artifacts (issue #440).
```
MakisH: Referring to (solved) issues in code is a bit strange.

changelog-entries/440.md
```diff
@@ -0,0 +1 @@
+- Archive `precice-*-iterations.log` files into `iterations-logs/` and compare them by SHA-256 hash against reference data for implicit-coupling regression checks; reference hashes are stored in `.iterations-hashes.json` sidecar files (fixes [#440](https://github.com/precice/tutorials/issues/440)).
```
MakisH: FYI, changelogs are typically written in past tense (in contrast to commit messages, which are typically written in imperative/present).

In this repository, the changelog entries are expected to be named after the PR, not after the issue they solve. They should also link to the respective PR.

MakisH (changes requested):
Some first comments. I have not yet looked at the complete PR or tested it.

PranjalManhgaye:
thanks @makish for your review , i will make this pr review ready and also resolve conflicts .

tools/tests/generate_reference_results.py
PranjalManhgaye: yeah seems correct , having the actual reference files is more useful than only storing hashes , therefore i  changed it so generate_reference_results copies the precice-*-iterations.log files into a *.iterations-logs/ folder next to each reference .tar.gz  (quite similar idea to the export archives). The test runner still compares by SHA-256,  but it reads the expected hashes from those archived files. Older .iterations-hashes.json sidecars still work as a fallback  if they exist.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: removed the issue references from the docstrings and kept them about what the code actually does

changelog-entries/440.md
PranjalManhgaye: my mistake sorry , thanks I switched to changelog-entries/743.md, wrote it in past tense, and linked to this PR instead of naming it after the issue.

PranjalManhgaye:
Hi @MakisH  , rebased onto develop, updated the PR description, and replied on your review threads. ready for another look when you have time.

tools/tests/systemtests/Systemtest.py
```diff
@@ -491,6 +494,119 @@ def __archive_fieldcompare_diffs(self) -> None:
...
+        if ref_dir.is_dir():
+            ref_hashes = {}
+            for log_file in ref_dir.rglob("precice-*-iterations.log"):
+                if log_file.is_file():
+                    rel = log_file.relative_to(ref_dir).as_posix()
+                    ref_hashes[rel] = self._sha256_file(log_file)
+            if ref_hashes:
+                return ref_hashes
+
+        sidecar = self.reference_result.path.with_suffix(".iterations-hashes.json")
```
MakisH: Such files should not exist, since they were added in a previous version of this PR. Removing this feature should also remove the dependency on json.

MakisH (commented):
I have not yet completed my review, but here is already a first comment on something to check.

Tests I have executed locally:

- `python3 systemtests.py --suites="openfoam_adapter_release"` (no iterations in the reference results) succeeds
- `python3 generate_reference_results.py --suites="openfoam_adapter_release"` worked and generated results in an (uncompressed) folder `flow-over-heated-plate/reference-results/fluid-openfoam_solid-openfoam.iterations-logs` with `fluid-openfoam/precice-Fluid-iterations.log` and `solid-openfoam/precice-Solid-iterations.log`.
- after updating the reference results: `python3 systemtests.py --suites="openfoam_adapter_release"` worked, but I don't see `sha256` or `iterations` in the output to confirm.

tools/tests/systemtests/Systemtest.py
```diff
@@ -491,6 +494,119 @@ def __archive_fieldcompare_diffs(self) -> None:
...
+                dest_name = prefix + dest_name
+            shutil.copy2(src, dest_dir / dest_name)
+        logging.debug(
+            "Archived %d iterations log(s) to %s for %s",
+            len(collected),
+            dest_dir,
+            self,
+        )
+
+    def __compare_iterations_hashes(self) -> bool:
```
MakisH, line 795: Is there output for successful checks? Where?

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: that was leftover from an earlier approach. I removed the .iterations-hashes.json path entirely in 095502d, along with import json. Expected hashes now come only from the archived logs under *.iterations-logs/.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: in 095502d, __compare_iterations_hashes() now logs at INFO when it passes

PranjalManhgaye:
Hi @MakisH , pushed 095502d ; removed the JSON sidecar fallback and added INFO output when the iterations hash check passes.
Re-tested openfoam_adapter_release locally => same steps as you, and the second run now shows the INFO line.readme updated ,
Ready for another look when you can.

tools/tests/README.md
MakisH, line 1: When merging `develop` into this branch, you will find a new section on "Numerical regressions": https://precice.org/dev-docs-system-tests.html#numerical-regressions

We should list these files there as well.

changelog-entries/743.md
```diff
@@ -0,0 +1 @@
+- Archived `precice-*-iterations.log` files into `iterations-logs/` during system tests and compared them against reference copies for implicit-coupling regression checks ([#743](https://github.com/precice/tutorials/pull/743)).
```
MakisH:
```suggestion
- Added a feature to archive `precice-*-iterations.log` files into `iterations-logs/` during system tests and to compare them against reference copies for implicit-coupling regression checks ([#743](https://github.com/precice/tutorials/pull/743)).
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -664,6 +774,21 @@ def run(self, run_directory: Path):
...
+            )
+            return SystemtestResult(
+                False,
+                std_out,
+                std_err,
+                self,
+                build_time=docker_build_result.runtime,
+                solver_time=docker_run_result.runtime,
+                fieldcompare_time=0)
+
```
MakisH, line 983: Since the iterations comparison is stricter than the fieldcompare step (and it might be a false alarm), I think we should first run fieldcompare and then compare the iterations. I like that you first do the cheap comparison and only then do the expensive comparison, but I think in practice it will be a bit confusing.

We also need a way to communicate per-test that this has been checked and passed. By appending the output to the end of `system-tests-compare.log`, it will be picked up in the report if that fails, without any further changes.

MakisH (commented):
Here are some first comments. Even though I would hope for a simpler/shorter solution, this is also fine, as it is rather clear.

You could merge back from `develop` now, I am through with the larger changes I wanted to do for now. And I think this would then be rather easy to merge and it would make sense to do it soon.

tools/tests/README.md
PranjalManhgaye: yeah done

changelog-entries/743.md
PranjalManhgaye: done thanks

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: reordered so fieldcompare runs first ,and iterations check output is appended to system-tests-compare.log (pass, fail, or skip) , thanks for the suggestion ,,

PranjalManhgaye:
@MakisH i have resolved your initial comments and also merged develop , it is ready for another review thanks ,,

tools/tests/README.md
```diff
@@ -6,136 +6,157 @@ keywords: pages, development, tests
...
+Each tutorial contains automation scripts (mainly `run.sh` and `clean.sh`), as well as metadata (`metadata.yaml`). The metadata file describes the available cases, how to run them, as well as their dependencies. A central `tests.yaml` file defines test suites, which execute different combinations of cases. The Python script `systemtests.py` executes the tests, filter for specific test suites.
+
+Read more about the system tests in the publication [System Regression Tests for the preCICE Coupling Ecosystem](https://doi.org/10.14279/eceasst.v83.2614).
+
+[![ECEASST](https://img.shields.io/badge/DOI-10.14279%2Feceasst.v83.2614-green)](https://doi.org/10.14279/eceasst.v83.2614)
 
-Let's dive deeper into some of these aspects.
+<details markdown="1"><summary>Expand the implementation details...</summary>
```
MakisH, line 158: Something went wrong in the conflict resolution here.

tools/tests/README.md
```diff
@@ -6,136 +6,157 @@ keywords: pages, development, tests
...
+
+### Numerical regressions
+
+When the tests fail at the results comparison step, this typically means that there are numerical regressions (unless something wrong went undetected in a previous step). We use [fieldcompare](https://gitlab.com/dglaeser/fieldcompare) to compare all [preCICE exports](https://precice.org/configuration-export.html) to reference results generated from a previous run. Relevant files:
+
+- `precice-exports/`: The coupling meshes of the test run.
+- `reference-results/`: The coupling meshes of the reference run, as stored on Git LFS, expanded into `reference-results-unpacked`.
+- `diff-results/`: Numerical difference of the results in the two directories (computed with `fieldcompare dir --diff precice-exports/ reference/`). These are only present on failed comparisons.
+- `iterations-logs/`: For implicit-coupling runs, archived `precice-*-iterations.log` files from the test run, compared by SHA-256 hash against reference copies.
+- `*.iterations-logs/`: Reference iterations logs stored next to each reference `.tar.gz` on Git LFS (e.g. `fluid-openfoam_solid-openfoam.iterations-logs/`).
```
MakisH: These lines can be simplified a bit (but I don't currently have the option to directly suggest changes):

```
- `iterations-logs/`: The `precice-*-iterations.log` files of the test run. Only present in implicit coupling tests and compared by SHA-256 hash against reference copies.
- `*.iterations-logs/`: The reference iterations log files, stored next to each reference results archive on Git LFS (e.g. `fluid-openfoam_solid-openfoam.iterations-logs/`).
```

MakisH (commented):
A test run to see what happens for the current cases: https://github.com/precice/tutorials/actions/runs/28050285585/job/83038819578

Other than that, the documentation needs another look.

tools/tests/README.md
PranjalManhgaye: checked against develop , that section matches ; only the two iterations bullets i added.

tools/tests/README.md
PranjalManhgaye: fine i have done it as per your wording

tools/tests/README.md
MakisH: Ah, indeed, the diff was a bit confusing.

MakisH:
Testing locally partially works, but it seems like the reference files are not added to the branch.

## Testing on GHA

<details><summary>(not that important for now, read first the local testing)</summary>

A [run without the iterations in the reference results](https://github.com/precice/tutorials/actions/runs/28050285585/job/83038819578) shows that the check is correctly skipped when iterations logs are not in the reference data.

A [run adding reference results](https://github.com/precice/tutorials/actions/runs/28086775915) logs that the iterations logs are archived in `reference-results/fluid-cpp_solid-cpp.iterations-logs`, and I see those in the runner.

A [subsequent run still reports that iterations files are not found](https://github.com/precice/tutorials/actions/runs/28087099368).
</details>

## Testing locally

Testing locally, again with the elastic-tube-1d fluid-cpp/solid-cpp, I do see the the directory:

```
..../tutorials/elastic-tube-1d/reference-results [test-743]$ tree
.
├── fluid-cpp_solid-cpp.iterations-logs
│   ├── fluid-cpp
│   │   └── precice-Fluid-iterations.log
│   └── solid-cpp
│       └── precice-Solid-iterations.log
├── ...
└── reference_results.metadata
```

which is excluded by the `.gitignore` rules.

Why not place this directory directly in the respective `fluid-cpp_solid-cpp.tar.gz`? It could be a separate directory inside it than the one with the `vtu` files, in case that helps.

PranjalManhgaye:
thanks for checking @MakisH , its correct if *.iterations-logs/ dirs are gitignored by *.log , so i will do it as you suggested and update the comparison logic accordingly

elastic-tube-1d/reference-results/fluid-cpp_solid-cpp.tar.gz
MakisH, line 1: The `reference-results` changes should not be added here. They will be generated on the runner.

MakisH:
Now the files are in the reference result archives, but `fieldcompare` fails:

```
field-compare-1  | [ FAILED  ] 2 file comparisons, listed below:
field-compare-1  | [ FAILED  ] iterations-logs/fluid-python/precice-Fluid-iterations.log: (Missing source file)
field-compare-1  | [ FAILED  ] iterations-logs/solid-python/precice-Solid-iterations.log: (Missing source file)
```

GHA run: https://github.com/precice/tutorials/actions/runs/28097293364

I think that having two separate directories in the archive would help. Ignoring specifically the `iterations-logs/` directory in `fieldcompare` would also be fine, as it would be moving the `iterations-logs/` outside the directory before running fieldcompare.

PranjalManhgaye:
VTUs and iteration logs are now separate top-level tar members ({case}/ and {case}.iterations-logs/) ,so fieldcompare no longer hits log files in the VTU reference dir , verified locally , fieldcompare and hash check both pass ,refs will be regenerated on GHA ,,

tools/tests/systemtests/Systemtest.py
```diff
@@ -27,6 +27,12 @@
 DIFF_RESULTS_DIR = "diff-results"
 ITERATIONS_LOGS_DIR = "iterations-logs"
 
+
+def reference_iterations_logs_tar_arcname(stem: str) -> str:
+    """Top-level tar member name for reference iterations logs (separate from VTU data)."""
+    return f"{stem}.{ITERATIONS_LOGS_DIR}"
```
MakisH: Similar to previous reviews: why make this rather trivial operation a standalone function?

Note: I would not call this "VTU data", but "preCICE exports". VTU is a general file format which is also used by some solvers.

PranjalManhgaye: i added the helper so both Systemtest.py and generate_reference_results.py would share the same {stem}.iterations-logs arcname , fair point that it's trivial , now inlined it in both places and removed the function ,,

tools/tests/systemtests/Systemtest.py
```diff
@@ -709,6 +711,137 @@ def __archive_fieldcompare_diffs(self) -> None:
...
+                logging.critical(message)
+                self._append_compare_log(message, error=True)
+                return False
+        if len(current) != len(ref_hashes):
+            extra = set(current) - set(ref_hashes)
+            message = f"Unexpected iterations log(s) {extra}; {self} fails"
+            logging.critical(message)
+            self._append_compare_log(message, error=True)
+            return False
+        self._append_compare_log("=== iterations.log hash check ===")
```
MakisH:
```suggestion
        self._append_compare_log("=== Comparing the iterations.log files (checksums only) ===")
```

tools/tests/README.md
```diff
@@ -94,8 +94,9 @@ In addition, in the directories of the cases executed, you can find `system-test
 When the tests fail at the results comparison step, this typically means that there are numerical regressions (unless something wrong went undetected in a previous step). We use [fieldcompare](https://gitlab.com/dglaeser/fieldcompare) to compare all [preCICE exports](https://precice.org/configuration-export.html) to reference results generated from a previous run. Relevant files:
 
 - `precice-exports/`: The coupling meshes of the test run.
-- `reference-results/`: The coupling meshes of the reference run, as stored on Git LFS, expanded into `reference-results-unpacked`.
+- `reference-results/`: The coupling meshes of the reference run, as stored on Git LFS, expanded into `reference-results-unpacked`. For implicit-coupling tests, the reference `.tar.gz` also contains a separate `{case}.iterations-logs/` directory with the reference `precice-*-iterations.log` files.
```
```suggestion
- `reference-results/`: The coupling meshes of the reference run, as stored on Git LFS, expanded into `reference-results-unpacked`. For test cases using implicit coupling, the reference `.tar.gz` also contains the reference `precice-*-iterations.log` files.
```

tools/tests/README.md
```diff
@@ -94,8 +94,9 @@ In addition, in the directories of the cases executed, you can find `system-test
...
 - `diff-results/`: Numerical difference of the results in the two directories (computed with `fieldcompare dir --diff precice-exports/ reference/`). These are only present on failed comparisons.
+- `iterations-logs/`: The `precice-*-iterations.log` files of the test run. Only present in implicit coupling tests and compared by SHA-256 hash against the copies unpacked from `{case}.iterations-logs/` in the reference `.tar.gz`.
```
```suggestion
- `iterations-logs/`: The `precice-*-iterations.log` files of the test run. Only present in test cases using implicit coupling. The comparisons to references only take into account the file SHA-256 checksums.
```

MakisH (approved):
Looks good now, and the tests I ran on GHA do what I would expect. Thanks for the updates!

General note: I guess some things are easier to implement inside these Python scripts, but it might also be a bit confusing that `fieldcompare` runs inside a container but the iterations comparison does not. In any case, it does not really matter here, but it is a bit strange structure-wise.

PR #744, opened 2026-03-08
Add per-test max_time override for system tests

### LINKED ISSUE #402: System tests: Restrict preCICE time for selected tests

AdityaGupta716:
**What this PR does**

Adds an optional `max_time` field to entries in `tests.yaml` to cap preCICE simulation time per test case.

**How it works:**
- Reads a `max_time` value from `tests.yaml`
- Overwrites the `<max-time>` (or `<max-time-windows>`) value in `precice-config.xml` while preparing the case to run
- Applies consistently to both test runs and reference result generation

Closes #402

**Checklist**
- [x] I added a summary of any user-facing changes in `changelog-entries/402.md`


AdityaGupta716:
@MakisH Plz review

MakisH:
How does this PR relate to the previously opened https://github.com/precice/tutorials/pull/738?

AdityaGupta716:
@MakisH  yes PR tackles the same issue as #738. I opened it separately because I wanted to keep the scope focused  just the max_time override, without the iterations-log changes that #738 also includes.
There are a couple of small differences in approach too: this one fails fast on invalid max_time values instead of silently skipping, and adds a README example.

changelog-entries/402.md
```diff
@@ -0,0 +1 @@
+- Add per-test `max_time` override in `tests.yaml` to cap preCICE simulation time without editing `precice-config.xml` manually. Applies consistently to both test runs and reference result generation. Handles multiple `<max-time>` tags with a warning.
```
MakisH: > Handles multiple `<max-time>` tags with a warning.

This should be a warning in preCICE library itself, not in the system tests.

tools/tests/systemtests/Systemtest.py
```diff
@@ -19,7 +19,7 @@
 import os
 
 
-GLOBAL_TIMEOUT = 900
+BUILD_TIMEOUT = 900
```
MakisH: Unrelated change, comes from #731 / #735. Please revert, to keep the PRs independently mergeable.

tools/tests/systemtests/Systemtest.py
```diff
@@ -134,6 +134,7 @@ class Systemtest:
     arguments: SystemtestArguments
     case_combination: CaseCombination
     reference_result: ReferenceResult
+    max_time: Optional[float] = None
```
MakisH: `Optionl` seems to be an old practice: https://stackoverflow.com/a/51710151/2254346

But I am not very up-to-date in Python.

tools/tests/systemtests/Systemtest.py
```diff
@@ -394,7 +395,7 @@ def _run_field_compare(self):
                                        cwd=self.system_test_dir)
 
             try:
-                stdout, stderr = process.communicate(timeout=GLOBAL_TIMEOUT)
+                stdout, stderr = process.communicate(timeout=self.timeout)
```
MakisH: Originates from #735, please separate the two PRs.

tools/tests/systemtests/Systemtest.py
```diff
@@ -513,11 +514,52 @@ def __write_logs(self, stdout_data: List[str], stderr_data: List[str]):
...
+        if not (isinstance(self.max_time, (int, float)) and self.max_time > 0):
+            logging.warning(
+                f"Invalid max_time {self.max_time!r} for {self}; must be a positive number. Skipping override.")
+            return
+        config_path = self.system_test_dir / "precice-config.xml"
+        if not config_path.exists():
+            logging.warning(
+                f"Requested max_time override for {self}, but no precice-config.xml "
+                f"found in {self.system_test_dir}")
+            return
```
MakisH: This should never happen in the system tests: all tests have a `precice-config.xml`, by definition. I would remove it to keep the code simple.

tools/tests/systemtests/Systemtest.py
```diff
@@ -513,11 +514,52 @@ def __write_logs(self, stdout_data: List[str], stderr_data: List[str]):
...
+        if len(matches) > 1:
+            logging.warning(
+                f"Multiple <max-time> tags found in {config_path}; overriding all to {self.max_time}")
+        new_text = re.sub(pattern, rf"\g<1>{self.max_time}\g<3>", text)
+        try:
+            config_path.write_text(new_text)
+            logging.info(f"Overwrote max-time in {config_path} to {self.max_time} for {self}")
+        except Exception as e:
+            logging.warning(f"Failed to write updated {config_path}: {e}")
+
```
MakisH: Note that `max-time` is not the only way to restrict the time. There is also `max-time-windows`.

https://precice.org/configuration-xml-reference.html#max-time

https://precice.org/configuration-xml-reference.html#max-time-windows

I would leave it up to the user to set the right thing, and not add so much error handling. I think this function is a bit more complicated than it needs to be.

Note that an alternative approach would be to extend the docker compose service before running. We already set some of these in the template, e.g., to always export VTK files:

https://github.com/precice/tutorials/blob/develop/tools/tests/docker-compose.template.yaml

tools/tests/systemtests/TestSuite.py
```diff
@@ -10,6 +10,7 @@ class TestSuite:
     name: str
     cases_of_tutorial: Dict[Tutorial, List[CaseCombination]]
     reference_results: Dict[Tutorial, List[ReferenceResult]]
+    max_times: Dict[Tutorial, List[Optional[float]]] = field(default_factory=dict)
```
MakisH: Similarly to https://github.com/precice/tutorials/pull/735: Why is this a (dictionary of a) list per tutorial, and not a value per `CaseCombination`?

tools/tests/components.yaml
MakisH, line 1: Unrelated, see #731 and #735.

MakisH (commented):
FYI, this PR (and some other) could benefit from:

- A short explanation of the main idea:
   - Reads a `max-time` value from `tests.yaml`
   - Overwrites the value in the `precice-config.xml` while preparing the case to run.
- Some Markdown formatting, especially where you show code.

Given my previous comment on the relation to another PR (https://github.com/precice/tutorials/pull/744#issuecomment-4112266021), note that I still need to review the other PR as well. Maybe don't rush implementing everything.

tools/tests/systemtests/Systemtest.py
MakisH: Actually, for shorter tests, it would be convenient to directly define `max-time-windows`, instead of `max-time`, because then we don't need to look at the `time-window-size` to decide on the value.

As a maintainer, I would be interested to know that, e.g., three coupling time windows complete, since most problems appear there. The time is more relevant for the application, not for the coupling.

changelog-entries/402.md
AdityaGupta716: removed it

tools/tests/systemtests/Systemtest.py
AdityaGupta716: Removed it

tools/tests/systemtests/Systemtest.py
AdityaGupta716: Changed to float | None

tools/tests/systemtests/Systemtest.py
AdityaGupta716: removed it

tools/tests/systemtests/Systemtest.py
AdityaGupta716: removed it

tools/tests/systemtests/Systemtest.py
AdityaGupta716: Done, no try/except, no file existence check. Added separate max_time_windows field so users can directly set the number of coupling windows without needing to calculate simulation time. Both fields are now supported.

tools/tests/systemtests/TestSuite.py
AdityaGupta716: followed the same pattern as cases_of_tutorial and reference_results

tools/tests/components.yaml
AdityaGupta716: removed it

MakisH (approved):
I tested this locally for `max_time`, and it works. Everything else looks good now.

Note that fieldcompare fails due to missing source files. I tried adding `--ignore-missing-source-files`, but then it still fails in the last time window, since the behavior is a bit different (probably a bug in the OpenFOAM adapter) and gave an error of just above the tolerance. This means that we cannot reuse the same reference results without some handling, but it is still useful for more complex cases.

Thanks again for contributing!

PR #748, opened 2026-03-08
systemtests: fix Python bindings version + robustness (fixes #584)

### LINKED ISSUE #584: Propagating Python bindings version in system tests (conflict with requirements.txt)

PranjalManhgaye:
**Description**

**Problem**
Tutorial run.sh scripts create their own .venv and install from requirements.txt (e.g. pyprecice~=3.0), silently overwriting the specific PYTHON_BINDINGS_REF version installed by the Docker image. This means the version argument passed to system tests was completely ignored.

**Root cause**
The Docker image installs the correct version into /home/precice/venv but never permanently activates it. At container runtime the venv isn't on PATH, so the run script sees no active venv and creates its own. PRECICE_TUTORIALS_NO_VENV (added in #680) was never set in the component templates, so the escape hatch was never triggered.

**Fix**
- Add ENV VIRTUAL_ENV and ENV PATH after venv creation in python_bindings, fenics_adapter, and nutils_adapter Dockerfile stages — permanently activating the correct venv at container runtime
- Set PRECICE_TUTORIALS_NO_VENV=1 in python-bindings, fenics-adapter, and nutils-adapter component templates — so run scripts skip creating their own venv and use the Docker image's version
- Add sanity check: before running, verify precice is importable when PRECICE_TUTORIALS_NO_VENV is set; fail fast with a clear error instead of a confusing failure later
- Changelog: note that ubuntu_2204 uses pip install --user (no venv) and is fixed by the template change alone

Fixes #584


MakisH:
Same comment as in https://github.com/precice/tutorials/pull/755#issuecomment-4112226709, reverse order:

This PR is almost identical to https://github.com/precice/tutorials/pull/745 (which was opened first), without referring to it. This is very bad practice (if copied without reference: plagiarism, if both using the same GenAI tool without disclosure: futuristic plagiarism).

I do not mind neither collaboration nor modern tools, but please disclose them. Copying from each other without attribution is not accepted.

PranjalManhgaye:
thanks @MakisH  for pointing this out, and sorry for the confusion,
I didn’t intend to duplicate or copy without reference. i should have linked and aligned this PR with #745 — that’s my mistake. i will update this by referencing the related PR and aligning the approach accordingly.
also, i will make sure to be clearer about any tools or collaboration involved going forward.

PR #749, opened 2026-03-08
Add requirements-reference.txt for reproducible Python deps

### LINKED ISSUE #610: Fix versions of Python dependencies for each distribution
Triggered by https://github.com/precice/tutorials/pull/609, as well as by #584 and related issues, I think we need a mechanism to make the versions of our Python dependencies very specific every time we release a new distribution, for reproducibility reasons.

This could be included in a release pull request template here, and maybe there is some Action or similar mechanism to check the version restrictions in `requirements.txt` files.

However, if we do this by changing the actual files here, we would need to merge back-and-forth, which would be a bit ugly.

An alternative would be something like a `Gemfile.lock` approach, where we would specify a set of versions that are known to work in a separate file (for example `requirements-reference.txt`).

PranjalManhgaye:
**Description**

**Problem**
For each distribution release we need reproducible Python dependency versions, but pinning in every tutorial `requirements.txt` would require a lot of merge back-and-forth. There is also no explicit list of versions known to work for a given release.

**Approach**
Introduce a separate `requirements-reference.txt` (lockfile-style) with pinned versions, without changing the existing tutorial `requirements.txt` or run scripts. The reference file is only a manifest of versions that are known to work for a distribution.

**What changed**
- Add `tools/tests/requirements-reference.txt` with pinned versions (pyprecice, numpy, matplotlib, nutils, setuptools)
- Add `update_requirements_reference.py` to regenerate it from `reference_versions.yaml` (pyprecice from PYTHON_BINDINGS_REF)
- Add `validate_requirements_reference.py` to ensure the pyprecice version matches `reference_versions.yaml`
- Add GitHub Action `check-requirements-reference.yml` to run this validation on PR/push
- Document the mechanism in `tools/tests/README.md` and add a reminder to the release PR template

**How to update at release**
- Preferred: `pip freeze` from the systemtest Docker image, then `pip freeze | python3 update_requirements_reference.py --from-freeze`
- Or: `python3 update_requirements_reference.py` (uses `reference_versions.yaml` for pyprecice and script defaults for other packages)

Fixes #610

---

Checklist:


MakisH (changes requested):
Thank you for contributing. I think the idea of having a script to generate `requirements-reference.txt` is indeed one possible solution to #610. However, I don't see the connection to `reference_versions.yaml`, which is about other dependencies than the `requirements.txt`.

Note that #610 mainly concerns the various Python-based tutorials that have `requirements.txt`. The system tests scripts are not that interesting for reproducibility: we mainly need the latest version to work. But when someone needs a older version of a Python-based tutorial, it is important that they can reproduce the original dependencies.

I think we need such a script to report the installed `requirements.txt` at a given time, on demand, plus some more version restrictions in the each `requirements.txt`. These version restrictions should state the requirements, without enforcing specific versions, so that the latest supported can generally be installed.

See how many tutorials require `numpy >1 <2`, or `precice~=3.0`.

Note that we should be able to get the latest available packages on PyPI, not just the ones that PIP might have cached on the local system.

Would you still like to continue working on this PR, or would you prefer to start a new one?

PranjalManhgaye:
@MakisH for now i am contnuing on this pr only and work with your given direction , 
 i am reworking it to focus on tutorial requirements.txt (looser constraints) plus an on-demand tools/report_tutorial_requirements.py that resolves versions from PyPI into root-level requirements-reference.txt , without tying it to reference_versions.yaml ,,

PranjalManhgaye:
Closing in favor of a new PR with the reworked implementation (rebased onto latest develop, tutorial-centric approach per review) ,,

PR #751, opened 2026-03-09
Make clean.sh scripts executable from any directory

k4rtikx:
Fix clean.sh scripts so they can be executed from any directory.

Many case-level clean.sh scripts source ../../tools/cleaning-tools.sh using
relative paths. When invoked from outside their directory (e.g.
./fluid-openfoam/clean.sh), the working directory differs and the relative
path breaks.

This PR ensures scripts always execute from their own directory by adding:

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
cd "$SCRIPT_DIR"

Changes:
- Added SCRIPT_DIR handling to 111 case-level clean.sh scripts
- Updated partitioned-heat-conduction-direct/clean-tutorial.sh
  (the only non-symlink clean-tutorial.sh)
- Added missing `set -e -u` to
  flow-around-controlled-moving-cylinder/solid-python/clean.sh
- Left 34 symlinked clean-tutorial.sh files unchanged

Verification:
- Spot-checked scripts across multiple tutorials
- Tested running scripts from parent directories


MakisH:
Hi! This seems to be a duplicate of https://github.com/precice/tutorials/pull/730, applying the same solution.

k4rtikx:
Thanks for pointing this out! I wasn’t aware of **#730** — I’ll make sure to check existing PRs more carefully before picking up similar issues next time.

Issue #672, opened 2025-10-07
Add a rough estimate for tutorial runtime


fsimonis:
It would be nice to have a rough estimate on how long tutorials run, similar to "estimated reading time" on some websites. We could use the all serial setup to give these estimates.

While this is very hardware dependent, it can give the user a good feeling on what to expect.

For example elastic-tube-1d runs in approx 1 minute, heat-exchanger runs in approx 10 minutes, elastic-tube-3d in approx 1h (maybe).

MakisH:
Good point, I thought we had such disclaimers, but apparently only on the heat exchanger: https://precice.org/tutorials-heat-exchanger.html

Other long-run tutorials are:

- Turek-hron FSI3
- 3D elastic tube
- any Nutils flow solvers

The default at the moment is "a couple of minutes". We could easily add alert boxes with disclaimers on the long-run tutorials.

Ayushi6002:
Hi! I’d be happy to help with this.

I can look into adding consistent runtime estimate alert boxes to the long-running tutorials (similar to the heat exchanger example).

Would you prefer:
- Hard-coded estimates per tutorial page, or
- A standardized alert component reused across tutorials?

Happy to prepare a PR once the preferred approach is clear.

MakisH:
Hi @Ayushi6002,

the implementation here would be trivial, but it would need some experience with the cases themselves (running them). If you find a good solution, happy to review it.

Ayushi6002:
Thanks for the clarification!

I’ll first try running a few representative tutorials in serial mode to gather realistic timing estimates, and then propose a consistent format for presenting them across the pages.

I’ll share an approach before opening a PR.

k4rtikx:
Hi @MakisH  @fsimonis 
My plan is to add runtime estimate notes to the tutorial documentation, similar to the alert used in the heat exchanger tutorial.

I will start with a few longer-running tutorials (for example elastic-tube-3d and turek-hron-fsi3) and add rough runtime estimates assuming serial execution.

If there is a preferred format or location for these notes in the tutorial pages, please let me know before I prepare a PR.

MakisH:
Transferring some notes from https://github.com/precice/tutorials/pull/753#issuecomment-4112013474, for later:

- A short text would be enough
- I would not add alert boxes, boldface, or other formatting for anything expected

The short-term approach would be to add alert boxes to the long tutorials.

[...]

PR #753, opened 2026-03-12
Add runtime estimate to elastic-tube-1d tutorial

k4rtikx:
Adds a runtime estimate note to the elastic-tube-1d tutorial.

This follows the suggestion in #672 to provide users with a rough expectation of how long tutorials take to run. The estimate assumes serial execution on a typical laptop and matches the runtime mentioned in the tutorial summary.

If the format is acceptable, similar notes can be added to other tutorials.


MakisH:
Thank you for the suggestion. I think the overall strategy still needs to be defined in #672 before we add messages in individual tutorials. Most important right now would anyway be a clearer disclaimer in the long tutorials.

Having something at the top would be useful, but I think the style can be improved:

- This is currently using a quote box (`>` symbol). Semantically, this is wrong. At the same time, we have the alert boxes that we could use, but a simple text would also be enough.
- This is currently adding a warning triangle in the front, which would make sense when something is unexpected. Since most of our tutorials take 1-2min to run, this looks strange.
- I would not overuse the boldface.

Closing for now, let's continue the discussion in the issue and revisit once there is a clearer plan. In any case, good point for updating only one tutorial first, to discuss.

Issue #393, opened 2023-11-17
Deciding on a accepted tolerance for the fieldcompare


valentin-seitz:
To give a bit of perspective: 
When running the systemtests we execute the coupled participants in the given configuration and then compare exported coupling meshes to a known "good solution" for that participant combination. 

Since we compile with a lot of potential hardware specific optimizations turned on, having different CPU capabilities might lead to floating point differences in the execution. 

Now to stop the systemtests from failing i just guesstimated a number out of thin air which was a relative tolerance of `3e-7`. 

I think is a rather high value, but i also dont have any better estimation. 

This issue is mainly here to discuss how to tackle the problem.

MakisH:
What I remember from our discussion:

Even when running the exact same tests inside the Docker containers, but on different systems (laptop and GitHub runner), floating-point results would have some minor differences, which are currently not fully explainable.

We suspect the following factors:

- Different underlying Linux kernel
- Different CPUs and optimization for different CPUs

uekerman:
>     Different CPUs and optimization for different CPUs

This can indeed lead to floating-point differences.

But I don't understand the title here, `RelTol` should be the solution, no?

valentin-seitz:
I added a more verbose description of the issue

uekerman:
Do the tests run with `3e-7`?

I guess, we should be able to go smaller, but will be try-and-error.
Everything we do is double precision, so sth like `1e-13` should be achievable?

MakisH:
We already use a relative tolerance (of 3e-7).

Note that, for some fields very close to zero, the relative tolerance does not help. We either need to a way to handle such small values, or be able to select which of the exported fields we use for comparisons.

See an example in https://github.com/precice/tutorials/pull/758#issuecomment-4160498433

MakisH:
Documentation of `fieldompare dir --help` (v0.4.0):

```
  -rtol RELATIVE_TOLERANCE, --relative-tolerance RELATIVE_TOLERANCE
                        Specify the relative tolerance to be used. Use e.g. '-rtol pressure:1e-3' to set the tolerance for a field named 'pressure', or '-rtol domain:1e-3' to define the tolerance used when
                        checking the domains for equality.
  -atol ABSOLUTE_TOLERANCE, --absolute-tolerance ABSOLUTE_TOLERANCE
[...]

PR #758, opened 2026-03-30
Add the DuMux adapter and Micro-Manager into the system tests workflows

MakisH:
Integrates the DuMux adapter tests into the regular system tests:

- Adds the `two-scale-heat-conduction` case to the `release_test` suite
- Adds the respective DuMux adapter variables to the nightly and PR workflows

Follow-up of #696 and #737.

The system tests run on this PR used the `tests.yml` from the `develop` branch. Here is ~~[another run](https://github.com/precice/tutorials/actions/runs/23745768212/job/69173892590)~~ that takes the `tests.yml` from this branch as well.

Further runs:
- https://github.com/precice/tutorials/actions/runs/23764549879
- https://github.com/precice/tutorials/actions/runs/23764701005
- https://github.com/precice/tutorials/actions/runs/23784037711/job/69302950180


IshaanDesai (commented):
We indeed need to set the Micro Manager version to `develop`. @Fujikawas, we had this at some point, right? Did we remove it?

Fujikawas:
> We indeed need to set the Micro Manager version to `develop`. @Fujikawas, we had this at some point, right? Did we remove it?

We have `MICRO_MANAGER_VERSION` instead of `MICRO_MANAGER_REF` (I'm not quite sure of the scope of the naming rules for these two).
In https://github.com/precice/tutorials/pull/737, I added the variable `MICRO_MANAGER_PR` to the Dockerfile. I thought this `MM_PR` would represent the latest status for testing when we want to use something newer than `MM_VERSION`, so I didn't add `MM_REF`.

MakisH:
The (partially undocumented) intention of these variables is generally the following:

- `_REF`: The git reference to checkout
- `_PR`: A GitHub PR to fetch, which contains the `_REF` to checkout
- `_VERSION`: A dependency version (fixed)

Since the Micro-Manager is a stage in the Dockerfile, and it is built from its repository, I would rename the `MICRO_MANAGER_VERSION` to `MICRO_MANAGER_REF` in this PR.

I will also add this to the documentation.

MakisH:
@Fujikawas any idea why [this run](https://github.com/precice/tutorials/actions/runs/23764701005) failed with both DuMux cases having issues building DuMux?

Sample log: [system-tests_macro-dumux.log](https://github.com/user-attachments/files/26361259/system-tests_macro-dumux.log)

IshaanDesai:
Seems like the incorrect adapter version got used again.

MakisH:
Ah, I see, I thought that using the default components in that job run would have been enough.

In [another run](https://github.com/precice/tutorials/actions/runs/23784037711/job/69302950180) with the latest components, the free-flow-over-porous-media worked, while the two-scale-heat-conduction failed at the numerical comparisons step, but I think that all of the absolute differences are too small to consider these relative differences valid: [stdout.log](https://github.com/user-attachments/files/26370662/stdout.log)


Until we find a better solution for https://github.com/precice/tutorials/issues/393, I suggest that we exclude this test from the `release_test` suite.

We could also add it again after adding an override option to run the tests but not compare results (partially related to https://github.com/precice/tutorials/issues/404).

tools/tests/tests.yaml
```diff
@@ -45,7 +45,7 @@ test_suites:
         case_combination:
           - macro-dumux
           - micro-dumux
-        reference_result: ./two-scale-heat-conduction/reference-results/macro-dumux_micro-dumux.tar.gz
```
MakisH, line 48: Are we fine with this for now? See explanation in https://github.com/precice/tutorials/pull/758#issuecomment-4160498433

Without this test in the `release_test` suite, the Micro-Manager will not be covered by any nightly test. I would still keep it in the `dumux_test` so that it is triggered in dumux-adapter PRs.

MakisH:
@IshaanDesai ready for review, you can use [this output](https://github.com/precice/tutorials/actions/runs/23784037711/job/69302950180) (without the `two-scale-heat-conduction`, which I removed in https://github.com/precice/tutorials/pull/758/commits/de856c5966be0851e58622551615a9b30a419898) for comparison.

tools/tests/tests.yaml
MakisH: I also added a `micro_manager_test` test suite in https://github.com/precice/tutorials/pull/758/commits/6af9ebde35ea22bb17712f8f2dcc2123d86b0c13, which will be the target of a system tests workflow in the Micro-Manager repository.

IshaanDesai: Is there a place I can see the relative measure of comparison or something similar? I want to find out how small *too small* is, and why the comparison now fails.

IshaanDesai (approved):
I do not fully understand all the changes, but I see they are consistent.

tools/tests/tests.yaml
MakisH: We don't have some single-number metric. I [suggested that to fieldcompare](https://gitlab.com/dglaeser/fieldcompare/-/work_items/69), but this seemed to be complicated at the time.

Looking at the [stdout.log](https://github.com/user-attachments/files/26370662/stdout.log), here are some examples:

```
field-compare-1  | Comparing '/runs/two-scale-heat-conduction_macro-dumux-micro-dumux_2026-03-31-085933/precice-exports/Macro-Mesh-Micro-Manager.dt2.vtk'
field-compare-1  |       and '/runs/two-scale-heat-conduction_macro-dumux-micro-dumux_2026-03-31-085933/reference-results-unpacked/macro-dumux_micro-dumux/Macro-Mesh-Micro-Manager.dt2.vtk'
field-compare-1  |   Reading '/runs/two-scale-heat-conduction_macro-dumux-micro-dumux_2026-03-31-085933/precice-exports/Macro-Mesh-Micro-Manager.dt2.vtk'
field-compare-1  |   Reading '/runs/two-scale-heat-conduction_macro-dumux-micro-dumux_2026-03-31-085933/reference-results-unpacked/macro-dumux_micro-dumux/Macro-Mesh-Micro-Manager.dt2.vtk'
field-compare-1  |    -- Comparing the field 'K00': FAILED
field-compare-1  |      -- Report: Deviation above tolerance detected -> [0.5141269880086446] vs. [0.5141186814702282] ([0.00] %)
field-compare-1  |      -- Predicate: DefaultEquality (abs_tol: 0., rel_tol: 3.e-07)
field-compare-1  |    -- Comparing the field 'K01': FAILED
field-compare-1  |      -- Report: Deviation above tolerance detected -> [-2.8569661945372316e-17] vs. [-3.152737885150525e-15] ([-10935.26] %)
field-compare-1  |      -- Predicate: DefaultEquality (abs_tol: 0., rel_tol: 3.e-07)
field-compare-1  |    -- Comparing the field 'K10': FAILED
field-compare-1  |      -- Report: Deviation above tolerance detected -> [1.0382603233439358e-15] vs. [-1.382181217583312e-15] ([233.12] %)
field-compare-1  |      -- Predicate: DefaultEquality (abs_tol: 0., rel_tol: 3.e-07)
```

This means that, at `dt2`:

- `K00` differed at the 5th decimal (smaller than the printed `0.00%`)
- `K01` values are extremely close to zero (e-15/e-17)
- `K10` the same (e-15)

The relative difference is in parentheses, but it is rather useless.

We could also already experiment with an absolute tolerance computed based on the maximum, see https://github.com/precice/tutorials/issues/393#issuecomment-4162057827

MakisH: @IshaanDesai Actually, could this result (again in `dt2`) show that something in the functionality has recently changed?

```
field-compare-1  |    -- Comparing the field 'Active-Steps': FAILED
field-compare-1  |      -- Report: Deviation above tolerance detected -> [1.] vs. [0.] ([100.00] %)
field-compare-1  |      -- Predicate: DefaultEquality (abs_tol: 0., rel_tol: 3.e-07)
```

IshaanDesai: The data `K01` and `K10` are expected to be close to zero, but the difference in `K00` concerns me, as it suggests the result of the two-scale problem has changed. This is probably because of the change in `Active-Steps`, which should ideally not change at all. All these differences point to the Micro Manager. Let me check the latest changes and try to find potential culprits.

IshaanDesai (approved):
The failure in the comparison of the run results to the reference results is expected. I will handle this in a separate PR. Lets merge this.

PR #772, opened 2026-05-11
Fix Ubuntu 24.04 systemtest Docker user setup

### LINKED ISSUE #766: System test Docker container for Ubuntu 24.04 fails in groupadd
Running the system tests locally using `PLATFORM:ubuntu2404` fails with:

```
#6 [prepare base_image 2/2] RUN groupadd -g 1000 precice && useradd -u 1000 -g 1000 -ms /bin/bash precice
#6 0.243 groupadd: GID '1000' already exists
#6 ERROR: process "/bin/bash -c groupadd -g ${PRECICE_GID} precice && useradd -u ${PRECICE_UID} -g ${PRECICE_GID} -ms /bin/bash precice" did not complete successfully: exit code: 4
------
 > [fluid-openfoam base_image 2/2] RUN groupadd -g 1000 precice && useradd -u 1000 -g 1000 -ms /bin/bash precice:
0.243 groupadd: GID '1000' already exists
```

This is not the case on Ubuntu 22.04.

Also reported in https://github.com/precice/tutorials/pull/731#discussion_r2993229721

PranjalManhgaye:
### Summary

This fixes the Ubuntu 24.04 systemtest Docker build when the requested host UID/GID already exists inside the base image.

With "ubuntu:24.04", UID/GID `1000` can already be used by the default `ubuntu` user/group. The previous Dockerfile always tried to create a new `precice` group and user with those IDs, which failed with:

groupadd: GID '1000' already exists

### Changes

The Dockerfile now checks whether the requested group or user already exists:
- If the requested GID already exists, it reuses that group by renaming it to precice.
- If the requested UID already exists, it reuses that user by renaming it to precice.
- If the UID/GID are unused, it creates the precice group/user as before.
- It ensures /home/precice exists and is owned by the requested UID/GID.
- This keeps the later USER precice behavior working while preserving the host UID/GID mapping used by local system tests.
Fixes  #766.

image

 ##### Both builds ran successfully

Checklist:

### DIFF
--- a/tools/tests/dockerfiles/ubuntu_2404/Dockerfile
+++ b/tools/tests/dockerfiles/ubuntu_2404/Dockerfile
@@ -8,7 +8,26 @@ ENV DEBIAN_FRONTEND=noninteractive
 # If not built via the systemtests.py its either possible to specify manually but 1000 would be the default anyway.
 ARG PRECICE_UID=1000
 ARG PRECICE_GID=1000
-RUN groupadd -g ${PRECICE_GID} precice && useradd -u ${PRECICE_UID} -g ${PRECICE_GID} -ms /bin/bash precice
+RUN set -eux; \
+    if getent group "${PRECICE_GID}" >/dev/null; then \
+        existing_group="$(getent group "${PRECICE_GID}" | cut -d: -f1)"; \
+        if [ "${existing_group}" != "precice" ]; then \
+            groupmod -n precice "${existing_group}"; \
+        fi; \
+    else \
+        groupadd -g "${PRECICE_GID}" precice; \
+    fi; \
+    if getent passwd "${PRECICE_UID}" >/dev/null; then \
+        existing_user="$(getent passwd "${PRECICE_UID}" | cut -d: -f1)"; \
+        if [ "${existing_user}" != "precice" ]; then \
+            usermod -l precice -d /home/precice -m "${existing_user}"; \
+        fi; \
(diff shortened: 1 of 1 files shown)


MakisH (commented):
Thanks for the fix! This is important for developing/debugging locally.

It runs on my system (Ubuntu 22.04, `UID/GID=1000`), and the changes make sense to me.

Since I need to get back into context, could you please elaborate on why this is an issue with the Ubuntu 24.04 builds, but not with the 22.04?

PranjalManhgaye:
Thanks @MakisH for a check , the difference seems to come from the base images themselves. " ubuntu:24.04" already includes a default "ubuntu" user/group with UID/GID `1000`, while " ubuntu:22.04" does not.
I checked this with :
docker run --rm ubuntu:24.04 bash -lc 'getent passwd 1000; getent group 1000'
which returns :
image
For " ubuntu:22.04", the same command does not return a UID/GID `1000` entry.
So the previous Dockerfile worked on 22.04 because `groupadd -g 1000 precice` and `useradd -u 1000 ...` had no collision there. on 24.04, the same commands collide with the existing `ubuntu` user/group before the `precice` user can be created , This patch handles both cases : 1)  if the requested UID/GID already exists, it reuses that entry for `precice`; otherwise  2) it keeps the previous behavior and creates the `precice` user/group normally.

MakisH (approved):

PR #774, opened 2026-05-11
Add apt retries in systemtest Dockerfiles

### LINKED ISSUE #768: Retries in retrieving artifacts
`wget` or `apt-get` might fail when there are network issues. These can not only happen when one runs the tests locally, but also when the respective server or runner has network issues. A common example is Sourceforge, which serves the OpenFOAM Debian packages.

We could use, for example, `wget --tries=3` or similar settings.

See a discussion in https://github.com/precice/tutorials/pull/731#discussion_r2993265332

PranjalManhgaye:
## Summary
This PR adds limited retries to the network-sensitive package and download commands in the systemtest Dockerfiles.

when we build the systemtest Docker images, we depend on external package mirrors and download servers. Sometimes these fail temporarily, even though nothing is wrong with our code or the tutorial setup. In those cases, a local run or CI job can fail just because a mirror was slow or unreachable.

tthis is especially relevant for steps such as installing packages, setting up OpenFOAM, and downloading CalculiX or SU2 sources. with a few bounded retries, we keep the normal behavior the same, but make the builds a bit more robust against temporary network issues.
## Changes

This PR adds bounded retry behavior in both systemtest Dockerfiles:
- `apt-get update` and `apt-get install` now use `Acquire::Retries=3`
- `wget` downloads now use `--tries=3 --waitretry=5`
- Scope is limited to:
  - `tools/tests/dockerfiles/ubuntu_2204/Dockerfile`
  - `tools/tests/dockerfiles/ubuntu_2404/Dockerfile`

This keeps the Dockerfile behavior the same in normal cases, but makes builds more robust against temporary network or mirror issues.

Fixes #768.

## Validation
image

#### Both ran successfully .

Checklist:

### DIFF
--- a/tools/tests/dockerfiles/ubuntu_2204/Dockerfile
+++ b/tools/tests/dockerfiles/ubuntu_2204/Dockerfile
@@ -20,8 +20,8 @@ USER precice
 FROM base_image AS precice_dependecies
 USER root
 # Installing necessary dependecies for preCICE
-RUN apt-get -qq update && \
-    apt-get -qq -y install \
+RUN apt-get -o Acquire::Retries=3 -qq update && \
+    apt-get -o Acquire::Retries=3 -qq -y install \
         build-essential \
         software-properties-common \
         cmake \
@@ -63,9 +63,9 @@ RUN git clone https://github.com/precice/precice.git precice && \
 FROM precice_dependecies AS openfoam_adapter
 ARG OPENFOAM_EXECUTABLE
 USER root
-RUN apt-get update &&\
-    wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
-    apt-get -qq install ${OPENFOAM_EXECUTABLE}-dev &&\
+RUN apt-get -o Acquire::Retries=3 update &&\
+    wget --tries=3 --waitretry=5 -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
+    apt-get -o Acquire::Retries=3 -qq install ${OPENFOAM_EXECUTABLE}-dev &&\
     ln -s $(which ${OPENFOAM_EXECUTABLE} ) /usr/bin/openfoam
 USER precice
 COPY --from=precice /home/precice/.local/ /home/precice/.local/
@@ -95,8 +95,8 @@ FROM precice_dependecies AS fenics_adapter
 COPY --from=python_bindings /home/precice/.local /home/precice/.local
 USER root
 RUN add-apt-repository -y ppa:fenics-packages/fenics && \
-    apt-get -qq update && \
-    apt-get -qq install --no-install-recommends fenics
+    apt-get -o Acquire::Retries=3 -qq update && \
+    apt-get -o Acquire::Retries=3 -qq install --no-install-recommends fenics
 USER precice
 ARG FENICS_ADAPTER_REF
 # Building fenics-adapter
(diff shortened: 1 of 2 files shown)


tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -63,9 +63,9 @@ RUN git clone https://github.com/precice/precice.git precice && \
 FROM precice_dependecies AS openfoam_adapter
 ARG OPENFOAM_EXECUTABLE
 USER root
-RUN apt-get update &&\
-    wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
-    apt-get -qq install ${OPENFOAM_EXECUTABLE}-dev &&\
+RUN apt-get -o Acquire::Retries=3 update &&\
+    wget --tries=3 --waitretry=5 -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
```
MakisH: From `man wget`:

```
 --tries=number
           Set number of tries to number. Specify 0 or inf for infinite retrying.  The default is to retry 20 times, with the exception of fatal errors like "connection refused" or "not found" (404),
           which are not retried.
```

Since the default is 20, why set a smaller value?

Similarly:

```
       --waitretry=seconds
           If you don't want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option.  Wget will use linear backoff, waiting 1 second after the first
           failure on a given file, then waiting 2 seconds after the second failure on that file, up to the maximum number of seconds you specify.

           By default, Wget will assume a value of 10 seconds.
```

Why overwriting the default with a tighter value?

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
```diff
@@ -20,8 +20,8 @@ USER precice
 FROM base_image AS precice_dependecies
 USER root
 # Installing necessary dependecies for preCICE
-RUN apt-get -qq update && \
-    apt-get -qq -y install \
+RUN apt-get -o Acquire::Retries=3 -qq update && \
+    apt-get -o Acquire::Retries=3 -qq -y install \
```
MakisH, line 24: Reading https://manpages.debian.org/buster/apt/apt.conf.5.en.html#THE_ACQUIRE_GROUP I understand that there is no default value, so three retries here sounds valid.

Do we know if this option actually has an effect on the `update`? I am not sure if "acquire" here means both "download the metadata" and "download the packages".

MakisH (changes requested):
Thanks for the PR! Something in this direction could be proven helpful, but I still have some doubts.

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
PranjalManhgaye: thanks , its totally correct , actually i missed the point that wget already has more permmisive defaults than the values i added , so lets keep it in its default behaviour ,

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
PranjalManhgaye: i looked at the apt.conf manpage , from what i understand that the " Acquire " group is not only abt installing `.deb` packages , but about the files apt fetches through its acquire mechanism , the retries option says that apt retries failed files when this value is non-zero , since the `apt-get update` also fetches files, like `Release`/`InRelease` and `Packages` index files, I think this should also apply there.  so i kept on both update and install .

PranjalManhgaye:
thanks,  looking at it again, this PR now only covers the `apt-get` part of #768 by adding explicit retries for apt fetches. for the rest of the issue, `wget` already seems to have retry defaults, and the bigger remaining problem may be things like `git clone`/`git fetch`, especially for external dependencies such as DUNE/GitLab , do you think it is fine to keep this PR as a small apt-only improvement, or would you prefer a broader follow-up for the non-apt downloads?

tools/tests/dockerfiles/ubuntu_2204/Dockerfile
MakisH: @PranjalManhgaye it looks like the `Acquire::Retries=3` is also the default, since APT 2.3.2, which is included in both Ubuntu 22.04 and 24.04, which we currently use:

https://github.com/Debian/apt/blob/main/debian/changelog

MakisH: Besides that, the easier implementation would be to add a system-wide rule to retry: https://askubuntu.com/a/1107071/142834

PranjalManhgaye: i think it may be better to close this pr , If `Acquire::Retries=3` is already the default in the APT versions used by both Ubuntu 22.04 and 24.04, then this PR does not really change the behavior anymore after removing the `wget` options , i will come up with different approach to address this issue .

MakisH:
@PranjalManhgaye let's make this even smaller for now: I would focus on OpenFOAM and add a large `--waitretry` (e.g., 2min). If even that fails, then there is probably a more significant issue with the package repository.

PranjalManhgaye:
@MakisH i will open a new small focused pr for only changing the OpenFOAM repository setup download in the systemtest Dockerfiles by adding a larger `wget --waitretry=120` ,,

Issue #767, opened 2026-05-09
Duplication in tests.yaml


MakisH:
The `tests.yaml` file duplicate cases into every test suite defined in `tests.yaml`. Since we are defining everything in one file, we could use [YAML anchors](https://yaml.com/resources/cheatsheet/#anchors-aliases), as already used in https://github.com/precice/tutorials/pull/731#discussion_r2993380534.

PR #775, opened 2026-05-12
Deduplicate system test entries in tests.yaml

PranjalManhgaye:
## Summary

This PR addresses #767 by reducing duplication in " tools/tests/tests.yaml" , We use YAML anchors and aliases for tutorial entries that appear in multiple test suites. The resolved test configuration stays the same, but the file becomes shorter and easier to maintain when we need to update shared tutorial entries later.

##### What changed

- Added YAML anchors for repeated tutorial entries.
- Replaced duplicated entries with YAML aliases.
- Kept the same test suites, tutorial paths, case combinations, and reference result paths.

#### Why

Before this change, we had the same tutorial entries repeated in several suites. If we wanted to update one shared entry later, we would have to remember to update every copy. With this change, we define the repeated entries once and reuse them where needed. This should make future maintenance simpler without changing how the system tests behave.

## Validation
I checked that this only changes the YAML structure, not the resolved test configuration.
Ran locally:
bash
`git diff --check
python3 -m compileall tools/tests
python3 tools/tests/systemtests.py --help`
I also compared the parsed `tests.yaml` against `upstream/develop`; both resolved to the same `13` test suites and `37` tutorial entries.

Checklist:

### DIFF
--- a/tools/tests/tests.yaml
+++ b/tools/tests/tests.yaml
@@ -10,62 +10,62 @@ test_suites:
   openfoam_adapter_pr:
     tutorials:
       - *quickstart
-      - path: flow-over-heated-plate
+      - &flow_over_heated_plate_openfoam
+        path: flow-over-heated-plate
         case_combination:
           - fluid-openfoam
           - solid-openfoam
         reference_result: ./flow-over-heated-plate/reference-results/fluid-openfoam_solid-openfoam.tar.gz
-      - path: perpendicular-flap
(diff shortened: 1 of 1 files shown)


MakisH:
Thank you for the effort! While this was in the right direction, more cleanup and restructuring was needed in the `tests.yaml`. I rewrote the complete file, also introducing some new and renaming some existing test suites, in https://github.com/precice/tutorials/pull/799

PR #777, opened 2026-05-12
System tests components.yaml: Fix description for various arguments

MakisH:
There is currently a recurrent typo in the `components.yaml`: a `semnantic:` that should be `description:`.

Still, I am wondering whether this is actually used at the moment. I see a reference in https://github.com/precice/tutorials/blob/develop/tools/tests/metadata_parser/metdata.py, but I have not looked closer.

@PranjalManhgaye what do you think?

### DIFF
--- a/tools/tests/components.yaml
+++ b/tools/tests/components.yaml
@@ -32,7 +32,7 @@ python-bindings:
       description: Tutorial git reference to use
       default: "master"
     PYTHON_BINDINGS_REF:
-      semnantic: Git ref of the Python bindings to use
+      description: Git ref of the Python bindings to use
       default: "master"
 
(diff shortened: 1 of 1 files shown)


PranjalManhgaye:
@makish i checked this quickly  , the parser does read the " description" key in `BuildArguments.from_components_yaml()`, and if it is missing it falls back to `No description provided for <argument_name>` , So I think this typo does not affect the actual systemtest runs, since the argument names and defaults are still parsed. But we do lose the intended description metadata, so this cleanup still makes sense to me.

MakisH:
Thanks! Then I am merging this, we can revisit the "what to actually do with this" later.

PR #786, opened 2026-05-17
Remove redundant default substeps options

### LINKED ISSUE #377: Remove substeps="false" option when it becomes supported
Reminder issue, spinning off from https://github.com/precice/precice/issues/1803#issuecomment-1741731097:

> **Important** as soon as quasi-Newton schemes are supported with `substeps="true"`, we should also update the cases to use this feature as they are important for testing the feature.

Related issue that tracks the status of support: https://github.com/precice/precice/issues/1825

cc @BenjaminRodenberg

MakisH:
This PR removes the now redundant default `substeps="true"` option from all tutorial cases where this was defined. Closes https://github.com/precice/tutorials/issues/377.

Related to https://github.com/precice/tutorials/pull/281, https://github.com/precice/tutorials/pull/388, https://github.com/precice/tutorials/pull/498, and https://github.com/precice/tutorials/pull/597 and 

@uekerman any reason to still keep these explicit?

Checklist:

### DIFF
--- a/changelog-entries/786.md
+++ b/changelog-entries/786.md
@@ -1,0 +1,1 @@
+- Remove redundant `substeps="true"` option (default) from the oscillator, oscillator-overlap, partitioned-heat-condiction, and resonant-circuit tutorials. [#786](https://github.com/precice/tutorials/pull/786)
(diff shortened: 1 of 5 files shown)


uekerman (approved):
This should indeed be a "nop" as it is the default now.
Do we have any `substeps="false"` in the tutorials? If yes, we could also think about removing these, but there, we should have a closer look.

MakisH:
> This should indeed be a "nop" as it is the default now. Do we have any `substeps="false"` in the tutorials? If yes, we could also think about removing these, but there, we should have a closer look.

No, the only instances of `substeps` are in these tutorials, with now default values.

PR #801, opened 2026-05-27
Split the system tests log file into stages

PranjalManhgaye:
## Description

Fixes #790.
Right now, `system-tests-stdout.log` and `system-tests-stderr.log` only appear after a test finishes, which makes it hard to inspect failures during long Docker builds or solver runs.
This PR changes the logging so output is written to disk while the test is running. The existing combined logs are still kept, but there are now also separate phase logs:
* `system-tests-build.log`
* `system-tests-solver.log`
* `system-tests-fieldcompare.log`

Scope: 4 files only (~350 LOC). No tutorial or reference-result changes.
Largest change: streaming Docker output in `Systemtest.py`  (replaces communicate() buffering).
I also cleaned up the console output a bit:

* `[n/m]` progress prefix
* less noisy logging (`INFO:` removed)
* clearer grouping/banners
* `PASS` / `FAIL` markers on TTY
The PR is split into 4 commits so the logging change can be reviewed separately from the console formatting cleanup.

## Checklist

- [x] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.

## Resources

- [Contributing tutorials](https://precice.org/community-contribute-to-precice.html#contributing-tutorials)
- [System tests documentation](https://precice.org/dev-docs-system-tests.html)

Issue #789, opened 2026-05-25
Parallel execution of the system tests


MakisH:
Our system tests currently run sequentially, in one GitHub Actions workflow and from one execution of the `systemtests.py`:

1. Trigger GitHub Actions workflow
2. Trigger `systemtests.py` with a test suite
3. Start Docker Compose services for each specific test combination

We could parallelize at any of these levels:

1. Use a matrix in the GitHub Actions workflow and trigger different test suites in parallel.
   - Advantage: potentially using more hosted runners and getting more refined output on GitHub
   - Disadvantage: we can only parallelize on GitHub (which is generally what we need).
2. Parallelize the `systemtests.py` so that it calls Docker Compose in parallel.
   - Advantage: Runs also locally
   - Disadvantage: Will be messy, not portable to other runners.

@PranjalManhgaye what do you think?

PR #821, opened 2026-05-29
Add incremental system test progress to job summary

### LINKED ISSUE #791: System tests job summary is only written at the end
The job summary of the system tests is only written once all tests have been completed, even though the output in the `Run tests` step is produced progressively. Example: https://github.com/precice/tutorials/actions/runs/26386467220/job/77665975926

It would be nice to add the progress to the job summary (e.g., as a bullet point for each test).

@PranjalManhgaye is working on this.

PranjalManhgaye:
Write progress entries to GITHUB_STEP_SUMMARY after each finished system test while keeping the final results table.

## Description

Closes #791.

This PR adds incremental progress reporting to the system tests job summary, instead of only writing the summary table after the full suite completes, the runner now:
- initializes a `System tests progress` section in `GITHUB_STEP_SUMMARY`
- appends one markdown bullet after each finished system test
- keeps the final system test results table at the end
Locally verified incremental GITHUB_STEP_SUMMARY updates (progress bullets after each test + final table). 
This makes long-running system test workflows easier to follow from the GitHub Actions summary page.

### Validation

verified locally by writing to a temporary `GITHUB_STEP_SUMMARY` file and by running `release_test` locally to confirm that progress entries are appended during execution and that the final results table is still written.

## Checklist

- [ ] New tutorial case (e.g., new `fluid-openfoam` folder for existing tutorial)? Add it to the respective `README.md`.
- [ ] New tutorial? Update the website.
  - Add a [sidebar entry](https://github.com/precice/precice.github.io/blob/master/_data/sidebars/tutorial_sidebar.yml)
  - Add it to the [overview](https://github.com/precice/precice.github.io/blob/master/content/tutorials/tutorials.md)

## Resources

- [Contributing tutorials](https://precice.org/community-contribute-to-precice.html#contributing-tutorials)
- [System tests documentation](https://precice.org/dev-docs-system-tests.html)


changelog-entries/821.md
```diff
@@ -0,0 +1 @@
+- System tests now update the GitHub Actions job summary incrementally after each test finishes, with per-test status and timings, while keeping the final results table (fixes [#791](https://github.com/precice/tutorials/issues/791)).
```
MakisH: See similar comments as in #801.

MakisH (commented):
Thank you for looking into this. I tested it in this run: https://github.com/precice/tutorials/actions/runs/26725640841

However, it looks like the summary is only posted after the step finishes. Is there a way to tell GitHub Actions to display the summary while the job is running?

PranjalManhgaye:
thanks for checking ,
but ig on the live summary question: with one long “Run tests” step, i don’t think GitHub will refresh the Summary tab after each test. i do append to GITHUB_STEP_SUMMARY when each test finishes, but the UI usually only shows it when the step ends , so your run looks normal to me, not like the writes failed ,,
For progress while the job is still going, the log is what actually updates (and #801 helps there). and a summary that updates in real time would probably mean changing the workflow  , e.g. more steps — which feels out of scope for this PR.
Happy to either ig go with #821 as “summary fills in at step end” or park the live-update part for a follow-up if you’d rather go that way. let me know what you prefer ,

MakisH:
But then I guess we cannot directly fix #791. Without GitHub Actions updating the summary in real time, adding intermediate output to `GITHUB_STEP_SUMMARY` does not add any value.

Let's instead focus on splitting the `release_test_suite` to intermediate steps (https://github.com/precice/tutorials/issues/789, #829).

PR #830, opened 2026-06-10
Append failed system test log tails to GHA job summary

### LINKED ISSUE #792: System tests job summary does not include the actual error
The job summary includes a table with the run times, but it does not include any logs of the simulations, requiring to download the log files individually.

It would be nice to have the tail (an adequate number of lines) of the log of the failing tests added to the bottom of the job summary.

Related to #790 (if the log files are split, this will need to be adjusted) and #791.

PranjalManhgaye:
## Summary
- on failure now , we append the last (100 lines, collapsible, all stages) of the relevant stage log  ( `build` / `run` / `compare` ) to the GHA job summary.
- stage is inferred from timings  ( `fieldcompare_time` → compare , else `solver_time` → run, else build ).
- Closes #792.
## Test plan
- [x] local quickstart pass — summary shows timing table only
- [x] local failed run — summary shows ( `## Failed test logs` ) with compare log tail
- [x] added changelog entry in `changelog-entries/ 830.md`

### DIFF
--- a/tools/tests/systemtests/Systemtest.py
+++ b/tools/tests/systemtests/Systemtest.py
@@ -14,6 +14,7 @@ from datetime import datetime
 import tarfile
 import time
 
+import html
 import unicodedata
 import re
 import logging
@@ -145,9 +146,14 @@ def _append_failure_log_tails_to_summary(results: List[SystemtestResult]) -> Non
                 print("<details>", file=summary_file)
                 print(f"<summary>{log_name} tail</summary>", file=summary_file)
                 print("", file=summary_file)
-                print("```text", file=summary_file)
-                print(tail, file=summary_file)
-                print("```", file=summary_file)
+                print(
+                    '<div style="max-height: 24em; overflow-y: auto; '
+                    'border: 1px solid #d0d7de; padding: 8px; '
+                    'font-family: monospace; font-size: 12px; white-space: pre-wrap;">',
+                    file=summary_file,
+                )
+                print(html.escape(tail), file=summary_file)
+                print("</div>", file=summary_file)
                 print("</details>", file=summary_file)
                 print("", file=summary_file)
 
--- a/tools/tests/systemtests/Systemtest.py
+++ b/tools/tests/systemtests/Systemtest.py
@@ -31,7 +31,7 @@ STAGE_LOG_FILES = {
     "compare": "system-tests-compare.log",
 }
 
-FAILURE_LOG_TAIL_LINES = 40
+FAILURE_LOG_TAIL_LINES = 100
 
 
 class _SystemtestLogSink:
@@ -121,6 +121,15 @@ def _failing_stage_for_result(result: SystemtestResult) -> str:
     return "build"
 
 
+def _stages_that_ran(result: SystemtestResult) -> List[str]:
+    stages = ["build"]
+    if result.solver_time > 0:
+        stages.append("run")
+    if result.fieldcompare_time > 0:
+        stages.append("compare")
+    return stages
+
+
 def _read_log_tail(log_path: Path, max_lines: int = FAILURE_LOG_TAIL_LINES) -> str:
     if not log_path.is_file():
         return f"(log file not found: {log_path.name})"
@@ -142,18 +151,26 @@ def _append_failure_log_tails_to_summary(results: List[SystemtestResult]) -> Non
     with open(summary_path, "a", encoding="utf-8") as summary_file:
         print("\n## Failed test logs (last lines)\n", file=summary_file)
         for result in failed_results:
-            stage = _failing_stage_for_result(result)
-            log_name = STAGE_LOG_FILES[stage]
-            log_path = result.systemtest.get_system_test_dir() / log_name
-            tail = _read_log_tail(log_path)
+            failing_stage = _failing_stage_for_result(result)
             print(
-                f"### {_success_status_symbol(False)} {result.systemtest} — "
-                f"{stage} log (last {FAILURE_LOG_TAIL_LINES} lines)\n",
+                f"### {_success_status_symbol(False)} {result.systemtest}\n",
                 file=summary_file,
             )
-            print("```text", file=summary_file)
-            print(tail, file=summary_file)
-            print("```\n", file=summary_file)
+            for stage in _stages_that_ran(result):
+                log_path = result.systemtest.get_system_test_dir() / STAGE_LOG_FILES[stage]
+                tail = _read_log_tail(log_path)
+                failure_label = " — failure stage" if stage == failing_stage else ""
+                summary_label = (
+                    f"{stage} log (last {FAILURE_LOG_TAIL_LINES} lines){failure_label}"
+                )
+                print("<details>", file=summary_file)
+                print(f"<summary>{summary_label}</summary>", file=summary_file)
+                print("", file=summary_file)
+                print("```text", file=summary_file)
+                print(tail, file=summary_file)
+                print("```", file=summary_file)
+                print("</details>", file=summary_file)
+                print("", file=summary_file)
 
 
 def display_systemtestresults_as_table(results: List[SystemtestResult]):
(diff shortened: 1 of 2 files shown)


tools/tests/systemtests/Systemtest.py
```diff
@@ -152,11 +197,15 @@ def _get_length_of_name(results: List[SystemtestResult]) -> int:
...
     if "GITHUB_STEP_SUMMARY" in os.environ:
-        with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
-            print("\n\n", file=f)
+        with open(os.environ["GITHUB_STEP_SUMMARY"], "a", encoding="utf-8") as f:
+            print("\n", file=f)
             print(
-                "In case a test fails, download the archive from the bottom of this page and inspect the per-stage logs (`system-tests-build.log`, `system-tests-run.log`, `system-tests-compare.log`). The stage runtimes might already give useful hints.",
+                "For full logs, download the archive from the bottom of this page "
+                "(`system-tests-build.log`, `system-tests-run.log`, `system-tests-compare.log`). "
+                "The stage runtimes in the table above might already give useful hints.",
```
MakisH: Unrelated/unintended change?

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,49 @@ def _success_status_symbol(success: bool) -> str:
     return "✅" if success else "❌"
 
 
+def _failing_stage_for_result(result: SystemtestResult) -> str:
+    if result.fieldcompare_time > 0:
+        return "compare"
+    if result.solver_time > 0:
+        return "run"
+    return "build"
```
MakisH: One would normally implement this by checking status codes (on linux: `echo $?`). However, since we do not have such status codes, this is a valid workaround, since this is what a human would also check at the moment.

MakisH (commented):
Thanks for implementing! A first run looks good: https://github.com/precice/tutorials/actions/runs/27284473467

Since the job summary renders Markdown, I assume it can also render HTML. What if we display more lines (say 100) and put them into [collapsible sections](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections)? Sometimes (like in this job failing at `run`), there is useful information higher up in the other participant.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: yeah keeping as-is per your comment ,,

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: i will revert to the original text

PranjalManhgaye:
thanks @MakisH  for testing this , and as per your feedback i have updated 100 lines per stage (better for interleaved run logs) , collapsible details for each run like :  
1 (build log) 100 lines <details> 100 lines </details>
2 (run log) 100 lines <details> 100 lines </details>
3 (compare log) if error then mentioned failure stage here in case <details> 100 lines </details>
ofc reverted footer and stage detection is unchanged , also tested locally

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,66 @@ def _success_status_symbol(success: bool) -> str:
...
+    summary_path = os.environ.get("GITHUB_STEP_SUMMARY")
+    if not summary_path:
+        return
+
+    failed_results = [result for result in results if not result.success]
+    if not failed_results:
+        return
+
+    with open(summary_path, "a", encoding="utf-8") as summary_file:
+        print("\n## Failed test logs (last lines)\n", file=summary_file)
```
MakisH:
```suggestion
        print("\n## Failed test logs\n", file=summary_file)
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,66 @@ def _success_status_symbol(success: bool) -> str:
...
+        for result in failed_results:
+            failing_stage = _failing_stage_for_result(result)
+            print(
+                f"### {_success_status_symbol(False)} {result.systemtest}\n",
+                file=summary_file,
+            )
+            for stage in _stages_that_ran(result):
+                log_path = result.systemtest.get_system_test_dir() / STAGE_LOG_FILES[stage]
+                tail = _read_log_tail(log_path)
+                failure_label = " — failure stage" if stage == failing_stage else ""
```
MakisH: I would skip marking the "failure stage". On the one hand, a failure in `build` might only manifest at `run`. On the other hand, to simplify the code.

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,66 @@ def _success_status_symbol(success: bool) -> str:
...
+
+    with open(summary_path, "a", encoding="utf-8") as summary_file:
+        print("\n## Failed test logs (last lines)\n", file=summary_file)
+        for result in failed_results:
+            failing_stage = _failing_stage_for_result(result)
+            print(
+                f"### {_success_status_symbol(False)} {result.systemtest}\n",
+                file=summary_file,
+            )
+            for stage in _stages_that_ran(result):
```
MakisH: Instead of checking which stages ran, I would just append all files that exist. That should also let you drop the `_failing_stage_for_result` and `_stages_that_ran` and make the code simpler / the PR shorter.

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,66 @@ def _success_status_symbol(success: bool) -> str:
...
+                log_path = result.systemtest.get_system_test_dir() / STAGE_LOG_FILES[stage]
+                tail = _read_log_tail(log_path)
+                failure_label = " — failure stage" if stage == failing_stage else ""
+                summary_label = (
+                    f"{stage} log (last {FAILURE_LOG_TAIL_LINES} lines){failure_label}"
```
MakisH: Instead of `build log`, `run log` etc, I would reuse the same log file names as one would find in the artifacts archive. That would help one find more details.

I would also drop the `(last 100 lines)` and just write something like "`system-tests-build.log` tail". The number of lines is evident / does not matter.

tools/tests/systemtests/Systemtest.py
```diff
@@ -111,6 +113,66 @@ def _success_status_symbol(success: bool) -> str:
...
+                )
+                print("<details>", file=summary_file)
+                print(f"<summary>{summary_label}</summary>", file=summary_file)
+                print("", file=summary_file)
+                print("```text", file=summary_file)
+                print(tail, file=summary_file)
+                print("```", file=summary_file)
+                print("</details>", file=summary_file)
```
MakisH, line 151: Great! Much better now!

Is there any simple way to allow vertical scrolling in the code snippet? Then, we could display even more lines.

MakisH (commented):
Looks even better now, nice! See https://github.com/precice/tutorials/actions/runs/27292023168

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: done

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: Done , removed the failure-stage label ,,

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: Done , we now iterate the three stage log filenames and include each file that exists on disk. also i removed `_failing_stage_for_result` and `_stages_that_ran` ,,

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: summaries now use e.g. `system-tests-build.log tail` (same names as in the artifact archive) ,,

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: re scrolling: added a scrollable div around log tails in a separate commit ,happy to keep or revert , your call ,,

MakisH: This is now missing the code fences and it is displayed as regular text instead of code: https://github.com/precice/tutorials/actions/runs/27303864243

Let's just revert for now.

PranjalManhgaye: okay

PranjalManhgaye: yeah i reverted , i am not sure but locally scrolling worked fine may be it is unreliable on ci ,,

PranjalManhgaye: @MakisH looked into scrollable logs , [github/markup#893](https://github.com/github/markup/issues/893) suggests there’s no supported way to add scrollbars inside code blocks on GitHub; `<details>` is the common workaround , the `<div>` approach didn’t work in the GHA job summary ,,

MakisH: > yeah i reverted , i am not sure but locally scrolling worked fine may be it is unreliable on ci ,,

How did you render the summary locally?

MakisH (approved):

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: @MakisH i didnt use the real github summary page locally, i just set GITHUB_STEP_SUMMARY to a file like runs/summary-test.md and ran the same code that writes the summary , either a small python script on an existing failed run folder or at the end of systemtests.py , then i opened that file in the editors markdown preview (online) to see how it would look , thats why the scrollable <div> looked fine here but broke on gha, preview and the actual job summary dont render the same way ,,

PR #834, opened 2026-06-12
Add flow-over-heated-plate-particles tutorial

mike-tree-corvid:
add flow-over-heated-plate/fluid-openfoam-basicReactingMultiphaseCloud tutorial case to show verification of implicit coupling of Lagrangian particles

TODO: Summarize and motivate the changes, link to issues, remove the checklist entries that are not relevant.

## Checklist

For new tutorials or tutorial cases:

- [x] I followed the [tutorial folder structure](https://precice.org/community-contribute-to-precice.html#contributing-tutorials)
- [x] I added/updated the tutorial `README.md`
- [x] I added/updated the tutorial `metadata.yaml`
- [ ] I added tests in `tools/tests/tests.yaml`
- [ ] I submitted a pull request to the website with:
   - An entry in [`_config.yaml`](https://github.com/precice/precice.github.io/blob/master/_config.yml)
   - A [sidebar entry](https://github.com/precice/precice.github.io/blob/master/_data/sidebars/tutorial_sidebar.yml)
   - An entry in the [tutorials overview](https://github.com/precice/precice.github.io/blob/master/content/tutorials/tutorials.md)


mike-tree-corvid:
corresponds with precice/openfoam-adatper pull request here:

https://github.com/precice/openfoam-adapter/pull/405

MakisH:
Thanks for contributing the tutorial as well! I have not yet looked closer, but the problem setup is different than the current flow-over-heated-plate tutorial, right? In that case, the convention would be to make a separate tutorial directory (e.g., `flow-over-heated-plate-particles` or similar, whatever describes it best) and describe the case in the respective `README.md`.

mike-tree-corvid:
> Thanks for contributing the tutorial as well! I have not yet looked closer, but the problem setup is different than the current flow-over-heated-plate tutorial, right? In that case, the convention would be to make a separate tutorial directory (e.g., `flow-over-heated-plate-particles` or similar, whatever describes it best) and describe the case in the respective `README.md`.

The problem setup is the exact same as the flow-over-heated-plate tutorial with the exception that the necessary files / settings are such that 100 Lagrangian parcels / particles (1 particle per parcel) are injected at time = 0 and are left to propagate through the domain. The intent was to give the particles the same properties as the fluid so that they simply follow the flow rather than affect it, thus making any changes in the solution potential flags for something being wrong with the Lagrangian particle implicit coupling implementation.

I debated whether to slide this contribution under the flow-over-heated-plate tutorial, or make it a stand alone tutorial. I'm happy to do either. Do you want to wait to decide what should be done until you take a closer look, or are you pretty decided that it should be a separate tutorial and I should go ahead and make the change?

MakisH:
I will take a closer look to this tutorial, the adapter PR, and the related discussion in ~a week.

For comparison, we have:

- [channel transport](https://precice.org/tutorials-channel-transport.html)
- [channel transport with particles](https://precice.org/tutorials-channel-transport-particles.html)

as separate tutorials.

mike-tree-corvid:
> I will take a closer look to this tutorial, the adapter PR, and the related discussion in ~a week.
> 
> For comparison, we have:
> 
>     * [channel transport](https://precice.org/tutorials-channel-transport.html)
> 
>     * [channel transport with particles](https://precice.org/tutorials-channel-transport-particles.html)
> 
> 
> as separate tutorials.

I reformatted this contribution as its own tutorial rather than being a subset of the flow-over-heated-plate tutorial.

Anything else you see that's needed before we can take this off draft status?

MakisH:
@mike-tree-corvid with a quick look, it seems like the structure is fine and I have everything I need to do a review. I am in a different context than OpenFOAM these days, but I will eventually do a detailed review together with the changes in the adapter.

Regarding the CI failure:

```
flow-over-heated-plate-particles/images/temperature_with_particles.pvsm: 1308 kb exceeds the limit of 750 kb. 
flow-over-heated-plate-particles/images/temperature_with_particles.gif: 3716 kb exceeds the limit of 2200 kb.
```

if you could reduce a bit the size of the GIF file (e.g., lower frequency), it would be great, otherwise I will see how we can better handle that. The `.pvsm` file is not something we typically have, but it could be useful. If portable, I will add an exception.

flow-over-heated-plate-particles/solid-openfoam/0/T
```diff
@@ -0,0 +1,47 @@
+/*--------------------------------*- C++ -*----------------------------------*\
+| =========                 |                                                 |
+| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
+|  \\    /   O peration     | Version:  5                                     |
+|   \\  /    A nd           | Web:      www.OpenFOAM.org                      |
+|    \\/     M anipulation  |                                                 |
+\*---------------------------------------------------------------------------*/
```
MakisH: Note that we typically don't have these headers and comment separateor lines in our configuration files, unless they are really from OpenFOAM (and the version is important).

mike-tree-corvid: removed header with recent commit

mike-tree-corvid:
> @mike-tree-corvid with a quick look, it seems like the structure is fine and I have everything I need to do a review. I am in a different context than OpenFOAM these days, but I will eventually do a detailed review together with the changes in the adapter.
> 
> Regarding the CI failure:
> 
> ```
> flow-over-heated-plate-particles/images/temperature_with_particles.pvsm: 1308 kb exceeds the limit of 750 kb. 
> flow-over-heated-plate-particles/images/temperature_with_particles.gif: 3716 kb exceeds the limit of 2200 kb.
> ```
> 
> if you could reduce a bit the size of the GIF file (e.g., lower frequency), it would be great, otherwise I will see how we can better handle that. The `.pvsm` file is not something we typically have, but it could be useful. If portable, I will add an exception.

I coarsened the .gif file so it takes up less space. I removed the .pvsm file.

flow-over-heated-plate-particles/solid-openfoam/0/T
MakisH: Oh, I see that the original case (only this one) still has such headers. I will clean this up.

flow-over-heated-plate-particles/fluid-openfoam/system/controlDict
```diff
@@ -0,0 +1,53 @@
...
+
+application     reactingParcelFoam;
+
+startFrom       startTime;
+
+startTime       0;
+
+stopAt          endTime;
+
+endTime         20;
```
MakisH, line 17: Is the t=20 important (here and in the `precice-config.xml`)? Does it already show clear-enough results earlier? The tutorials are meant to run in <1-2min, unless there is a good reason.

flow-over-heated-plate-particles/images/tutorials-flow-over-heated-plate-particles-example.png
MakisH, line 1: This image needs to be updated for the current tutorial, to show the particles. I can do that.

flow-over-heated-plate-particles/images/tutorials-flow-over-heated-plate-particles-precice-config.png
MakisH, line 1: If nothing changes in the preCICE or adapter configuration, there is a question on whether this is best published as a preCICE tutorial, or as an [application case](https://precice.org/community-guidelines-application-cases.html), which will later be showcased on the preCICE website. @uekerman what do you think?

Tradeoffs:

- as a tutorial:
   - we maintain it easily, directly here
   - it is visible next to other tutorials as a regular tutorial
   - we misuse a bit the role of the tutorials: the user does not learn how to do something different, but that the adapter can handle a certain type of simulations.
- as an application case:
   - you/we maintain it in another repository
   - it is visible next to other application cases, with your name and details (but that list has not yet been created, besides what we have as examples on the [website](https://precice.org/community-guidelines-application-cases.html) and [under review](https://github.com/precice/preeco-orga/pulls))
   - more freedom to shape it as you wish (e.g., longer, more complex, more auxiliary files)
- either way:
   - we showcase that the feature works
   - we can integrate it into the system tests
   - it has the same structure, and the only difference is where we render it and who maintains it.

Note that we already discussed the "specialization of the `flow-over-heated-plate/fluid-openfoam` vs `dedicated tutorial`", and I still think that this is too complex to be just a specialization.

In any case, the same review process is needed for both, and making it ready for the tutorials is also making it ready to publish as an application case. So I suggest that we continue with the review normally here for now.

tools/openfoam-remove-empty-dirs.sh
```diff
@@ -11,7 +11,7 @@ openfoam_remove_empty_dirs() {
 				rm -rf "${f}"
 			fi
 		done
-		if [ -d processor0 ]; then
+		if [ -d processor0 ] || [ -d processor* ]; then
```
MakisH, line 14: What is the motivation for this change? In which situation will `processor0` not be there, but other `processor*` directories will be?

flow-over-heated-plate-particles/README.md
```diff
@@ -0,0 +1,81 @@
...
...ticles involves not only accounting for existing particles in the domain during checkpointing, but also adjusting the particle injection logic for implicit coupling loops within time-windows as well. Particle injection is handled in OpenFOAM via injectionModels. injectionModels lie outside the OpenFOAM solver objectRegistry, so they cannot be directly modified by the existing OpenFOAM preCICE adapter functionObject. Therefore, the OpenFOAM preCICE adapter source code was modified to build an additional library, `libImplicitInjectors.so`. `libImplicitInjectors.so` contains "implicit" versions of the following existing OpenFOAM-native injectionModel types: patchInjection, coneNozzleInjection, and manualInjection. In practice, the user simply needs to make sure they have the `libImplicitInjectors.so` library loaded in their `system/ControlDict` file (alongside `libpreciceAdapterFunctionObject.so`) and then swap a native OpenFOAM injectionModel type (like patchInjection) for an "implicit" version (implicitPatchInjection) in the injectionModels sub-dictionary of their particle cloud file. The following table maps native OpenFOAM injectionModels to their corresponding "implicit" version:
```
MakisH, line 20: This reads a bit like a pull request description. Assuming that the feature would be merged in the OpenFOAM adapter, this could be written from the perspective of "this type of OpenFOAM simulation works as well".

Some of this information (including the table) could go into the feature `README.md` (see adapter PR) and/or to the adapter documentation.

But these are details that I could also directly sort out at the end of the review.

MakisH (commented):
Thank you for all the updates, I think I now understand enough to catch up with the progress here. I added a few minor points and started a thread to discuss where to actually host and render this in the end.

tools/openfoam-remove-empty-dirs.sh
mike-tree-corvid: OpenFOAM has the option to collate its file output. By default, OpenFOAM creates a single directory for every processor on which it's run. The system administrator on the systems where I run OpenFOAM didn't like this very much because of the number of small files created by my large simulations, so I run sims using `OptimisationSwitches { fileHandler collated; }` in my system/controlDict file. This results in a SINGLE processor directory, but the name of the directory changes depending on the number of processors on which the simulation is run. For example, if I run a simulation using 64 procs, the name of the directory is "processors64". If I run a simulation using 128 processors, the name of the directory is "processors128". If one is running with collated file handling they will never have a "processor0" directory.

flow-over-heated-plate-particles/fluid-openfoam/system/controlDict
mike-tree-corvid: I ran it for 20 seconds to make sure some particles make it from one end of the domain to the other. The particles don't reach the fluid next to the heated plate until t=5. The first particles make it all the way across the heated plate by about t=12. t=20 ran in about 5-6 minutes wall clock time on my setup, but I understand that run times can vary wildly. With that said, I'll set the endTime to whatever you suggest.

tools/openfoam-remove-empty-dirs.sh
MakisH: Ah, I didn't know that, thanks! Then, I guess that renaming `processor0` to `processor*` would cover all cases. I will bring that directly to develop. -> Does not, I will find a more general solution.

MakisH: https://github.com/precice/tutorials/pull/879

flow-over-heated-plate-particles/images/tutorials-flow-over-heated-plate-particles-precice-config.png
uekerman: I would tend towards including this as a tutorial.

> we misuse a bit the role of the tutorials: the user does not learn how to do something different, but that the adapter can handle a certain type of simulations.

"Misuse" is too strong IMO. This case is a similar kind of extension to many of the cases we already have.

There is a huge grey area, but a "real" application case is sth connected to a publication IMO, see the [three examples in the introduction of the guidelines](https://precice.org/community-guidelines-application-cases.html).

PR #838, opened 2026-06-14
Add solids4foam to the system tests

MakisH:
Adds a solids4foam component and adds the [solids4foam tutorial case](https://github.com/precice/tutorials/tree/develop/perpendicular-flap/solid-solids4foam) to the system tests.

To avoid building it from scratch every time that a previous build stage is updated, I am defining an independent build stage based on the [official solids4foam image](https://hub.docker.com/r/solids4foam/solids4foam/tags) (FYI @philipcardiff: this image is useful! :pray:).

Since the image is based on Ubuntu 22.04, and since anyway we test the adapter and the respective tutorial with various preCICE versions, I am using a fixed version of preCICE from scratch. I thought of moving that to a different Dockerfile, but a bug in the system tests currently prevents that: https://github.com/precice/tutorials/issues/873

Keeping the two components (`openfoam-adapter` and `solids4foam`) independent also allows them to use different OpenFOAM versions.

To be able to test a complete simulation ([test run](https://github.com/precice/tutorials/actions/runs/29103797145)), I cherry-picked the current state of #872. I will clean this up before merging.


perpendicular-flap/solid-solids4foam/system/fvSchemes
MakisH, line 1: Cherry-picked #872 for testing. Revert / update.

PR #847, opened 2026-06-17
Add per-test fieldcompare tolerance and skip_compare

### LINKED ISSUE #837: Overriding the fieldcompare tolerance (or skipping)
Some cases are currently always failing at the fieldcompare step, as they include values too close to machine precision. There are currently defined in a separate test suite:

```yaml
  expected-to-fail:
    tutorials:
      - *elastic-tube-3d_fluid-openfoam_solid-fenics # too small values to compare
      - *two-scale-heat-conduction_macro-dumux_micro-dumux # too small values to compare
```

It would be useful to have a per-test option, similar to the `max_time`, to override the tolerance used by fieldcompare, or skip the step altogether. For example:

```diff
      - &elastic-tube-3d_fluid-openfoam_solid-fenics
         path: elastic-tube-3d
         case_combination:
           - fluid-openfoam
           - solid-fenics
         max_time_windows: 1
         reference_result: ./elastic-tube-3d/reference-results/fluid-openfoam_solid-fenics.tar.gz
+        tolerance: 1e-2
```

Alternatively, we could revisit these cases to see why they produce such values, and avoid them.

PranjalManhgaye:
### summary
closes #837

i added optional tolerance and skip_compare fields in tests.yaml so we can override fieldcompare rtol per test (or skip compare entirely if we only care about build + run).

before this we had some tests stuck in expected-to-fail or commented out in component suites because exported vtk values are tiny and the default 3e-7 rtol basically always fails. now those cases use tolerance: 1e-2 instead.

### what changed
- parse tolerance and skip_compare in the system test yaml loader
- pass rtol into the fieldcompare docker compose template (-rtol {{ tolerance }})
- wire it through systemtests.py and generate_reference_results.py
- remove the expected-to-fail suite
- re-enable the affected tests in fenics-adapter, openfoam-adapter, micro-manager, and nutils-adapter
- docs + changelog entry
tests without tolerance still use the default 3e-7 so behavior for everything else should be the same.

### local testing
i tried to run things locally on my machine , elastic-tube-3d (fluid-openfoam + solid-fenics) : full build / run / fieldcompare pass with -rtol 0.01 (32/32 comparisons) , this is the main case we cared about for this issue.

macro-nutils_micro-nutils : reference tarball doesn’t exist on develop yet so i couldn’t validate that one end-to-end
so yeah locally we’ve got solid proof for the fenics case ; the rest we’ll need ci .

### notes for reviewers
would be great to trigger system tests on fenics-adapter, openfoam-adapter, micro-manager, dumux-adapter (and nutils-adapter once the nutils reference exists)
macro-nutils_micro-nutils is re-enabled in yaml but the reference file might still be missing upstream : heads up if ci complains there

happy to tweak tolerance values if 1e-2 is too loose or too tight for any of the three cases ,,


tools/tests/tests.yaml
```diff
@@ -729,9 +723,7 @@ test_suites:
       - *quickstart_openfoam_cpp
       - *volume-coupled-flow_fluid-openfoam_source-nutils
       - *water-hammer_fluid1d-left-nutils_fluid3d-right-openfoam
-
-      # Excluded:
-      # *elastic-tube-3d_fluid-openfoam_solid-fenics # too small values to compare
+      - *elastic-tube-3d_fluid-openfoam_solid-fenics
```
MakisH: Lists in this file are meant to be sorted alphabetically. Similar issue in the other previously commented-out tests in this file.

tools/tests/README.md
```diff
@@ -106,6 +106,8 @@ fieldcompare dir precice-exports/ reference-results-unpacked/<case>/ \
              -rtol 3e-7
 ```
 
+The default relative tolerance is `3e-7`. Per-test overrides are available in `tests.yaml` via `tolerance` (passed to fieldcompare as `-rtol`). Set `skip_compare: true` to skip the comparison step and only verify that build and run succeed.
```
```suggestion
The default relative tolerance (`-rtol`) is `3e-7`. Per-test overrides are possible in `tests.yaml` (e.g., `tolerance: 1e-2`). Set `skip_compare: true` to skip the comparison step and only verify that the build and run steps succeed.
```

tools/tests/generate_reference_results.py
MakisH, line 1: Why does the `generate_reference_results.py` need information about numerical regression checks?

changelog-entries/847.md
```diff
@@ -0,0 +1 @@
+- Add optional `tolerance` and `skip_compare` fields to `tests.yaml` for per-test fieldcompare configuration closes #837.
```
MakisH:
```suggestion
- Add optional `tolerance` and `skip_compare` fields to `tests.yaml` for per-test fieldcompare configuration [#847](https://github.com/precice/tutorials/pull/847)
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -21,6 +21,7 @@
 
 
 GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 600))
+DEFAULT_FIELDCOMPARE_RTOL = 3e-7
```
MakisH, line 27: As more and more parameters accumulate here, I am wondering if we need a `defaults` dictionary, to reduce the number of arguments we pass aroung. But maybe implement that in a separate PR.

tools/tests/systemtests/Systemtest.py
```diff
@@ -669,6 +673,15 @@ def _run_field_compare(self):
...
 
+    def _log_skipped_fieldcompare(self) -> None:
+        log_sink = getattr(self, "_log_sink", None)
+        if log_sink is not None:
+            log_sink.begin_stage("compare")
+            log_sink.append_stdout(
+                f"(skipped: skip_compare=true, default rtol would be {DEFAULT_FIELDCOMPARE_RTOL})",
+                "compare",
+            )
+
```
MakisH: Why is a dedicated function needed to log that the `fieldcompare` step was skipped?

tools/tests/systemtests/Systemtest.py
```diff
@@ -219,6 +220,8 @@ class Systemtest:
     max_time: float | None = None
     max_time_windows: int | None = None
     timeout: int = GLOBAL_TIMEOUT
+    tolerance: float = DEFAULT_FIELDCOMPARE_RTOL
+    skip_compare: bool = False
```
MakisH, line 228: I am a bit torn on whether the `skip_comparison` is needed. I would also rather formulate it positively: `compare_results: false`. One reason to remove it would be that it seems that most of the code additions come due to that. While we could completely skip the comparisons to work around failing runs, these would anyway fail at the `run` step.

~~I would suggest removing it. Setting a large `tolerance` already provides the capability to effectively skip comparisons.~~

Edit: Here is a use case: an external test case that provides the case configuration but not any reference results. Let's keep it. Or make it that missing `reference_result` means automatic skipping, with a warning.

MakisH (commented):
Thanks for implementing!

Here are some comments to apply before testing on GHA.

tools/tests/generate_reference_results.py
PranjalManhgaye: that script only builds and runs to produce reference tarballs , so it doesn’t need tolerance/skip_compare , reverted those lines; systemtests.py still passes them for the actual regression run.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: a defaults dict would help as more parameters accumulate , i would prefer to keep that for a separate refactor PR so this one stays focused.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: removed the helper , skip is now just a logging.info at the call site ,,

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: keeping skip_compare for now , external tests with config but no reference tarball (#732) is a good use case , happy to revisit compare_results : false or auto-skip on missing reference_result in a follow-up ,,

MakisH (commented):
[Test on GHA](https://github.com/precice/tutorials/actions/runs/28332880882) - state:

- the two-phase-heat-conduction cases fail to build the `micro-dumux` participant. This is irrelevant, I will investigate. -> See https://github.com/precice/tutorials/pull/852 -> results again have a too large range, but the feature works - disabled comparisons for now, see #393 for a potential solution.
- the `Elastic tube 3D (fluid-openfoam, solid-fenics)` case passes, but running takes long, and comparing takes even longer. Maybe this just happened once, I will investigate. -> See https://github.com/precice/tutorials/issues/853 -> disabled the comparisons for this case ([test](https://github.com/precice/tutorials/actions/runs/28388776654))

MakisH (approved):
I did some more tests and adjustments, and opened a few issues.

The two-scale-heat-conduction case was not running. It runs now, but the results are still not covered by a larger tolerance. See https://github.com/precice/tutorials/issues/393#issuecomment-4162057827 for a potential solution.

The elastic-tube-3d/solid-fenics takes too long to compare, so I have disabled the comparison of that one as well for now.

I added the following to the `release` test suite:

- elastic-tube-3d_fluid-openfoam_solid-fenics
- two-scale-heat-conduction_macro-dumux_micro-dumux

and the following to the `extra`:

- two-scale-heat-conduction_macro-nutils_micro-nutils (will not work before the next release of the micro-manager)

PR #848, opened 2026-06-20
Add per-component build_timeout for system test Docker builds

### LINKED ISSUE #840: Timeout for building system tests containers
While #735 added a timeout for the solver execution in the system tests, the building step is still covered by the global timeout, and the relation between the two is a bit tangled.

Since some components take longer to build, we would need a timeout for building. Since different components take different time to build, having the timeout at the component level would make sense.

This would currently help at least the DuMux adapter (see #822) and solids4foam (see #838), and would let us make the global timeout shorter, which would be very helpful for debugging.

For example:

```diff
 dumux-adapter:
   repository: https://github.com/precice/dumux-adapter
   template: component-templates/dumux-adapter.yaml
+  build_timeout: 600
   build_arguments:
     PLATFORM:
       description: Dockerfile platform used
       default: "ubuntu_2404"
     PRECICE_REF:
       ...
```

PranjalManhgaye:
## Summary
-  add optional `build_timeout` to components in `components.yaml`  (build step only)
- resolve per-test build timeout as the maximum across distinct participants
-   set `dumux-adapter : build_timeout: 1800` for long Dumux Docker builds
- Document separate build vs run/compare timeouts in readme
fixes   #840


changelog-entries/848.md
```diff
@@ -0,0 +1 @@
+- Add optional per-component `build_timeout` in `components.yaml` for the system test Docker build step ([#848](https://github.com/precice/tutorials/pull/848)).
```
MakisH: Remember, changelogs are written in past tense:

```suggestion
- Added optional per-component `build_timeout` in `components.yaml` for the system test Docker build step ([#848](https://github.com/precice/tutorials/pull/848)).
```

tools/tests/metadata_parser/metdata.py
```diff
@@ -97,6 +97,7 @@ class Component:
     name: str
     template: str
     parameters: BuildArguments
+    build_timeout: Optional[int] = None
```
MakisH: I think that `Optional` is not the current way of implementing optional arguments in Python. One reference: https://stackoverflow.com/a/63442322/2254346

tools/tests/systemtests/Systemtest.py
```diff
@@ -21,6 +21,8 @@
 
 
 GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 600))
+DEFAULT_BUILD_TIMEOUT = int(
```
MakisH, line 25: Similar comment as in https://github.com/precice/tutorials/pull/847: I get the impression we need a dictionary of defaults here. But maybe in a separate PR.

tools/tests/components.yaml
```diff
@@ -50,6 +50,7 @@ dealii-adapter:
 
 dumux-adapter:
   template: component-templates/dumux-adapter.yaml
+  build_timeout: 1800
```
MakisH: DuMux already builds with 600s. Let's set this to 600s, and reduce the `GLOBAL_TIMEOUT` to 300s. The SU2 adapter probably also needs the 600s.

There will be more cases with longer timeouts later.

tools/tests/README.md
```diff
@@ -329,8 +329,10 @@ This template defines:
 
 ### Timeouts
 
-A `GLOBAL_TIMEOUT` is used for all operations. Its default value is 600s (5min), it is set in the beginning of [`Systemtests.py`](https://github.com/precice/tutorials/blob/develop/tools/tests/systemtests/Systemtest.py), and it can be overridden via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
+Build and run/compare use separate timeouts.
```
```suggestion
The build and and the run/compare steps use separate timeouts.
```

tools/tests/README.md
```diff
@@ -329,8 +329,10 @@ This template defines:
...
 
-Tests can define a different `timeout` in their `tests.yaml` entry, which applies to the running and results comparison steps.
+**Build:** Each component in `components.yaml` may define `build_timeout` (seconds). For a test, the build step uses the maximum `build_timeout` among its components; components without `build_timeout` use `DEFAULT_BUILD_TIMEOUT` (600s by default, same as `GLOBAL_TIMEOUT`). Override the default via `PRECICE_SYSTEMTESTS_BUILD_TIMEOUT`.
```
MakisH: Does that mean that if I have a test that uses OpenFOAM and DuMux, building the OpenFOAM Docker image will timeout at the DuMux build timeout? Why is that?

MakisH (commented):
Thanks for implementing! Here are some comments before testing on GHA.

tools/tests/metadata_parser/metdata.py
PranjalManhgaye: switched to int | None for build_timeout.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: happy to do a defaults dict in a follow-up refactor pr

tools/tests/components.yaml
PranjalManhgaye: thanks

tools/tests/README.md
PranjalManhgaye: docker compose build is a single subprocess for all images , so there is one wall-clock limit , we take the maximum so the slowest adapter is not cut off early ; faster images are not capped to a shorter per-component limit , clarified this in the README and in _resolve_build_timeout s docstring.

tools/tests/README.md
```diff
@@ -329,8 +329,10 @@ This template defines:
 
 ### Timeouts
 
-A `GLOBAL_TIMEOUT` is used for all operations. Its default value is 600s (5min), it is set in the beginning of [`Systemtests.py`](https://github.com/precice/tutorials/blob/develop/tools/tests/systemtests/Systemtest.py), and it can be overridden via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
+The build and the run/compare steps use separate timeouts.
 
-Tests can define a different `timeout` in their `tests.yaml` entry, which applies to the running and results comparison steps.
+**Build:** Each component in `components.yaml` may define `build_timeout` (seconds). The build step runs `docker compose build` once for all participant images, with one wall-clock subprocess timeout. That limit is the maximum `build_timeout` among the test's distinct components (so the slowest adapter, e.g. DuMuX or SU2, is not cut off early). Components without `build_timeout` contribute `DEFAULT_BUILD_TIMEOUT` (600s by default). Override the default via `PRECICE_SYSTEMTESTS_BUILD_TIMEOUT`.
```
MakisH: Plus some minor changes to make this shorter, easier to read, and more accurate.

We generally don't need to refer to the exact variables in the scripts, if some section is user-facing.

```suggestion
**Build:** Each component in `components.yaml` may set a `build_timeout` (default: 600s). The build step runs `docker compose build` once for all participant images, with one wall-clock subprocess timeout. That limit is the maximum `build_timeout` among the distinct components of the test (so that the slowest component is not cut off too early). You can override the default via the `PRECICE_SYSTEMTESTS_BUILD_TIMEOUT` environment variable.
```

tools/tests/README.md
```diff
@@ -329,8 +329,10 @@ This template defines:
...
+
+**Run and compare:** `GLOBAL_TIMEOUT` defaults to 300s and can be overridden via `PRECICE_SYSTEMTESTS_TIMEOUT`. Tests can define a different `timeout` in their `tests.yaml` entry, which applies to the running and results comparison steps only.
```
```suggestion
**Run and compare:** Each test in `tests.yaml` may set a `timeout` (default: 300s), which applies to the running and results comparison steps only. You can override the default via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -21,7 +21,9 @@
 import os
 
 
-GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 600))
+GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 300))
+DEFAULT_BUILD_TIMEOUT = int(
+    os.environ.get("PRECICE_SYSTEMTESTS_BUILD_TIMEOUT", 600))
```
MakisH: To be consistent, this also needs to be 300. Do I overlook anything?

```suggestion
    os.environ.get("PRECICE_SYSTEMTESTS_BUILD_TIMEOUT", 300))
```

MakisH:
A test run for the `release` test suite: https://github.com/precice/tutorials/actions/runs/28336482485 (attempt #2 is with a cleared cache)

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: @MakisH You are not overlooking anything - builds usually need more time than runs, so I kept DEFAULT_BUILD_TIMEOUT at 600 for the docker compose build and only lowered GLOBAL_TIMEOUT to 300 for run/compare. That's the point of splitting them in this PR.

MakisH: It looks like 300 will be anyway too tight also for the first component built after clearing the cache: https://github.com/precice/tutorials/actions/runs/28336482485

Let's keep the default at 600 for now, unless you have any better idea.

I am trying to minimize the time that someone would wait for a build to time out.

MakisH: Setting it to 8min should already help:

```suggestion
    os.environ.get("PRECICE_SYSTEMTESTS_BUILD_TIMEOUT", 480))
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -21,7 +21,9 @@
 import os
 
 
-GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 600))
+GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 300))
```
MakisH: Let's also set this to a tighter value:


```suggestion
GLOBAL_TIMEOUT = int(os.environ.get("PRECICE_SYSTEMTESTS_TIMEOUT", 180))
```

tools/tests/README.md
```diff
@@ -330,8 +330,10 @@ This template defines:
 
 ### Timeouts
 
-A `GLOBAL_TIMEOUT` is used for all operations. Its default value is 600s (5min), it is set in the beginning of [`Systemtests.py`](https://github.com/precice/tutorials/blob/develop/tools/tests/systemtests/Systemtest.py), and it can be overridden via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
+The build and the run/compare steps use separate timeouts.
 
-Tests can define a different `timeout` in their `tests.yaml` entry, which applies to the running and results comparison steps.
+**Build:** Each component in `components.yaml` may set a `build_timeout` (default: 600s). The build step runs `docker compose build` once for all participant images, with one wall-clock subprocess timeout. That limit is the maximum `build_timeout` among the distinct components of the test (so that the slowest component is not cut off too early). You can override the default via the `PRECICE_SYSTEMTESTS_BUILD_TIMEOUT` environment variable.
+
+**Run and compare:** Each test in `tests.yaml` may set a `timeout` (default: 300s), which applies to the running and results comparison steps only. You can override the default via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
```
```suggestion
**Build:** Each component in `components.yaml` may set a `build_timeout` (default: 480s (8min)). The build step runs `docker compose build` once for all participant images, with one wall-clock subprocess timeout. That limit is the maximum `build_timeout` among the distinct components of the test (so that the slowest component is not cut off too early). You can override the default via the `PRECICE_SYSTEMTESTS_BUILD_TIMEOUT` environment variable.

**Run and compare:** Each test in `tests.yaml` may set a `timeout` (default: 180s (3min)), which applies to the running and results comparison steps only. You can override the default via the `PRECICE_SYSTEMTESTS_TIMEOUT` environment variable.
```

MakisH (approved):
I made both timeouts a bit shorter and ran again all tests starting from a clean cache - all work: https://github.com/precice/tutorials/actions/runs/28362375986

Ready to merge!

PR #849, opened 2026-06-21
Add run-before / run-after hooks for system tests

### LINKED ISSUE #827: Run a command before starting a test
Most tutorials work as system tests by navigating into each and running `./run.sh`.

However, some tutorials provide alternatives that we should also test, and which require running some script to modify configuration files before starting the simulation.

In the following tutorials:

- [partitioned-pipe-multiscale](https://github.com/precice/tutorials/tree/develop/partitioned-pipe-multiscale)
- [water-hammer](https://github.com/precice/tutorials/tree/develop/water-hammer)

we have a `set-case.sh` script, which we want to standardize. This script can take arguments. For example, `./set-case.sh 3d1d` switches the `precice-config.xml` symlink to point to `precice-config-3d-1d.xml`.

It would be nice to be able to write the following in the [`tests.yaml`](https://github.com/precice/tutorials/blob/develop/tools/tests/tests.yaml:

```diff
   water-hammer:
     tutorials:
       - &water-hammer_fluid1d-left-nutils_fluid3d-right-openfoam
         path: water-hammer
         case_combination:
           - fluid1d-left-nutils
           - fluid3d-right-openfoam
+        run-before: ./set-case.sh 1d3d
         max_time: 0.05
         reference_result: ./water-hammer/reference-results/fluid1d-left-nutils_fluid3d-right-openfoam.tar.gz
```

Note that, in this example, we define four participants in the [`metadata.yaml`](https://github.com/precice/tutorials/blob/develop/water-hammer/metadata.yaml), but in reality, only two are needed to construct the case. We could rename the participants in `metadata.yaml` (only) to `Left` and `Right`.

While implementing this, it might be easy to also add a `run-after`, for any post-processing.

PranjalManhgaye:
## Summary

Closes #827.

Some tutorials like `water-hammer` and `partitioned-pipe-multiscale` have more than one coupling setup. We pick which one to use with `./set-case.sh 1d3d` - it switches the `precice-config.xml` symlink. System tests already copy the tutorial and run Docker, but we had no way to run that script automatically before the test.

This PR adds optional `run-before` and `run-after` in `tests.yaml`. These are shell commands that run on your machine, inside the copied tutorial folder, as part of the normal system test flow.

**What runs when:**
copy tutorial -> run-before -> max_time override -> docker build -> docker run -> run-after -> fieldcompare

I have put `run-before` before the Docker build (and before `max_time`) so all host-side file setup happens in one place. For `set-case.sh` it does not matter if we run it before build or before run - build does not read the precice config. But `max_time` has to come **after** `run-before`, because it edits whichever config file the symlink points to.

In this PR we wire `water-hammer` and `partitioned-pipe-multiscale` with `run-before: ./set-case.sh 1d3d`. `run-after` is there too, but nothing uses it yet.

## Test plan

- [x] `tests.yaml` still parses with the new fields
- [x] Smoke test: `run-before` switches the symlink, and `max_time` runs after it
- [x] `water-hammer` locally with Docker: build and run passed, saw the run-before log
- [x] Tests without hooks behave the same as before
- [x] Merged latest `develop` and addressed the review (changelog, typing, hook logging, validation)


changelog-entries/849.md
```diff
@@ -0,0 +1 @@
+- Add optional `run-before` and `run-after` hooks in `tests.yaml` for system test setup ([#827](https://github.com/precice/tutorials/pull/849)).
```
MakisH:
```suggestion
- Added optional `run-before` and `run-after` hooks in `tests.yaml` for system test setup ([#849](https://github.com/precice/tutorials/pull/849)).
```

tools/tests/systemtests/Systemtest.py
```diff
@@ -219,6 +219,8 @@ class Systemtest:
     max_time: float | None = None
     max_time_windows: int | None = None
     timeout: int = GLOBAL_TIMEOUT
+    run_before: Optional[str] = None
+    run_after: Optional[str] = None
```
MakisH: Instead of the `Optional`, you need the `|` syntax. Also in the `_run_hook`.

tools/tests/systemtests/Systemtest.py
```diff
@@ -783,11 +785,42 @@ def __apply_max_time_override(self):
                 logging.info(f"Overwrote <max-time-windows> to {self.max_time_windows} in {config_path}")
         config_path.write_text(new_text)
 
+    def _run_hook(self, stage: str, command: Optional[str]) -> bool:
+        """
+        Run a shell command in the copied tutorial directory (e.g. run-before / run-after).
+        """
```
MakisH, line 953: The overall concept needs to be explained in the PR description.

tools/tests/systemtests/Systemtest.py
```diff
@@ -783,11 +785,42 @@ def __apply_max_time_override(self):
...
+                text=True,
+                start_new_session=True,
+            )
+        except Exception as e:
+            logging.critical(f"Failed to start {stage} for {self}: {e}")
+            return False
+        if result.stdout:
+            logging.debug(f"{stage} stdout for {self}:\n{result.stdout.rstrip()}")
+        if result.stderr:
+            logging.debug(f"{stage} stderr for {self}:\n{result.stderr.rstrip()}")
```
MakisH: Why distinguishing between `stdout` and `stderr` here? What is the motivation behind this?

tools/tests/systemtests/TestSuite.py
```diff
@@ -5,6 +5,15 @@
 import yaml
 
 
+def _validate_hook_command(field_name: str, value, tutorial: Tutorial) -> Optional[str]:
+    if value is None:
+        return None
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(
+            f"{field_name} must be a non-empty string for tutorial '{tutorial}', got {value!r}")
+    return value
```
MakisH: Why is a dedicated function needed to check if `field_name` is non-empty?

Why is it called `field_name`?

Also, if it is empty, it should just do nothing, right? That would make the code simpler.

tools/tests/README.md
```diff
@@ -120,6 +120,12 @@ The available cases are listed in the `metadata.yaml` of each tutorial. To add a
 
 Use the `max_time` or `max_time_windows` parameters to restrict the runtime of the test to the first few coupling time windows, to save time. Aim for a runtime of less than a minute (assuming cached components), if possible.
 
+Some tutorials require setup before the simulation (e.g. switching configuration files). Use optional `run-before` and `run-after` fields in `tests.yaml` to run shell commands in the copied tutorial directory after copying and before Docker build (`run-before`), or after the simulation and before field comparison (`run-after`). Example:
```
MakisH, line 128: Why before the Docker build? I would expect that to be after building and before running. Is there a particular use case or complication I overlook?

For the `set-case`, it should make no difference.

MakisH (commented):
Thanks for implementing! Here are some comments before testing on GHA.

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: yess done

tools/tests/systemtests/Systemtest.py
PranjalManhgaye: No strong reason, I split them for debugging but it is not needed. I have merged them into one debug log in `_run_hook`.

tools/tests/README.md
PranjalManhgaye: For `set-case.sh` it does not matter for build vs run, but I kept `run-before` before max_time because max_time edits the active `precice config` after the symlink is set. Happy to move it after build / before run if you prefer, I left README as-is until we agree.

tools/tests/systemtests/TestSuite.py
PranjalManhgaye: Agreed, I have removed `_validate_hook_command`. If the hook is missing or empty, we just skip it now. No separate validation function.

tools/tests/README.md
MakisH: Having `run-before` before the `max_time` edit makes sense, I did not have that in mind. This looks good enough for the use case we currently have at hand. If there is any reason to run something in the containers, we can address it separately once there is a good reason to.

MakisH (approved):
Thanks for the updates! This seems to be doing what it needs to.

- Test with `1d3d`: https://github.com/precice/tutorials/actions/runs/28460562953/attempts/1
- Test with `3d1d` (fails, as it should): https://github.com/precice/tutorials/actions/runs/28460562953/attempts/2

PR #852, opened 2026-06-29
Port two-scale-heat-conduction to latest micro-manager

MakisH:
Trying to integrate the two-scale-heat-conduction tutorial into the system tests, I noticed that it does not work anymore, with the following error:

```
(0) 06/28/2026 07:08:48 PM - micro_manager.micro_manager - INFO - Micro Manager version: 0.10.1.post14+git.2bf393d1
(0) 06/28/2026 07:08:48 PM - micro_manager.micro_manager - INFO - Reading JSON configuration file: /runs/two-scale-heat-conduction_macro-dumux-micro-dumux_2026-06-28-210801/micro-dumux/micro-manager-config.json
(0) 06/28/2026 07:08:48 PM - micro_manager.micro_manager - INFO - 'micro_file_name' must be specified!
Traceback (most recent call last):
  File "/home/precice/venv/bin/micro-manager-precice", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/__init__.py", line 114, in main
    manager = MicroManagerCoupling(config_file_path, log_file=args.log_file)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/micro_manager.py", line 64, in __init__
    self._config.read_json_micro_manager()
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/config.py", line 653, in read_json_micro_manager
    self._read_json_base(self._config_file_name)
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/config.py", line 533, in _read_json_base
    file_names = self.json["micro_file_names"].get_or_raise(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/config.py", line 202, in get_or_raise
    raise e
  File "/home/precice/venv/lib/python3.12/site-packages/micro_manager/config.py", line 194, in get_or_raise
    current_element = current_element[key]
                      ~~~~~~~~~~~~~~~^^^^^
KeyError: 'micro_file_names'
```

[...]

PR #856, opened 2026-07-03
Increase timeouts for some nutils test cases

MakisH:
Triggered by https://github.com/precice/tutorials/pull/745

### DIFF
--- a/tools/tests/tests.yaml
+++ b/tools/tests/tests.yaml
@@ -29,6 +29,7 @@ test_suites:
           - fluid-nutils
           - transport-nutils
         max_time: 0.05
+        timeout: 600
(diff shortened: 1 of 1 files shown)


MakisH:
A [test](https://github.com/precice/tutorials/actions/runs/28648795643) works for all but the two-scale-heat-conduction case, which fails for a known unrelated issue (see https://github.com/precice/tutorials/pull/852, which needs a new release of the micro-manager.

The times required were:

- channel-transport_fluid-nutils_transport-nutils: 140s
- channel-transport-particles_fluid-nutils_particles-mercurydpm: 82s
- two-scale-heat-conduction: (error)

The cases from https://github.com/precice/tutorials/pull/745 were all killed at 190s, which raises questions on whether that approach increases the runtime. In any case, I want to exclude Nutils from that PR.

PR #858, opened 2026-07-05
Rework Python-based system tests components

MakisH:
#857 enabled the installation of the correct versions of the python-based components under test at the time of running the `run.sh` scripts. This now makes some of the Docker build stages obsolete (which were already incorrectly ignored).

At the same time, the FMI runner had no way to pass the Git reference under test.

This PR:

- Removes the Docker build stage `nutils_adapter`, redirecting the `nutils-adapter` component to the `python_bindings` stage. Keeping it for clarity and for ease of extension.
- Adds an `fmi-runner`. This works similarly to the `nutils-adapter`, but also accepts a new `FMI_RUNNER_REF`.
   - To make sure that the trick from #857 can correctly recognize that the `fmiprecice` package depends on `pyprecice` (without making the trick more complex), I added `pyprecice` in the respective `requirements.txt` (implied by `fmiprecice`). We already do the same for the `micro-manager-precice`, and I think this is useful (e.g., looking for tutorials that use `pyprecice` would also list these ones).
- Not directly relevant, but starting to pile up: Removes the unused and unmaintained `ubuntu_2204/Dockerfile`.

In [this test run](https://github.com/precice/tutorials/actions/runs/28756266309), the `pip-installed-packages.log` show that `pyprecice` and `fmirunner` are installed from their Git refs.

[Test run with the release test suite](https://github.com/precice/tutorials/actions/runs/28757046553)

## Checklist


MakisH:
@PranjalManhgaye what do you think of these changes? Did I forget anything? Would you wish for any other simplification in these files?

PranjalManhgaye:
@MakisH I have reviewed it locally, and it looks good to me; I guess nothing blocks this PR.
This cleanly finishes what #857 started: drop the redundant `nutils_adapter` Docker stage (runtime `requirements.txt` is what actually matters now), and give `FMI tutorials` their own `fmi-runner` component so we can pin `FMI_RUNNER_REF`. Adding pyprecice explicitly to the FMI requirements is the right fix so the prepare-step injection works for both deps.

No strong simplification wishes for this PR from my side.

PranjalManhgaye (approved):

PR #871, opened 2026-07-09
Add requirements-reference.txt

### LINKED ISSUE #610: Fix versions of Python dependencies for each distribution

PranjalManhgaye:
### Summary
This PR implements [#610](https://github.com/precice/tutorials/issues/610) using the approach discussed after [#749](https://github.com/precice/tutorials/pull/749) and it is the rework of that:

- Tutorial `requirements.txt files` keep loose version ranges `(for example numpy >1, <2, pyprecice~=3.0)`. This PR fills gaps where develop still had unpinned packages (~~39 files; ~ ~13 already had ranges).
- A new root file `requirements-reference.txt` records the exact PyPI versions that satisfy those ranges at generation time `(Gemfile.lock style)`.
- `tools/report_tutorial_requirements.py` generates and checks that file from PyPI.
- CI workflow `check-requirements-reference.yml` runs `--check` when requirements or the script change.
Tutorial `run.sh scripts` are unchanged and still install from local `requirements.txt`. The reference file is for reproducibility and releases, not for day-to-day installs.

This is not tied to `reference_versions.yaml` and does not change `tools/tests/requirements.txt` (CI git refs are already handled by #857/#858).

### Test Plan
 - python3 tools/report_tutorial_requirements.py

 - python3 tools/report_tutorial_requirements.py --check

 - pre-commit run actionlint --files .github/workflows/check-requirements-reference.yml

 - pip install --dry-run -r elastic-tube-1d/fluid-python/requirements.txt

Closes #610


requirements-reference.txt
MakisH, line 1: Interesting approach, but this will not be very visible, will not be directly usable, and will quickly become outdated.

Why not just make a `requirements-reference.txt` next to each `requirements.txt` file? This would let one just do `pip install -r requirements-reference.txt`.

README.md
```diff
@@ -13,3 +13,20 @@ As a general rule, you can start each participant from inside their `<tutorial>/
...
+```bash
+python3 tools/report_tutorial_requirements.py
+python3 tools/report_tutorial_requirements.py --check   # CI-style check
+```
+
+This reference manifest is for documentation and reproducibility; it is not used by tutorial run scripts.
+
+Legacy FEniCS packages (for example `fenics-dolfin`) are installed from the system or PPA in several tutorials and are listed in the reference file as non-PyPI dependencies.
+
+When preparing a release pull request to `master`, regenerate and commit `requirements-reference.txt` so the pinned snapshot matches the current tutorial constraints.
```
MakisH: This feels a bit disconnected here, but I cannot think of a better place for it right now. In the Python script, I suggested making a `tools/releasing/` directory. It could be moved there.

channel-transport-particles/fluid-nutils/requirements.txt
```diff
@@ -1,5 +1,5 @@
-setuptools # required by nutils
+setuptools>=69.0.0  # required by nutils
```
MakisH, line 1: Why specifically this version? No need to document, just for me to understand.

channel-transport-particles/fluid-nutils/requirements.txt
```diff
@@ -1,5 +1,5 @@
...
 nutils~=7.3
 numpy >1, <2
 pyprecice~=3.0
-setuptools
+setuptools>=69.0.0
```
MakisH: Duplicate, can be removed (here and in more files)

elastic-tube-1d/fluid-python/requirements.txt
```diff
@@ -1,3 +1,3 @@
-matplotlib
+matplotlib>=3.5,<4
```
MakisH: Why this version range?

oscillator-overlap/solver-python/requirements.txt
```diff
@@ -1,3 +1,3 @@
 numpy >1, <2
 pyprecice~=3.0
-scipy
+scipy>=1.10,<2
```
MakisH, line 3: Why this version range?

oscillator/mass-left-fmi/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pyprecice~=3.0
-fmiprecice
-pandas
\ No newline at end of file
+fmiprecice~=0.2
+pandas>=2.0,<3
```
MakisH, line 3: Why this version range?

partitioned-burgers-1d/neumann-surrogate/requirements.txt
```diff
@@ -5,5 +5,5 @@
 numpy~=2.0      # Known to work with 2.3.5
 scipy~=1.0      # Known to work with 1.16.3
 torch~=2.0      # Known to work with 2.9.1
-matplotlib      # Known to work with 3.10.8
+matplotlib~=3.0  # Known to work with 3.10.8
```
MakisH: In other places, there is a range `<4`. Why is this different?

partitioned-burgers-1d/neumann-surrogate/requirements.txt
```diff
@@ -5,5 +5,5 @@
 numpy~=2.0      # Known to work with 2.3.5
```
MakisH: How should we treat these comments now that this PR adds the `requirements-reference.txt`?

partitioned-heat-conduction-3d/solver-fenicsx/requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy >1, <2
-fenicsxprecice
-scipy
-sympy
-gmsh
+fenicsxprecice~=1.0
+scipy>=1.10,<2
+sympy>=1.10,<2
+gmsh>=4,<5
```
MakisH, line 5: Why these version ranges?

two-scale-heat-conduction/micro-nutils/requirements.txt
```diff
@@ -1,5 +1,5 @@
-setuptools # required by nutils
+setuptools>=69.0.0  # required by nutils
 nutils~=7.3
 numpy >1, <2
 pyprecice~=3.0
-micro-manager-precice
+micro-manager-precice~=0.10
```
MakisH: Actually, this can already be changed to:


```suggestion
micro-manager-precice>=0.11
```

which is still to be released. Not sure if the `~=0.11,>=0.11` would be the right way to write it, then.

water-hammer/fluid1d-left-nutils/requirements.txt
```diff
@@ -1,6 +1,6 @@
-setuptools # required by nutils
+setuptools>=69.0.0  # required by nutils
 nutils~=9.0
 numpy >1, <2
 pyprecice~=3.0
-matplotlib
-treelog
\ No newline at end of file
+matplotlib>=3.5,<4
+treelog>=1,<2
```
MakisH, line 6: Why this version range?

tools/report_tutorial_requirements.py
```diff
@@ -0,0 +1,204 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import subprocess
+import sys
+import urllib.error
+import urllib.request
+from datetime import datetime, timezone
```
MakisH: Why not use the same style?

```suggestion
from urllib import error, request
from datetime import datetime, timezone
```

tools/report_tutorial_requirements.py
```diff
@@ -0,0 +1,204 @@
...
+from packaging.specifiers import SpecifierSet
+from packaging.utils import canonicalize_name
+from packaging.version import Version
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+REFERENCE_FILE = REPO_ROOT / "requirements-reference.txt"
+EXCLUDED_RELATIVE = {
+    "tools/tests/requirements.txt",
+    # Not a pip requirements file (lists stdlib os, unpinned vtk); see develop naming quirk.
+    "flow-over-heated-plate/plot-final-interface-temperature-requirements.txt",
```
MakisH: This should actually be a `requirements.txt` file. I wanted to tidy these plotting python scripts a bit.

Useful to have an exclusion list, I guess (mainly for unmaintained tutorials), but why does this need to be excluded?

tools/releasing/update-requirements-reference.py
MakisH, line 1: Let's move this into a new directory `tools/releasing/`. The documentation (from the root `README.md`) can also go there then.

MakisH (commented):
Thank you for the contribution; this will be useful, and it comes at the right time.

My main suggestion, which will change a bit the code, is to split the `requirements-reference.txt` into one file per `requirements.txt`.

Other than that, most comments are regarding the version ranges. We do not need to document these in detail, but the comments here might be helpful in the future.

requirements-reference.txt
PranjalManhgaye: Done, each requirements.txt now has a sibling requirements-reference.txt, so you can `pip install -r requirements-reference.txt` in that folder. The root file is removed.

README.md
PranjalManhgaye: Docs moved to `tools/releasing/README.md`, with a short pointer from the root README.

channel-transport-particles/fluid-nutils/requirements.txt
PranjalManhgaye: Kept as a lower bound that works with current nutils; happy to loosen further if you prefer.

channel-transport-particles/fluid-nutils/requirements.txt
PranjalManhgaye: Yeah, I removed the duplicate lines.

elastic-tube-1d/fluid-python/requirements.txt
PranjalManhgaye: Aimed to stay on matplotlib 3.x. Unified to matplotlib>=3, <4 everywhere for consistency.

oscillator-overlap/solver-python/requirements.txt
PranjalManhgaye: Stay on scipy 1.x with a recent-enough floor; exact pin is in the sibling reference file.

oscillator/mass-left-fmi/requirements.txt
PranjalManhgaye: Stay on pandas 2.x; the exact pin is in the sibling reference file.

partitioned-burgers-1d/neumann-surrogate/requirements.txt
PranjalManhgaye: Done, aligned all matplotlib constraints to matplotlib >= 3, < 4.

partitioned-burgers-1d/neumann-surrogate/requirements.txt
PranjalManhgaye: Removed them. Exact versions now live in the sibling `requirements-reference.txt`.

partitioned-heat-conduction-3d/solver-fenicsx/requirements.txt
PranjalManhgaye: Loose major-compatible ranges for installs; exact versions are in the sibling reference file.

water-hammer/fluid1d-left-nutils/requirements.txt
PranjalManhgaye: Keep the treelog on 1.x; the exact pin is in the sibling reference file.

tools/report_tutorial_requirements.py
PranjalManhgaye: Done, moved to `plot-final-interface-temperature/requirements.txt` (dropped invalid OS), and removed the special exclusion.

tools/releasing/update-requirements-reference.py
PranjalManhgaye: Done, script and docs are under `tools/releasing/`.

tools/releasing/update-requirements-reference.py
```diff
@@ -0,0 +1,228 @@
...
+import subprocess
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from urllib import error, request
+
+from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
+from packaging.utils import canonicalize_name
+from packaging.version import Version
```
MakisH, line 30: It looks like we also need a `requirements.txt` for this tool.

tools/releasing/report_tutorial_requirements.py
```diff
@@ -0,0 +1,228 @@
...
+"""
+Report resolved Python dependency versions for tutorial requirements.txt files.
+
+Scans every tutorial requirements.txt (excluding tools/tests/), queries PyPI for
+the latest version satisfying each constraint, and writes a sibling
+requirements-reference.txt next to each file.
+
+Run from the repository root:
+  python3 tools/releasing/report_tutorial_requirements.py
+  python3 tools/releasing/report_tutorial_requirements.py --check
```
MakisH: Some suggestions for the interface:

- I would name it `update-requirements-reference.py`: dashes for consistency with most of our tools (yes, we can improve that), clearly point to these files.
- It would be nice to accept a path to only update files in that directory. In the PR template, you added a step to do that, but one can only update all files at once.
- By default, all files will be updated, because the date/time is always different. This makes it difficult to see which of them actually changed in terms of versions. An option `--all` would be mostly intuitive to me (any better ideas?).

tools/releasing/report_tutorial_requirements.py
```diff
@@ -0,0 +1,228 @@
...
+        if outdated:
+            print("ERROR: outdated requirements-reference.txt files:", file=sys.stderr)
+            for path in outdated:
+                print(f"  - {path}", file=sys.stderr)
+        print(
+            "Run: python3 tools/releasing/report_tutorial_requirements.py",
+            file=sys.stderr,
+        )
+        return 1
+
```
MakisH: I think that the outdated ones should only give a warning, not an error. We should only update them when making a new release.

.github/workflows/check-requirements-reference.yml
```diff
@@ -0,0 +1,32 @@
...
+      - '**/requirements-reference.txt'
+      - tools/releasing/report_tutorial_requirements.py
+  pull_request:
+    branches:
+      - master
+      - develop
+    paths:
+      - '**/requirements.txt'
+      - '**/requirements-reference.txt'
+      - tools/releasing/report_tutorial_requirements.py
```
MakisH: I think that we should check:

- In a PR to `develop`: That there are `requirements-reference.txt` files: The first time a tutorial is added, it would be good to archive that information.
- In a PR to `master` (i.e., a release): That the `requirements-reference.txt` files are up to date: This is the time I would expect them to be updated. 

I would say that the `push` event is not needed, then.

tools/report_tutorial_requirements.py
MakisH: Good point, but the script itself should also be in that directory then.

You can also make a directory `utils/`: https://precice.org/community-contribute-to-precice.html#structure-of-a-tutorial

tools/releasing/report_tutorial_requirements.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python3
+"""
+Report resolved Python dependency versions for tutorial requirements.txt files.
```
MakisH: Not sure if you only mean the files corresponding to tutorials cases, but I think we should cover every `requirements.txt` file in this repository. There is a list with exceptions, which is nice, but I think it should be empty by default, unless there is a good reason not to.

tools/releasing/update-requirements-reference.py
```diff
@@ -0,0 +1,228 @@
...
+    return sorted(files)
+
+
+def reference_path_for(requirements_file: Path) -> Path:
+    return requirements_file.with_name(REFERENCE_NAME)
+
+
+def parse_requirement_line(line: str) -> str | None:
+    stripped = line.strip()
+    if not stripped or stripped.startswith("#") or stripped.startswith("--"):
```
MakisH, line 88: What is the `--` for?

MakisH (commented):
The version ranges look fine. The system tests pass: https://github.com/precice/tutorials/actions/runs/29193941172?pr=871

I had a closer look at the script, here are some more comments.

tools/report_tutorial_requirements.py
PranjalManhgaye: Done, I have moved `plot-final-interface-temperature.py` into `plot-final-interface-temperature/` next to its `requirements.txt`, and updated the tutorial README. Paths resolve via the tutorial root so exports under `solid-*/` still work. Kept this folder instead of adding `utils/` since the requirements already lived there.

tools/releasing/update-requirements-reference.py
PranjalManhgaye: Yeah I guess we need it so I Added `tools/releasing/requirements.txt (packaging)`. CI now installs with `pip install -r tools/releasing/requirements.txt`, and there is a sibling `requirements-reference.txt` as well.

tools/releasing/report_tutorial_requirements.py
PranjalManhgaye: I have renamed it to `update-requirements-reference.py`. now You can pass a path to update just that directory. By default I only rewrite files when the pins actually change; so we now use `--all` in latest one commit if you want a full refresh, yes its intuitive to me as well.

tools/releasing/report_tutorial_requirements.py
PranjalManhgaye: Yeah, `--check` now only errors on missing files; outdated ones are just a warning. For release PRs to master we still fail with `--fail-on-outdated`.

.github/workflows/check-requirements-reference.yml
PranjalManhgaye: Okay, i have dropped it

tools/releasing/report_tutorial_requirements.py
PranjalManhgaye: Thanks, Exception list is empty now. Every tracked `requirements.txt` gets a sibling reference, including `tools/tests/ and tools/releasing/`.

tools/releasing/update-requirements-reference.py
PranjalManhgaye: It is for pip options like `--extra-index-url`, those aren’t packages, so we skip them when resolving versions (but still copy them into the reference file).

MakisH (approved):
Thanks for all the updates! I like the new interface, I think it is quite clear and flexible.

The script does what it should, but I have not reviewed it in much more detail than that.

Note that, while this solves #610 for the direct dependencies, but does not provide a full `pip freeze` output. Our tutorials, however, do generate `pip-installed-packages.log`, which are ignored by default, but could be archived together with any case files and logs. I think this is enough of a solution for now.

PranjalManhgaye:
Thanks, Glad the interface worked out , Happy to look at archiving `pip-installed-packages.log` later if that becomes useful.

PR #879, opened 2026-07-15
Simplify the openfoam_remove_empty_dirs.sh script

MakisH:
The [current script](https://github.com/precice/tutorials/blob/e9dd93a5c328c56bee342a394587b97abfd3b8bb/tools/openfoam-remove-empty-dirs.sh) is a bit verbose and has some duplication. It also does not handle the case of collated output, as discussed in https://github.com/precice/tutorials/pull/834#discussion_r3582926081

This PR rewrites the script in a more general way. It defines a list of directories to check (`.` and `processor*`) and a list of common result files (so this list can also easily be extended in one place), and then:

```
for each ./ processor*/
  for each numbered directory
    for each expected result
      if a `result` or `result.gz` exists, keep the directory
    if not marked to be kept, delete it
```

It also improves logging, to make more transparent what is happening. I am not sure if having a list of what was deleted is useful: I already removed this in the past, as the logs were getting too long.

At some point I switched this from standard `sh` to `bash`, as our `run.sh` scripts. Not sure it there is any reason to keep it with `sh` anymore. At least macOS users have not complained so far.

@mike-tree-corvid what do you think? Does this cover your use case?

## Checklist


MakisH:
I assume that there is no strong objection to this. Merging to start testing it in practice.

PR #882, opened 2026-07-17
Move reference results metadata into each archive

### LINKED ISSUE #811: reference_results.metadata is outside the respective archive
In continuation of https://github.com/precice/tutorials/issues/537 (solved by https://github.com/precice/tutorials/pull/721), we now have the following structure:

```
elastic-tube-1d/
  ...
  reference-results/
    fluid-cpp_solid-cpp.tar.gz        -> in LFS
    fluid-cpp_solid-python.tar.gz     -> in LFS
    fluid-python_solid-python.tar.gz  -> in LFS
    reference_results.metadata        -> text, directly in the repository
```

Since we now can generate reference results for individual test suites (added by https://github.com/precice/tutorials/pull/728), the `reference_results.metadata` is updated by the latest run, without necessarily corresponding to all of the archives in this directory.

At the same time, the text is a bit useless without the archives (in LFS), adds noise to the Git history, and often pops up in code searches.

We should break this file into one per archive, and put it into the respective archive, ending up with:

```
elastic-tube-1d/
  ...
  reference-results/
    fluid-cpp_solid-cpp.tar.gz        -> in LFS, including a reference_results.metadata
    fluid-cpp_solid-python.tar.gz     -> in LFS, including a reference_results.metadata
    fluid-python_solid-python.tar.gz  -> in LFS, including a reference_results.metadata
```

PranjalManhgaye:
## Summary

- Store `reference-results-metadata.txt` inside each archive's case directory

- Stop writing one shared metadata file per tutorial
- Drop the SHA256 checksum from the metadata (it described the outer `.tar.gz`, which no longer makes sense once the file lives inside that archive)
- Ignore the embedded metadata during field comparison
- Rename the metadata file to `reference-results-metadata.txt` so editors open it as plain text
- Update the system-test documentation

Closes #811

### Test Plan

Packed sample archives locally and verified `reference-results-metadata.txt` ends up inside each case folder, with exports and iteration logs unchanged, and no shared metadata file written anymore. Also verified fieldcompare ignores the metadata file, and ran pre-commit on the changed files. Full regeneration of reference archives happens after merge, as discussed.


tools/tests/docker-compose.field_compare.template.yaml
```diff
@@ -6,4 +6,4 @@ services:
     command:
       - /runs/{{ tutorial_folder }}/{{ precice_output_folder }}
       - /runs/{{ tutorial_folder }}/{{ reference_output_folder }}
-      - "-rtol {{ tolerance }} --ignore-missing-reference-files --diff"
+      - "-rtol {{ tolerance }} --ignore-missing-reference-files --ignore-unsupported-file-formats --diff"
```
MakisH: I thought we already had this - good catch!

tools/tests/reference-results-metadata.txt.template
```diff
@@ -4,17 +4,16 @@ This File has been generated by the generate_reference_results.py and should not
...
 
-| name | time | sha256 |
-|------|------|-------|
-{% for file in files -%}
-| {{ file.name }} | {{ file.time }} | {{ file.sha256 }} |
-{% endfor %}
-## List of arguments used to generate the files
+| name | generated at |
+|------|--------------|
+| {{ archive_name }} | {{ time }} |
```
MakisH, line 14: The PR description should also mention that the checksum had to be removed (it was not immediately clear to me only by the diff).

tools/tests/generate_reference_results.py
MakisH, line 1: Something looks wrong: In [this run](https://github.com/precice/tutorials/actions/runs/29635782053), I see the updated template in the `tools/` directory, but I cannot find the updated `reference-results.metadata` in it.

While we are on it, let's rename the file to `reference-results-metadata.txt`, to make it easier to open it with editors.

tools/tests/docker-compose.field_compare.template.yaml
PranjalManhgaye: Thanks, While re-checking this with a real `fieldcompare install`, I found that `--ignore-unsupported-file-formats` alone is not enough for the embedded metadata: fieldcompare fails earlier with "Missing source file" because the metadata exists only in the unpacked reference directory. 

I added `--ignore-missing-source-files` as well. Locally verified that this skips the metadata file while still catching real field regressions.

tools/tests/reference-results-metadata.txt.template
PranjalManhgaye: done, updated the description.

tools/tests/generate_reference_results.py
PranjalManhgaye: Thanks for testing the generation run. I checked the regenerated archive from that run: the updated metadata is already inside the case folder of the `.tar.gz` (new format, dated 2026-07-18). 

The shared `reference-results/reference_results.metadata` outside the archive is a leftover from before this PR and was not refreshed because we stopped writing it. 

Also renamed the embedded file to `reference-results-metadata.txt` (and the template accordingly), as suggested.

tools/tests/docker-compose.field_compare.template.yaml
MakisH: Actually, wait: Why do we need fieldcompare while generating the reference results?

PranjalManhgaye: yess ofc fieldcompare is not used during generation, It's needed later, when system tests compare against the new archives: the metadata is only in the unpacked reference dir, so fieldcompare would fail with "Missing source file" without that flag.

MakisH: Right, I forgot about it. However, I am afraid that the combination `--ignore-missing-reference-files --ignore-missing-source-files` will report false success in cases where the solver generated results at different times.

Could we instead switch fieldcompare to only check specific file extensions? The `fieldcompare --file-type` might be the right option to use.

tools/tests/generate_reference_results.py
MakisH: In the archive `quickstart_fluid-openfoam-solid-cpp_2026-07-18-092817/reference-results/fluid-openfoam_solid-cpp.tar.gz`, I do not see any metadata file. Should I look elsewhere?

[fluid-openfoam_solid-cpp.tar.gz](https://github.com/user-attachments/files/30148826/fluid-openfoam_solid-cpp.tar.gz)

MakisH: Clarified - I was looking at the wrong place, all good. Thanks!

tools/tests/docker-compose.field_compare.template.yaml
PranjalManhgaye: please check i have pushed an update.

MakisH (approved):
Looks good. I would further simplify and polish the template, but I can directly do that in a separate PR.

MakisH:
The system tests are now failing, complaining about the fieldcompare option:

```
#8 resolving provenance for metadata file
#8 DONE 0.0s
[stderr]  Image breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-field-compare Built 
[stderr] time="2026-08-10T03:51:43+02:00" level=warning msg="Found orphan containers (breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-solid-calculix-1, breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-fluid-openfoam-1, breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-prepare-1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up."
[stderr]  Container breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-field-compare-1 Creating 
[stderr]  Container breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-field-compare-1 Created 
Attaching to field-compare-1
[stderr]  Container breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-field-compare-1 Starting 
[stderr]  Container breaking-dam-2d_fluid-openfoam-solid-calculix_2026-08-10-035105-field-compare-1 Started 
[stderr] field-compare-1  | usage: fieldcompare [-h] [--version] {file,dir} ...
[stderr] field-compare-1  | fieldcompare: error: unrecognized arguments: --exclude-files reference-results-metadata.txt --ignore-unsupported-file-formats
```

I did not expect this to fail, but it looks like the fieldcompare version we use is too old: https://github.com/dglaeser/fieldcompare-action

I will contribute upstream. -> https://github.com/dglaeser/fieldcompare-action/pull/15

edit: Since this is anyway trivial, I dropped the dependency in https://github.com/precice/tutorials/pull/895

PR #883, opened 2026-07-17
Render fieldcompare diff visualizations

### LINKED ISSUE #441: Archive diff files in the reports

PranjalManhgaye:
## Summary

- Render archived fieldcompare diff VTK files as PNG images when result comparison fails.

- Generate one image per numeric point field using distinct sphere glyphs and a generic automatically selected view.
- Store images under `diff-results/visualizations/` in the system-test artifacts.
- Keep visualization best-effort so rendering errors do not hide the original fieldcompare failure.

Fixes #441.

## Test plan

- Generated PNGs from synthetic scalar and vector VTK fields.

- Verified the `archive-to-visualization` failure path locally.
- Rendered five real `CI diff VTK files`, producing 13 PNGs.
- Checked zero and non-zero differences visually.
- Ran `pre-commit` checks successfully.


changelog-entries/441.md
```diff
@@ -1 +1 @@
-- Archive fieldcompare diff VTK files into a `diff-results/` folder in each systemtest run directory on failure so they are easy to find in CI artifacts when investigating comparison failures (fixes [#441](https://github.com/precice/tutorials/issues/441)). Nested paths under `precice-exports/` are preserved under `diff-results/`.
+- Archive fieldcompare diff VTK files into a `diff-results/` folder in each systemtest run directory on failure so they are easy to find in CI artifacts when investigating comparison failures (fixes [#441](https://github.com/precice/tutorials/issues/441)). Nested paths under `precice-exports/` are preserved under `diff-results/`, and each numeric point field is rendered headlessly as a sphere-glyph PNG in `diff-results/visualizations/`.
```
MakisH: Links to both PRs should be mentioned here. It looks like this previously slipped.

tools/tests/requirements.txt
MakisH, line 1: It looks like we need more dependencies. [Running on GHA](https://github.com/precice/tutorials/actions/runs/29634822859/job/88055204430), I get:

```
2026-07-18 08:56:07.732 (   2.224s) [    7FE8965761C0]vtkOpenGLRenderWindow.c:645   WARN| Failed to load EGL! Please install the EGL library from your distribution's package manager.
2026-07-18 08:56:07.732 (   2.225s) [    7FE8965761C0]vtkOSOpenGLRenderWindow:149   WARN| libOSMesa not found. It appears that OSMesa is not installed in your system. Please install the OSMesa library from your distribution's package manager.
2026-07-18 08:56:07.734 (   2.226s) [    7FE8965761C0]vtkOpenGLRenderWindow.c:645   WARN| Failed to load EGL! Please install the EGL library from your distribution's package manager.
2026-07-18 08:56:07.734 (   2.226s) [    7FE8965761C0]vtkOSOpenGLRenderWindow:149   WARN| libOSMesa not found. It appears that OSMesa is not installed in your system. Please install the OSMesa library from your distribution's package manager.
CRITICAL: Fieldcompare returned non zero exit code, therefore Channel transport (fluid-openfoam, transport-openfoam) failed
```

tools/tests/visualize_fieldcompare_diffs.py
MakisH, line 1: The respective images could also be added to the `_logs` build artifact in the GHA workflow.

Even better would be if we could add them on the job summary. I guess that images on job summaries are supported, but I am not sure where to source the image from.

MakisH (commented):
Thank you for the PR! Some first comments, I will continue once I can run it.

changelog-entries/441.md
PranjalManhgaye: I’ve updated `441.md `and added links to both related PRs (#740 and #883).

tools/tests/requirements.txt
PranjalManhgaye: Thanks for sharing the logs here, now I have moved the rendering path to Docker and added `OSMesa` there `(libosmesa6 + diff_visualizer stage)`, so local and GHA now use the same headless setup.

tools/tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: Thanks now I have added the PNG visualizations to the` _logs artifact` and also added a `small job-summary link` so they’re easier to find.

changelog-entries/441.md
```diff
@@ -1 +1 @@
-- Archive fieldcompare diff VTK files into a `diff-results/` folder in each systemtest run directory on failure so they are easy to find in CI artifacts when investigating comparison failures (fixes [#441](https://github.com/precice/tutorials/issues/441)). Nested paths under `precice-exports/` are preserved under `diff-results/`.
+- Archive fieldcompare diff VTK files into a `diff-results/` folder in each systemtest run directory on failure so they are easy to find in CI artifacts when investigating comparison failures (fixes [#441](https://github.com/precice/tutorials/issues/441), [#740](https://github.com/precice/tutorials/pull/740), [#883](https://github.com/precice/tutorials/pull/883)). Nested paths under `precice-exports/` are preserved under `diff-results/`, and each numeric point field is rendered headlessly as a sphere-glyph PNG in `diff-results/visualizations/`.
```
MakisH: I find changelog entries with more than one sentence too verbose, especially if they list implementation details:


```suggestion
- Archive fieldcompare diff VTK files into a `diff-results/` folder in each systemtest run directory on failure so they are easy to find in CI artifacts when investigating comparison failures (fixes [#441](https://github.com/precice/tutorials/issues/441), [#740](https://github.com/precice/tutorials/pull/740), [#883](https://github.com/precice/tutorials/pull/883)).
```

tools/tests/README.md
```diff
@@ -99,7 +99,7 @@ When the tests fail at the results comparison step, this typically means that th
 
 - `precice-exports/`: The coupling meshes of the test run.
 - `reference-results/`: The coupling meshes of the reference run, as stored on Git LFS, expanded into `reference-results-unpacked`. For test cases using implicit coupling, the reference `.tar.gz` also contains the reference `precice-*-iterations.log` files.
-- `diff-results/`: Numerical difference of the results in the two directories (computed with `fieldcompare dir --diff precice-exports/ reference/`). These are only present on failed comparisons.
+- `diff-results/`: Numerical difference of the results in the two directories (computed with `fieldcompare dir --diff precice-exports/ reference/`). These are only present on failed comparisons. The `visualizations/` subdirectory contains one PNG per numeric point field, rendered as sphere glyphs and colored by the difference values (via the `diff_visualizer` Docker stage using OSMesa, so no display server is required).
```
MakisH: I would avoid implementation details in the user documentation. These are meant either for code comments or for the PR description.

```suggestion
- `diff-results/`: Numerical difference of the results in the two directories (computed with `fieldcompare dir --diff precice-exports/ reference/`). These are only present on failed comparisons and accompanied by visualization in `diff-results/visualization/`.
```

tools/tests/README.md
```diff
@@ -123,7 +123,7 @@ When a system test fails in CI, download the **full** artifact:
 
 `system_tests_run_<run_id>_<run_attempt>_full`
 
-(a smaller `_logs` archive contains only log files). The archive contains a shared `runs/` directory:
+(a smaller `_logs` archive contains the stage log files and, on comparison failures, the PNG diff visualizations under `runs/*/diff-results/visualizations/`). The archive contains a shared `runs/` directory:
```
MakisH: It is clear where to find these once one looks at the archive:

```suggestion
(a smaller `_logs` archive contains the stage log files and, on comparison failures, difference visualizations. The archive contains a shared `runs/` directory:
```

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,204 @@
...
+        if path.is_file()
+        and path.suffix.lower() in SUPPORTED_SUFFIXES
+        and "diff" in path.name.lower()
+    )
+
+
+def _safe_name(value: str) -> str:
+    """Return a filesystem-safe part of an output filename."""
+    cleaned = re.sub(r"[^\w.-]+", "_", value, flags=re.UNICODE)
+    return cleaned.strip("_.") or "unnamed"
```
MakisH: I would generally not implement these as separate functions; instead, I would directly define them in place. Especially the `discover_diff_files` is only used once and is quite straight-forward.

tools/tests/dockerfiles/ubuntu_2404/Dockerfile
```diff
@@ -60,7 +60,8 @@ RUN apt-get -qq update && \
         python3-venv \
         pkg-config \
         wget \
-        inotify-tools
+        inotify-tools \
+        libosmesa6
```
MakisH, line 64: While I could (iirc) previously test this, trying to test it now leads me to extremely slow APT downloads. After 40min, APT is still fetching packages: https://github.com/precice/tutorials/actions/runs/30921073577

If this persists tomorrow / in the next days, we would probably need to find alternatives. I don't know if this is only in Docker; if that's the case, it might make sense to migrate the rendering to the host.

Did you face any issues in this step?

tools/tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: done, Inlined `discover_diff_files` and `_safe_name` where they’re used.

tools/tests/dockerfiles/ubuntu_2404/Dockerfile
PranjalManhgaye: I have checked that run too, it was still stuck fetching apt packages during the Docker build, so it never reached `fieldcompare/visualization`.

I didn’t hit this as badly in earlier local Docker tests as its here, but ig if apt stays this slow on the runners, I am fine moving rendering to the host as a fallback. and also to retry the run or prepare that change if you prefer.

MakisH: Worked now: https://github.com/precice/tutorials/actions/runs/30982129537

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,202 @@
...
+        plotter.set_background("white")
+        max_abs_value = float(np.max(np.abs(values)))
+        color_limit = max_abs_value if max_abs_value > 0 else 1.0
+        plotter.add_mesh(
+            glyphs,
+            scalars=scalar_name,
+            cmap="coolwarm",
+            clim=(-color_limit, color_limit),
+            scalar_bar_args={"title": field_name},
+        )
```
MakisH, line 107: In [this run](https://github.com/precice/tutorials/actions/runs/30982129537), fieldcompare complained about the force in the last time window:

```
field-compare-1  | Comparing '/runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-05-084014/precice-exports/Solid-Mesh-Solid.dt500.vtu'
field-compare-1  |       and '/runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-05-084014/reference-results-unpacked/fluid-openfoam_solid-calculix/Solid-Mesh-Solid.dt500.vtu'
field-compare-1  |   Reading '/runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-05-084014/precice-exports/Solid-Mesh-Solid.dt500.vtu'
field-compare-1  |   Reading '/runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-05-084014/reference-results-unpacked/fluid-openfoam_solid-calculix/Solid-Mesh-Solid.dt500.vtu'
field-compare-1  |    -- Comparing the field 'Force': FAILED
field-compare-1  |      -- Report: Deviation above tolerance detected -> [ 8.05858e-03 -5.47671e-05  0.00000e+00] vs. [ 8.05464e-03 -5.46938e-05  0.00000e+00] ([ 0.05 -0.13   inf] %)
field-compare-1  |      -- Predicate: DefaultEquality (abs_tol: 0., rel_tol: 3.e-07)
field-compare-1  |   Wrote diff into '/runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-05-084014/precice-exports/diff_Solid-Mesh-Solid.dt500.vtu.vtu'
field-compare-1  |   File comparison FAILED with 2 PASSED / 1 FAILED / 0 SKIPPED
```

but looking at the image, this difference is not clearly visible:

Image

The `[-1.0,1.0]` range is a bit arbitrary. Why is that being picked? I would generally rely on the color scale and range to understand how large a regression is.

Also, if the difference is visualized (and not the absolute value of it), why have the range symmetric? There must be a minimum value and a maximum value.

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,202 @@
...
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+
+import numpy as np
+import pyvista as pv
+
+
+SUPPORTED_SUFFIXES = {".vtk", ".vtp", ".vtu"}
+WINDOW_SIZE = (1024, 768)
```
MakisH, line 17: In some cases, this resolution is a bit small. See [a run for the heat exchanger](https://github.com/precice/tutorials/actions/runs/30984250804):

Image

Can pyvista export PDF files (vector graphics)? If we cannot include PNG files in the GHA summary, then PDF files would also do, and maybe they would make the archive smaller.

.github/workflows/system-tests.yml
```diff
@@ -91,6 +92,18 @@ jobs:
           runs/*/system-tests-run.log
           runs/*/system-tests-compare.log
           runs/*/*/system-tests_*.log
+          runs/*/diff-results/visualizations/**/*.png
```
MakisH, line 95: It now looks like the `_logs` archive is several MB large, which defeats the original purpose of having something small to download for inspection on a bad connection.

Let's move these to a new archive (`_diffs`).

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,202 @@
...
+            scalars=scalar_name,
+            cmap="coolwarm",
+            clim=(-color_limit, color_limit),
+            scalar_bar_args={"title": field_name},
+        )
+        plotter.add_text(
+            f"{source_file.name}\npoint field: {field_name}",
+            font_size=10,
+            color="black",
+        )
```
MakisH, line 112: An additional useful text here would be what the exact field operation is: `Difference: computed - reference` or `Difference: reference - computed`?

tools/tests/visualize_fieldcompare_diffs.py
MakisH: Notes from our discussion:

- We should show the range of the diff files
- Even if the values tend to zero, if that doesn't cause any issues, let's have the real values: seeing a range `-1.23e-33 to +4.56e-33` is also clear, and it is how ParaView does it
- We noticed that the visualizations for the `Solid` participant are missing from the picture.

PranjalManhgaye: I have updated the color scale to use the real `min/max` from the diff field `(clim=(vmin, vmax))`, including near-zero ranges, and show that range in the overlay.
On the Solid side: I re-checked locally with real CI diffs and synthetic Solid/Force cases, Solid PNGs are produced. Some look almost uniform when the diff is `~0 (e.g. range 0 to 0)`; that is a real zero-difference field.

tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,202 @@
...
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+
+import numpy as np
+import pyvista as pv
+
+
+SUPPORTED_SUFFIXES = {".vtk", ".vtp", ".vtu"}
+WINDOW_SIZE = (1024, 768)
```
PranjalManhgaye: Kept pngs as per our discussion

.github/workflows/system-tests.yml
PranjalManhgaye: Done, visualizations now go to a separate` _diffs artifact`; `_logs` stays logs-only. The job summary link points to `_diffs`.

tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,202 @@
...
+            scalars=scalar_name,
+            cmap="coolwarm",
+            clim=(-color_limit, color_limit),
+            scalar_bar_args={"title": field_name},
+        )
+        plotter.add_text(
+            f"{source_file.name}\npoint field: {field_name}",
+            font_size=10,
+            color="black",
+        )
```
PranjalManhgaye: Added to the overlay: difference: `computed - reference`, plus the actual value range.

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,207 @@
...
+        and path.suffix.lower() in SUPPORTED_SUFFIXES
+        and "diff" in path.name.lower()
+    )
+    for diff_file in diff_files:
+        try:
+            generated.extend(
+                visualize_diff_file(diff_file, diff_results_dir, output_dir)
+            )
+        except Exception as error:
+            errors.append(f"Could not visualize {diff_file}: {error}")
```
MakisH: In https://github.com/precice/tutorials/actions/runs/31705035580, I still don't see the images for the `Solid-Mesh-Solid` files. Since there are no logs, I cannot see if these files triggered this exception.

We need a log file, probably called `system-tests-compare-diff.log`, next to `system-tests-compare.log`.

tools/tests/visualize_fieldcompare_diffs.py
MakisH: Still, fieldcompare reports:

```
field-compare-1  | [ FAILED  ] runs/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-13-152820/precice-exports/Solid-Mesh-Fluid.dt500.vtu: ('Force')
```

but the archive does not include the respective files, so I cannot yet check if this was fixed.

tests/README.md
```diff
@@ -118,13 +118,23 @@ The differences are only shown per file, and there is no global metric or other
 
 Alternatively, [visualize the `precice-exports/diff_*.vtu` in ParaView](https://precice.org/configuration-export.html#visualization-with-paraview).
 
+To regenerate the PNG visualizations locally from an archived `diff-results/` folder (for example after downloading a CI artifact):
+
+```bash
+python3 visualize_fieldcompare_diffs.py /path/to/diff-results
```
MakisH, line 127: I am running locally:

```
python3 visualize_fieldcompare_diffs.py /home/gc/Downloads/system_tests_run_31705035580_1_full/perpendicular-flap_fluid-openfoam-solid-calculix_2026-08-13-152820/diff-results
```

but this seems to be stuck without any output for a while. Some output in the script would be good. I assume something is wrong; it cannot be that slow.

MakisH: Actually, it is very slow. It took at least 30min on my system. I guess the GHA runners are faster.

There, the job timed out, I didn't notice before:

```
WARNING: Could not render fieldcompare diff visualizations for Perpendicular flap (fluid-openfoam, solid-calculix): Command '['docker', 'compose', '--file', 'docker-compose.diff_visualizer.yaml', 'up', '--exit-code-from', 'diff-visualizer', '--abort-on-container-exit']' timed out after 300 seconds
CRITICAL: Fieldcompare returned non zero exit code, therefore Perpendicular flap (fluid-openfoam, solid-calculix) failed
```

MakisH: Notice also that only `0m 49s` is reported as `compare time`.

So, overall, the logging could be improved.

tools/tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,207 @@
...
+    generated: list[Path] = []
+    errors: list[str] = []
+    diff_files = sorted(
+        path
+        for path in diff_results_dir.rglob("*")
+        if path.is_file()
+        and path.suffix.lower() in SUPPORTED_SUFFIXES
+        and "diff" in path.name.lower()
+    )
+    for diff_file in diff_files:
```
MakisH: This loop could be parallelized using the number of available threads.

tools/tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: Fixed in `8ee8a5ea`. Root cause was the 300s timeout while glyphs were still on Fluid files. We now use point sprites on dense meshes, parallel workers, 900s timeout, and `system-tests-compare-diff.log`. On your flap diff-results locally: `~2 min, 4509 PNGs`, `Solid-Mesh-Solid Force` including dt500.

tools/tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: the visualizer was killed mid-run before Solid finished. Re-ran on the same flap artifact, `Solid-Mesh-Fluid` and Solid-Mesh-Solid Force PNGs are produced now .

tests/README.md
PranjalManhgaye: Added `[i/n] Rendering  /  Wrote` with flush, and `PYTHONUNBUFFERED=1` in the compose template. Progress shows up immediately in the terminal and in `system-tests-compare-diff.log`.

tools/tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: Done with `ProcessPoolExecutor` (up to 8 workers, spawn). Each worker logs progress per file.

.github/workflows/system-tests.yml
```diff
@@ -83,14 +83,36 @@ jobs:
...
+        if-no-files-found: warn
+    - name: Archive fieldcompare diff visualizations
+      if: ${{ always() }}
+      id: archive-diffs
+      uses: actions/upload-artifact@v7
+      with:
+        name: system_tests_run_${{ github.run_id }}_${{ github.run_attempt }}_diffs
+        path: |
+          runs/*/diff-results/visualizations/**/*.png
+        if-no-files-found: warn
```
MakisH: This step is running always (`if: ${{ always() }}`), but the comparison step is not always expected to fail (to produce diffs). This means that `if-no-files-found: warn` would trigger a warning also when there are no diffs because there are no regressions:

```suggestion
```

Same for the previous step, but there all other logs are expected to always be present.

tests/systemtests/Systemtest.py
```diff
@@ -214,7 +218,7 @@ def _get_length_of_name(results: List[SystemtestResult]) -> int:
         with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
             print("\n\n", file=f)
             print(
-                "In case a test fails, download the archive from the bottom of this page and inspect the per-stage logs (`system-tests-build.log`, `system-tests-run.log`, `system-tests-compare.log`). The stage runtimes might already give useful hints.",
+                "In case a test fails, download the archive from the bottom of this page and inspect the per-stage logs (`system-tests-build.log`, `system-tests-run.log`, `system-tests-compare.log`, and `system-tests-compare-diff.log` when visualizations ran). The stage runtimes might already give useful hints.",
```
MakisH:
```suggestion
                "In case a test fails, download the archive from the bottom of this page and inspect the per-stage logs (`system-tests-build.log`, `system-tests-run.log`, `system-tests-compare.log`, and `system-tests-compare-diff.log`). The stage runtimes might already give useful hints.",
```

tests/README.md
```diff
@@ -120,13 +121,25 @@ The differences are only shown per file, and there is no global metric or other
...
+```
+
+The script prints progress per file (`[3/40] Rendering ...`) and uses several worker processes. Sparse meshes keep sphere glyphs; denser clouds use point sprites so large cases finish in minutes rather than tens of minutes.
```
MakisH:
```suggestion
The script prints progress per file, and processes files in parallel. Sparse meshes keep sphere glyphs; denser clouds use point sprites to save time.
```

MakisH: But why not just use point sprites for all? It would keep the code simpler.
Which cases currently use point sprites?

tests/visualize_fieldcompare_diffs.py
```diff
@@ -0,0 +1,298 @@
...
+    else:
+        mesh = point_cloud
+        mesh_kwargs = {
+            "render_points_as_spheres": True,
+            "point_size": _point_size(len(points)),
+        }
+
+    output_file.parent.mkdir(parents=True, exist_ok=True)
+    vmin = float(np.min(values))
+    vmax = float(np.max(values))
```
MakisH, line 93: If the diff is zero, one could skip rendering/writing the file. This would make discovering the files with regressions easier, it would make the archives smaller, and the generation faster.

MakisH (commented):
Nice updates, almost there! For the perpendicular-flap, generating the visualizations now took ~2min on my laptop.

Main comments:

1. why not just use point sprites everywhere?
2. skip writing files would show zero difference

tests/visualize_fieldcompare_diffs.py
PranjalManhgaye: done

.github/workflows/system-tests.yml
PranjalManhgaye: Thanks, done

MakisH (approved):
Everything looks good now, and I hope this will be useful in understanding some regressions. Thanks a lot for applying all suggestions!

PR #896, opened 2026-08-13
Remove unused Python imports and dependencies

### LINKED ISSUE #833: Check for unused Python dependencies
In some Python-based tutorials (mainly Nutils, FEniCS, and FEniCSx participants), we seem to be importing packages that are not directly used. For example:

- In `partitioned-heat-conduction-3d`, the `import dolfinx.mesh` seems to be unused.
- In many codes, the `from mpi4py import MPI` seems to be unused.

I guess that many of these originate from the codes that have been used as starting points. Removing any such unnecessary dependencies (also from the respective `requirements.txt`) would be nice.

Triggered by #714.

floze-the-genius:
## Summary

- remove unused Python imports across tutorial participants and test tooling
- drop `matplotlib` and `numpy` from the two participants that no longer use them
- regenerate the affected reference requirements with the repository tool
- preserve the `Axes3D` side-effect import needed by supported Matplotlib versions before 3.2

## Testing

- `uvx ruff check . --select F401`
- compile every changed Python file with `python3 -m py_compile`
- run the autopep8 and size pre-commit hooks on the changed files
- run `print_metadata.py`, `print_test_suites.py`, and `systemtests.py --help`
- check both updated reference requirements with `--check --fail-on-outdated`

The repository-wide `check-names` hook also ran, but reports pre-existing permalink errors in unchanged README files.

## Checklist

- [x] Added `changelog-entries/896.md`
- [x] Regenerated sibling `requirements-reference.txt` files for changed requirements

Fixes #833


elastic-tube-1d/plot-vtk.py
```diff
@@ -2,10 +2,9 @@
 
 import vtk
 import numpy as np
-import os
 import sys
 import matplotlib.pyplot as plt
-from mpl_toolkits.mplot3d import Axes3D
+from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 - registers projection='3d' on Matplotlib <3.2
```
MakisH, line 7: I would argue that matplotlib 3.2 is already too old, but we don't have version restrictions for this script yet, so that's all good for now.

MakisH (approved):
Thank you for the contribution! I guess that #833 is now sufficiently addressed.

The system tests pass, and I looked over all files, I don't see anything wrong: all removed imports are not mentioned anywhere in the file.

> The repository-wide `check-names` hook also ran, but reports pre-existing permalink errors in unchanged README files.

Where do you see that? `check-names` passes on the CI.

It would be useful to have such a check for unused imports as a pre-commit hook.

PR #900, opened 2026-08-14
Move tools/tests/ to tests/

MakisH:
Related to #703.

I only did:

- `git mv tools/tests/ tests/`
- Find and replace `tools/tests/` with `tests/`
- Adjusted the pre-commit check `check.sh` to ignore the `tests/` for the permalink check.

Looking at the diff, I also updated a few paths in the `Systemtest.py`, where `"tools","tests"` appeared, and a few more similar cases.

Note that `PRECICE_TOOLS_DIR` (pointing to `../..`) was removed and replaced by the already existing `PRECICE_TESTS_DIR` (pointing to `..`). All these path variables have now one layer less (and `PRECICE_TESTS_DIR` points to `..`).

Testing:

- Quickstart (passes): https://github.com/precice/tutorials/actions/runs/31827872141
   - Also tried `rerun-system-test.sh` from the archive, it works. 

I assume that other open PRs could be updated in a mostly fast-forward way. Only the `paths.py` does not seem to be recognized as a rename.

PR #904, opened 2026-08-15
Tools structure

### LINKED ISSUE #703: Mixed use of the tools/ directory
Our [`tools/`](https://github.com/precice/tutorials/tree/develop/tools) directory currently contains:

- Tools meant to be copied by users when they want to extract a tutorial (e.g., `log.sh`)
- Tools used by pre-commit (e.g., `check-size.sh`)
- System tests related tools (`tools/tests/`)

This is rather confusing for users. I would suggest splitting this into:

- `tools/`: Only for user-facing scripts
- `tests/`: For the system tests  

What do we then do with the scripts used by pre-commit? Alternatives:

- `checks/` (ugh)
- `pre-commit/` (clear for us, strange for someone that does not know what this is)
- keep them in `tools/`, but maybe in a subdirectory `tools/pre-commit`?

MakisH:
Closes #703 by moving the development-oriented tools out of the `tools/` directory, which is now only meant for runtime tools for users to copy.

While a better structure would rename these to, e.g., `runtime-tools/` or similar, I don't want to introduce such a massively breaking change.

At the same time, this repository is mainly meant for users. For this reason, I moved the respective tools to `.dev-tools`, which would keep them hidden by default.

This PR also:

- groups the tools together in `releasing/` and `linting/` (similarly to https://github.com/precice/precice/tree/develop/tools)
- extends the `releasing/README.md` to include `visualize-configs.sh`
- removes the executable flag from scripts that are only meant to be sourced (`openfoam-remove-empty-dirs.sh` and `cleaning-tools.sh`)
- renames `check.sh` to `check-filenames-permalinks.sh`, to make it more specific.

In #900, `tools/tests/` was already moved to `tests/`. That could even go to `.tests/`, to keep the list of directories strictly tutorials-related. That would at least help my file navigation a lot.

Any other ideas are welcome.


MakisH:
There has been no further feedback, and I have no better ideas. I am going to merge this soon.

My main point is that I want a user to copy `tools/` and run their simulations, having as few unrelated files in there as possible (thus also excluding `tools/dev/`).

I also want to make that a hidden directory, to keep it simpler for users, and on top of the files list for us.

I am going to leave `tests/` as-is, as this is already getting touched by many PRs and referenced by external sources.

PR #909, opened 2026-08-26
Use the high-order solid discretisation in perpendicular-flap/solid-solids4foam

### LINKED ISSUE #515: peprpendicular-flap solid-openfoam case needs tuning
In #507, while comparing all the solver options we have, it became apparent that the solid-openfoam (solidDisplacementFoam) gives very different results than everything else. Still, it is a nice technical test that only requires OpenFOAM. The solid-solids4foam case is mostly fine already.

I have tried the following:

- Varying the mesh size: I get completely different results
- Varying the `fvSolution/stressAnalysis/nCorrectors` (suggestion by @philipcardiff): I get a bit different results, but not enough.

Some pictures that might help someone pick it up later on:

tutorials-perpendicular-flap-displacement-openfoam-watchpoints

tutorials-perpendicular-flap-displacement-openfoam-convergence-watchpoints

For now, I just give up, and document that this is an outlier at the moment.

One case use the `plot-all-displacements.sh` script to generate these pictures (see #507).

philipcardiff:
Switches `perpendicular-flap/solid-solids4foam` to the high-order solid discretisation released in solids4foam v2.4, which brings the flap-tip displacement essentially onto the CalculiX / deal.II / FEniCS / DUNE curves.

Follow-up to the analysis in https://github.com/precice/tutorials/issues/515#issuecomment-3474219167, opened separately from #893 as discussed there. Note this does **not** close #515, which is about `solid-openfoam` (solidDisplacementFoam).

## Why

The current case is the second-largest outlier after `solid-openfoam`: its oscillation period is ~9% too short, so it drifts steadily out of phase and looks acceptable only over the first period. This is a discretisation error, not a version or solver-technology effect — re-running the original case config under solids4foam v2.4 / OpenFOAM v2606 gives no improvement over the stored v2.0 reference, and keeping the proposed configuration while disabling only the high-order residual makes things worse, not better.

| configuration | relL2 | period [s] | lag [s] | amp ratio |
|---|---|---|---|---|
| *consensus* (CalculiX/deal.II/FEniCS/DUNE) | – | 1.625 | +0.00 | 1.000 |
| the four of them individually | 0.6–2.0 % | | | 0.999–1.014 |
| solids4foam v2.0 / OF-v2012 (stored reference) | 47.7 % | 1.480 | −0.27 | 0.935 |
| solids4foam v2.4 / OF-v2606, original case config | 59.9 % | 1.440 | −0.37 | 0.903 |
| solids4foam v2.4 / OF-v2606, **high-order (p=3)** — this PR | **8.6 %** | **1.620** | **−0.03** | 1.076 |

relL2 is over the full 5 s of x-displacement, lag is by cross-correlation against the consensus over t ≥ 2 s, amplitude ratio is peak-to-peak. Runtime is unchanged in practice (~55 s for the solid participant, ~2.0 coupling iterations per time window).

It is a good match rather than a converged one: the amplitude is ~8% above the consensus and there is a residual waveform-shape difference of ~11% (versus 1.7–4.5% within the FEM group). I suspect the remainder comes from the low number of interface faces the fluid forces are mapped onto; a mesh study with the low-order solver does converge towards the other results, and I would expect the same here. For the purposes of this tutorial the results look acceptable as a demonstration.

## Changes

Case (`perpendicular-flap/solid-solids4foam`), no mesh and no preCICE config change:

- `constant/solidProperties` — `solutionAlgorithm PETScSNES`, a `highOrderCoeffs` block (`highOrderResidual true`, `highOrderJacobian false`, `polynomialOrder 3`, `faceStencilExtraCells 15`, radially-symmetric-exponential weighting with k=6), and an `alpha` momentum stabilisation.
- `system/fvSolution` — `D` solved with `solver petsc` driving a matrix-free `newtonls` SNES (`snes_mf` / `snes_mf_operator`), `lgmres` + `bjacobi`/`lu`, lagged preconditioner.
- `system/fvSchemes` — explicit per-term `laplacian`/`snGrad`/`interpolate` entries replaced by `default` entries, as the high-order path introduces terms the old explicit list does not name.

Supporting:

- `perpendicular-flap/README.md` — the case now needs solids4foam v2.4 or later built with PETSc; the note about v2.0 / OpenFOAM v2012 / OpenFOAM 9 / foam-extend no longer applies to it.
- `tests/dockerfiles/ubuntu_2404/Dockerfile` — base image bumped `v2.3-openfoam-v2412-latest` → `v2.4-openfoam-v2412-latest`. I verified that this image ships PETSc (`PETSC_DIR=/opt/petsc`) and that `libsolids4FoamModels.so` contains both the SNES and the high-order symbols.
- `tools/cleaning-tools.sh` — remove the `constant/*.withDefaultValues` dictionaries that solids4foam writes into the case, which `clean.sh` left behind.

## Still to do (happy for you to push onto this branch, @MakisH)

Reference results and plots need regenerating:

- `perpendicular-flap/reference-results/fluid-openfoam_solid-solids4foam.tar.gz` (used by the fieldcompare test via `tests/tests.yaml`)
- `reference-results/watchpoints/openfoam-solids4foam-v2404.log`, plus the `su2-`, `nutils-` and `fake-solids4foam` watchpoints
- the website plots from `plot-all-displacements.sh`

Two other notes:

- Part of the difference between my numbers and the stored reference may be the OpenFOAM v2606 behaviour change you pointed at (precice/openfoam-adapter#411); the *relative* comparisons in the table above were all run under the same version, so the conclusion should be unaffected.
- Unrelated pre-existing inaccuracy I noticed in the same README bullet: it says the geometrically non-linear solver is used by default, but this case has used `solidModel linearGeometryTotalDisplacement` since it was added in #286. I left it alone here — say the word and I will fix it in this PR.

## Environment

- OpenFOAM v2606, OpenFOAM-preCICE adapter, preCICE v3.4.1
- solids4foam `development` at `v2.4-4-g9edc0db3`, built with PETSc
- Serial runs; the proposed configuration reproduced bit-identically on a repeat

https://claude.ai/code/session_01CSURKmdsgTCHvWAQdXfTCo


changelog-entries/909.md
```diff
@@ -0,0 +1 @@
+- Switched the perpendicular-flap solid-solids4foam case to the high-order solid discretisation and the PETSc SNES solution algorithm, matching the FEM-based solid participants much more closely. This case now requires solids4foam v2.4 or later, built with PETSc.
```
MakisH:
```suggestion
- Switched the perpendicular-flap solid-solids4foam case to the high-order solid discretisation and the PETSc SNES solution algorithm, matching the FEM-based solid participants much more closely. This case now requires solids4foam v2.4 or later, built with PETSc [#909](https://github.com/precice/tutorials/pull/909).
```

MakisH (commented):
Thank you for the PR! I have not yet ran the case, but the PR is very clean.

For the plots, I can add a note that they are outdated (and referring to the v2404 distribution release). I wanted to re-generate all watchpoints and plots with the latest versions at some point (and maybe I can do that automatically this time). In the meantime, I can add an updated plot for solids4foam and refer to this PR.

> Unrelated pre-existing inaccuracy I noticed in the same README bullet

yes, please!

philipcardiff:
> > Unrelated pre-existing inaccuracy I noticed in the same README bullet
> 
> yes, please!

Done

PR #907, opened 2026-08-20
Fix per-component build arguments in system tests

PranjalManhgaye:
## Summary
- Resolve Docker build arguments per component from `components.yaml` instead of merging all parameters into one flat `params_to_use` dict
- Each Compose service gets its own `build_arguments` and `dockerfile_context` (from that component's `PLATFORM`)
- Global `--build_args` CLI overrides still apply to all components (backward compatible)
- Add unit tests for different `PLATFORM` / `PRECICE_REF` per component

Fixes #873

## Test plan
- [x] `cd tests && python3 -m unittest test_per_component_build_args.py`
- [x] Quickstart system test: Docker build + run pass locally

PR #910, opened 2026-08-26
Fix git ref resolution for default build arguments and adjust logging

PranjalManhgaye:
## Summary
- Restore git ref resolution for default `_REF` build arguments from `components.yaml`
- Regression introduced in #907: resolution only ran when args were passed via CLI
- Cache resolved refs by `(repository, ref)`; preserve per-component different refs

Follow-up to nightly failures reported in https://github.com/precice/tutorials/pull/907

## Test plan
- [x] `python3 -m py_compile tests/systemtests/Systemtest.py`
- [x] Local smoke: fluid-fake/solid-fake, rust-bindings, SU2 adapter - refs resolve to 40-char SHAs
- [ ] Nightly system tests pass on GHA

### DIFF
--- a/tests/systemtests/Systemtest.py
+++ b/tests/systemtests/Systemtest.py
@@ -290,7 +290,9 @@ class Systemtest:
         Previously, this function was also checking if all required parameters were provided, and was raising exceptions for parameters not provided and not having a default value. This check made adding optional parameters with empty defaults (e.g., the TUTORIALS_PR) complicated, and it was removed.
         """
         provided_arguments = self.arguments.arguments
+        explicit_cli_keys = frozenset(provided_arguments.keys())
         self.build_arguments_by_component = {}
+        resolved_ref_cache: Dict[Tuple[str, str], str] = {}
 
         for case in self.case_combination.cases:
             component = case.component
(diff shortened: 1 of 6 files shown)


MakisH:
Testing in https://github.com/precice/tutorials/actions/runs/32960301735/job/98150960461

The logging is now much longer (one group of entries per component). Maybe we should add a log separator for each component group and generally organize this part of the logs a bit nicer.

MakisH:
@PranjalManhgaye the same tests seem to be failing: https://github.com/precice/tutorials/actions/runs/32960301735/job/98150960461

MakisH:
As discussed, the remaining issues would be solved with:

- `fluid-fake`/`solid-fake`: Set `component: python-bindings` instead of `component: bare` of these cases in the `perpendicular-flap/metadata.yaml`.
- SU2: in the `components.yaml`, define `PYTHON_BINDINGS_REF` for the `su2-adapter`.

Both look like bugs that were only surfaced now but were always there.

PranjalManhgaye:
@makish please check.

MakisH:
Checking on https://github.com/precice/tutorials/actions/runs/33090348223/job/98581198940

Logging is still a bit too verbose. I would rather add a first line (along the lines "substituting default arguments and resolving git refs...") and move the rest to the debug mode.

PranjalManhgaye:
@makish, I have seen the tests are already passing here i have pushed a commit regarding verbosity of logging.

tests/systemtests/Systemtest.py
```diff
@@ -290,39 +290,54 @@ def __init_args_to_use(self):
         Previously, this function was also checking if all required parameters were provided, and was raising exceptions for parameters not provided and not having a default value. This check made adding optional parameters with empty defaults (e.g., the TUTORIALS_PR) complicated, and it was removed.
         """
         provided_arguments = self.arguments.arguments
+        explicit_cli_keys = frozenset(provided_arguments.keys())
         self.build_arguments_by_component = {}
+        resolved_ref_cache: Dict[Tuple[str, str], str] = {}
+
+        logging.info(
```
MakisH:
```suggestion
        logging.debug(
```

MakisH (commented):
Already better, but the logs are still a bit long for the nightly tests. I meant having only one line for all of the default substitutions, but we don't really need any at all.

MakisH (approved):
I adjusted a bit more the logging: I moved the logging in `sources.py` to debug mode, I moved the components arguments resolution to debug mode, and I added an opening message in the logging.

All tests are passing, this is now good to go!

PR #911, opened 2026-08-26
Define each test case as a reusable test suite

MakisH:
I often find it difficult to run a specific test case only. At the same time, in #789, we discussed a pre-processing step with @PranjalManhgaye, which would need each test case to be defined individually.

This PR pre-defines each test case combination as a reusable test suite, and refers to these in the test suite of each tutorial.

@PranjalManhgaye would that be enough for you?

I started refactoring part of the file manually, and then asked AI to finish the rest. Looking directly at the modified file (or comparing specific entries) will be easier than looking at the complete diff.

I have done a sanity check on all entries, and I am testing a few cases in https://github.com/precice/tutorials/actions/runs/32973715982

### DIFF
--- a/.github/workflows/generate-reference-results.yml
+++ b/.github/workflows/generate-reference-results.yml
@@ -106,9 +106,8 @@ jobs:
         if [ -z "${{ inputs.suites }}" ]; then
           git add ./*/*/*.tar.gz
           git add ./*/*.tar.gz
-          git add ./*/reference-results/*.metadata
         else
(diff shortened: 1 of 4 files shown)


PranjalManhgaye:
Yes, per-case suites are enough for #789 preprocessing. I’ll rebuild #829 on top of this with a per-tutorial GHA matrix + prepare-matrix job. Approve from my side.

PranjalManhgaye (approved):

PR #872, opened 2026-07-10
Update perpendicular-flap/solid-solids4foam

MakisH:
In recent solids4foam version (e.g., v2.3), the [perpendicular-flap tutorial](https://precice.org/tutorials-perpendicular-flap.html) does not run, as some of the configuration options have changed. Related issue: https://github.com/solids4foam/solids4foam/issues/279

I don't know much about the configuration, but blindly trying different candidates, I got it to work again.

Originally, the case was using the now removed `gradSchemes` option `extendedLeastSquares`. I tried with `leastSquares` and with `leastSquaresS4f`, and I am comparing to the old reference results in this figure:

openfoam-solids4foam

There is clearly room for improvement to match the previous state. @philipcardiff @iBatistic what else could I try?

The original case was already not tuned enough to match the other available setups (see the second figure on the [tutorial post-processing section](https://precice.org/tutorials-perpendicular-flap.html#looking-closer)). I am not concerned about that at the moment (I mainly don't want to make it work again), but we could give it another try. Bringing CalculiX into the picture (current and archived reference results):

openfoam-solids4foam-calculix

Current setup:

- solids4foam from the [current develop branch](https://github.com/solids4foam/solids4foam/tree/a2e43c33e91ec9e47ec815ccfcec0eaf03711643) (> v2.3)
- OpenFOAM v2512
- OpenFOAM adapter from the [current develop branch](https://github.com/precice/openfoam-adapter/tree/c39634e1d0ba62dfb4839acf3bab5133e58d8e1b)
- CalculiX v2.20
- preCICE v3.4.1

Reference setup: solids4foam v2.0, OpenFOAM v2012, OpenFOAM adapter v1.3.1, CalculiX v2.20, preCICE v3.1.1.

Plotting script and data: [solids4foam-debugging.tar.gz](https://github.com/user-attachments/files/29891108/solids4foam-debugging.tar.gz)

## Checklist

- [ ] I added a summary of any user-facing changes (compared to the last release) in the `changelog-entries/<PRnumber>.md`.

PR #912, opened 2026-08-26
Add a solids4foam solid participant to breaking-dam-2d

philipcardiff:
## Summary

Adds a **solids4foam** solid participant to the `breaking-dam-2d` tutorial, following the approach introduced in #909 for `perpendicular-flap/solid-solids4foam`.

Unlike the perpendicular flap, this flap undergoes genuinely large deformation (the tip reaches ~0.19 m, i.e. ~19% of the flap height, by t = 1 s), so this case uses the **nonlinear-geometry total-Lagrangian** solid model with a St. Venant-Kirchhoff material, matching the `NLGEOM` formulation of the CalculiX case, together with the **high-order (polynomial reconstruction) discretisation** and the **PETSc SNES** solution algorithm:

```
solidModel     nonLinearGeometryTotalLagrangianTotalDisplacement;

nonLinearGeometryTotalLagrangianTotalDisplacementCoeffs
{
    solutionAlgorithm PETScSNES;
    highOrderCoeffs { highOrderJacobian false; highOrderResidual true; displacement { polynomialOrder 3; ... } }
    stabilisation   { momentum { type alpha; scaleFactor 1.0; } }
}
```

The flap geometry, material properties (E = 4e7, nu = 0, rho = 3000) and time-step size match the CalculiX case, and gravity is not applied to the solid, so the two solid participants are directly comparable.

## Results

Flap-tip x-displacement, compared against the CalculiX reference results in `breaking-dam-2d/reference-results/` (which cover the first 0.2 s):

| t [s] | CalculiX | solids4foam | rel. diff |
|---|---|---|---|
| 0.02 | 3.785e-06 | 3.423e-06 | -9.6% |
| 0.04 | 2.236e-05 | 2.122e-05 | -5.1% |
| 0.06 | 7.521e-05 | 7.516e-05 | -0.07% |
| 0.10 | 2.350e-04 | 2.342e-04 | -0.34% |
| 0.14 | 4.019e-04 | 4.016e-04 | -0.07% |
| 0.16 | 5.088e-04 | 5.135e-04 | +0.91% |
| 0.20 | 7.094e-04 | 7.092e-04 | -0.03% |

Agreement is below 1% from t ≈ 0.06 s onwards. The larger relative differences at t = 0.02–0.04 s are on displacements of order 1 µm, where the absolute difference is ~0.3 µm.

The full 1 s simulation runs cleanly: ~2 min wall-clock (serial solid), "The momentum equation converged in all time-steps", and at most 6 implicit coupling iterations per time window.

Tested with OpenFOAM v2412, preCICE 3.4.1 and the OpenFOAM adapter at `develop`.

## Notes

* This case requires solids4foam v2.4 or later, built with PETSc, and the OpenFOAM-preCICE adapter v1.2.0 or later.
* The reference-results tarball for `fluid-openfoam_solid-solids4foam` still needs to be generated; the `tests/tests.yaml` entry already points at the expected path.
* The changelog entry is `changelog-entries/912.md`.

## Proposal: extending this to more tutorials

While preparing this, I surveyed the other tutorials with a deformable solid to see where else a solids4foam participant could be added. The blocker is not solids4foam, but the OpenFOAM adapter: for a `solverType solid` participant it currently supports only *reading `Force`* (at `faceCenters`) and *writing `Displacement`*. Two related code paths are stubs that raise an adapter error:

* `FSI/ForceBase.C` — `readFromBuffer()`: *"Reading stresses is not supported."* (openfoam-adapter issue precice/openfoam-adapter#240)
* `FSI/DisplacementDelta.C` — `write()`: *"Writing displacementDelta is not supported."*

That leaves:

| Tutorial | Exchanged data | Status |
|---|---|---|
| `perpendicular-flap` | Force / Displacement | done (#872, #909) |
| `breaking-dam-2d` | Force / Displacement | this PR |
| `turek-hron-fsi3` | **Stress** / Displacement | needs Stress reading |
| `multiple-perpendicular-flaps` | **Stress-Upstream/Downstream** / Displacement | needs Stress reading |
| `perpendicular-flap-stress` | **Stress** / Displacement | needs Stress reading |
| `elastic-tube-3d` | Force / **DisplacementDelta** | needs DisplacementDelta writing |

Both gaps look very doable:

* **Stress reading** would mirror the existing `Force` reader, filling the boundary field of a user-named traction field (e.g. `nameStress`, alongside the existing `nameForce`), which a solids4foam `solidTraction` boundary condition can consume directly. This alone unlocks three 2D tutorials, including Turek-Hron FSI3 — a natural fit, since solids4foam ships its own Turek-Hron FSI3 benchmark and this would let the tutorial cross-check against the deal.II and Nutils participants.
* **DisplacementDelta writing** is essentially `D - D.oldTime()` on the interface patches, and unlocks `elastic-tube-3d`. The solid there is a simple annulus (r = 5→6 mm, length 50 mm), easily meshed with a four-block `blockMesh` O-grid. Being a medium-sized 3D case, we would use a hypre BoomerAMG preconditioner rather than the block-Jacobi/LU setup used for the small 2D cases. One thing to check there: `elastic-tube-3d/precice-config.xml` maps with `nearest-projection` onto `Solid-Mesh`, and a face-centred OpenFOAM interface mesh carries no connectivity, so preCICE would fall back to nearest-neighbour.

@MakisH — would you like us to propose those two adapter additions and the follow-up tutorial cases? Happy to open them as separate PRs (adapter first, then one tutorial PR per case) if that suits.

---

https://claude.ai/code/session_01SBFTMffUN7qDigrDQrPVb2


MakisH:
@philipcardiff thanks for the tutorial case! This already looks very clean, I will review it and run the case asap.

Happy to get solids4foam as a participant in other tutorials as well. Regarding the missing features, PRs are very welcome and needed.

Issue #417, opened 2023-11-30
Inconsistent name of preCICE exports


MakisH:
_Originally posted by @davidscn in https://github.com/precice/tutorials/pull/411#discussion_r1410373649_

>  I think `preCICE-output` refers to the exports, right? Our configuration seems a bit arbitrary throughout the tutorials in this regard
```
./flow-over-heated-plate-nearest-projection/precice-config.xml:48:    <!-- <export:vtk directory="preCICE-output" /> -->
./replay-mode/precice-config.xml:24:         <export:vtu directory="exports" />
./two-scale-heat-conduction/precice-config.xml:53:      <export:vtu directory="precice-exports" />
./quickstart/precice-config.xml:41:    <export:vtk directory="coupling-meshes" />
```

PR #915, opened 2026-08-28
Add a solids4foam solid participant to turek-hron-fsi3

philipcardiff:
> [!IMPORTANT]
> **Draft: blocked on precice/openfoam-adapter#417** ("Support reading Stress in solid participants"). This case cannot run until that adapter change is released, and the reference-results tarball cannot be generated until it reaches the system-test images. Marking ready for review once #417 is merged.

## Summary

Adds a **solids4foam** solid participant to `turek-hron-fsi3`, adapted from the `HronTurekFsi3` benchmark that solids4foam ships. This is a natural fit: solids4foam maintains its own version of this benchmark, so the tutorial gains a cross-check against the deal.II and Nutils participants.

* Neo-Hookean, nonlinear-geometry total-Lagrangian solid model, PETSc SNES, `alpha` stabilisation.
* Material properties match the deal.II case: E = 5.6 MPa, nu = 0.4, rho = 1000 kg/m³.
* 630-cell beam (105 x 6), patches renamed to `interface` / `fixed` / `frontAndBack` to match the repo's conventions.
* The `interface` patch uses the solids4foam `solidTraction` boundary condition, whose `tractionField` option consumes the `solidTraction` volVectorField that the adapter fills.

## Results

Full 15 s run (15000 time windows), fluid on 20 MPI ranks, solid serial. Flap-tip y-displacement over the settled limit cycle (t = 8 to 15 s), against the Turek-Hron FSI3 benchmark value of uy = 1.48 +/- 34.38 mm:

| | mean [mm] | amplitude [mm] | amplitude error |
|---|---|---|---|
| benchmark | 1.48 | 34.38 | — |
| this case (`polynomialOrder 3`) | **1.47** | **34.50** | **+0.35%** |

## Choice of polynomial order

Since the high-order (polynomial reconstruction) discretisation is a cost/accuracy knob, I ran the full 15 s simulation three times, changing only the solid discretisation order and holding everything else fixed:

| solid discretisation | amplitude [mm] | amplitude error | mean error | solid solver cost | end-to-end runtime |
|---|---|---|---|---|---|
| p = 1, default scheme (no `highOrderCoeffs`) | 32.59 | -5.2% | -2.2% | 1.0x | 1.00x |
| p = 2, `polynomialOrder 2` | 33.34 | -3.0% | -2.2% | 1.8x | 0.96x |
| **p = 3, `polynomialOrder 3` (used here)** | **34.50** | **+0.3%** | **-0.4%** | 3.2x | 1.03x |

The first row is the default scheme rather than `highOrderCoeffs` with `polynomialOrder 1`. Both are first order and give essentially the same answer, differing only through the stencil size, so only the default one is listed.

Accuracy improves monotonically with order. The solid solver is genuinely 3.2x more expensive at p=3 (solid CPU time 5023 s -> 15913 s), but the **end-to-end runtime is essentially unchanged**, because in this tutorial the fluid participant dominates and the extra solid work hides behind the coupling. Wall-clock times were 28594 s, 27315 s and 29345 s respectively — within the noise of other jobs sharing the node.

I therefore propose `polynomialOrder 3`, but the trade-off is documented in the README and switching is a one-line edit, so please say if you would rather ship a cheaper solid. Note that `polynomialOrder 2` is both cheaper *and* more accurate than no high-order reconstruction at all, so the plain low-order option is dominated.

One observation independent of the solid settings: all three variants give an oscillation frequency of 5.50-5.64 Hz against the benchmark's 5.3 Hz. Since polynomial order does not move it, that offset comes from the fluid side or the coupling rather than the solid discretisation.

## Constitutive law

Flagging a discrepancy that predates this PR and is worth a decision. This case uses a compressible neo-Hookean law (`neoHookeanElastic`), which matches the **deal.II** participant (`set Model = neo-Hookean`) and the `HronTurekFsi3` case that solids4foam ships. However, the original Turek-Hron benchmark specifies a **St. Venant-Kirchhoff** solid, and that is what the **Nutils** participant implements (`S_ij = lambda E_kk delta_ij + 2 mu E_ij`). So the three solid participants in this tutorial are not currently consistent with each other on the material model.

At the strain levels reached in FSI3 the two laws are very close, and the p = 3 result above lands within 0.35% of the benchmark amplitude, so this does not look like a source of error here. But if you would prefer the solids4foam case to follow the benchmark definition, `StVenantKirchhoffElastic` is a drop-in replacement in `constant/mechanicalProperties` with the same `E`/`nu` entries, and I am happy to re-run the 15 s case with it and post the comparison. Say the word and I will do that rather than leave it as a note in the README.

Tested with OpenFOAM v2606, preCICE 3.4.1, solids4foam (development) and the OpenFOAM adapter from precice/openfoam-adapter#417.

## Notes

* Requires solids4foam v2.4 or later **built with PETSc**, and an OpenFOAM-preCICE adapter with support for reading `Stress` in solid participants.
* The reference-results tarball still needs to be generated; the `tests/tests.yaml` entry already points at the expected path.
* This PR and #914 / the `multiple-perpendicular-flaps` PR each add an entry to the `solids4foam` test suite in `tests/tests.yaml`, so whichever merges last will need a trivial rebase there.

---

https://claude.ai/code/session_01SBFTMffUN7qDigrDQrPVb2


philipcardiff:
Quick note on the constitutive-law point raised in the description: rather than leave it as an open caveat, I am running the check now.

The full 15 s case is re-running with `StVenantKirchhoffElastic` in place of `neoHookeanElastic`, holding everything else fixed — same mesh, same `polynomialOrder 3`, same fluid decomposition, same adapter build — so the only variable is the material law. I will post the result as an extra row in the polynomial-order table: mean, amplitude, and error against the benchmark's 34.38 mm, alongside the existing neo-Hookean p = 3 numbers.

That should settle the question rather than caveat it. If StVK is meaningfully closer to the benchmark, I will switch this case over to it, since it is a one-word change in `constant/mechanicalProperties` and would then match both the benchmark definition and the Nutils participant. If the two laws agree to within the noise, that is worth documenting too — it would mean the neo-Hookean choice inherited here (and used by the deal.II participant) is harmless at these strain levels, and the README note can say so plainly.

Results in roughly eight hours.

---
🤖 Written by Claude Code on behalf of @philipcardiff, and reviewed before posting.

philipcardiff:
## St. Venant-Kirchhoff vs neo-Hookean: they agree

Results from the check promised above. Short version: the constitutive law does not matter here, so this case can keep `neoHookeanElastic` and the discrepancy with the benchmark definition is not a source of error.

The case was re-run with `StVenantKirchhoffElastic` in place of `neoHookeanElastic`, holding everything else fixed — same mesh, same `polynomialOrder 3`, same 20-rank fluid decomposition, same adapter build — so the material law is the only variable.

Flap-tip y-displacement over the settled limit cycle, t in [8, 15] s (38-39 oscillation periods), against the benchmark value uy = 1.48 +/- 34.38 mm:

| case | mean [mm] | amplitude [mm] | amplitude error | f [Hz] |
|---|---|---|---|---|
| p = 1, neo-Hookean | 1.45 | 32.59 | -5.20% | 5.64 |
| p = 2, neo-Hookean | 1.45 | 33.34 | -3.02% | 5.57 |
| p = 3, neo-Hookean | 1.47 | 34.50 | +0.35% | 5.50 |
| **p = 3, St. Venant-Kirchhoff** | **1.48** | **34.48** | **+0.28%** | 5.50 |
| benchmark | 1.48 | 34.38 | — | 5.30 |

The two laws differ by **0.06% in amplitude**, and StVK matches the benchmark mean exactly. That separation is two orders of magnitude below the effect of the discretisation order, and well below the remaining gap to the benchmark, so at FSI3 strain levels the laws are interchangeable.

This also speaks to the deal.II participant, which uses neo-Hookean as well: the benchmark specifies StVK and the Nutils participant implements it, but on this problem that difference is not measurable in the watchpoint data.

The oscillation frequency is unmoved at 5.50 Hz against the benchmark's 5.3 Hz, and is identical for both laws — reinforcing the note in the description that this offset originates in the fluid or the coupling rather than in the solid model.

On cost, the two laws are close to equivalent: solid `ExecutionTime` over the full run was 15490 s for StVK against 15913 s for neo-Hookean, about 3% less. (An earlier partial-run measurement suggested ~15%; that was contaminated by other jobs sharing the node, and the full-length figure supersedes it.)

Given this, my suggestion is to leave the case on `neoHookeanElastic` — it matches both the deal.II participant and the `HronTurekFsi3` case that solids4foam ships, so it is the least surprising choice for anyone comparing the three solid participants — and to keep the README note recording that the benchmark specifies StVK, now with the measurement behind it rather than an open caveat. Happy to switch it if you would rather the case follow the benchmark definition literally; it is a one-word change in `constant/mechanicalProperties`.

<details>
<summary>Note on an earlier terminated run</summary>

A first StVK attempt was killed at t = 9.592 s by a `SIGTERM` to the solid participant, with no solver error preceding it; the fluid then failed on a broken pipe from preCICE as collateral. I could not attribute the signal — `sacct` is disabled on this machine and `dmesg` is not readable by the account, so an OOM kill could be neither confirmed nor excluded, and slurm preemption is off. The run reported above is a complete, clean 15 s run (both participants exit 0) and reproduced the terminated run's trace bit-for-bit over their common interval, so the earlier interruption was external and leaves no doubt over these numbers.

</details>

---
🤖 Written by Claude Code on behalf of @philipcardiff, and reviewed before posting.

PR #925, opened 2026-09-15
Add COMPONENT_PR as arguments in components.yaml

MakisH:
To be able to work with PRs from forks, the `COMPONENT_PR` (e.g., `PRECICE_PR`) is being passed by the calling workflow. Since this is not a required argument, it was not declared as one in `components.yaml`.

#907 changed the build argument parsing from "forward everything the user defines" to "collect whatever each component needs". This ignores the extra arguments. While this is better from a security point of view, it reduces the flexibility that was useful in this case.

As a workaround, this PR sets for each `COMPONENT` a `COMPONENT_PR` argument, with an empty string as a default value. [Testing without calling from a workflow](https://github.com/precice/tutorials/actions/runs/35025196908/job/104570981565) one does not seem to complain. The `TUTORIALS_PR` is set in the `bare` component, as this is the base for everything else.

This should enable working with forks again, which was detected in the context of https://github.com/precice/precice/pull/2601.


MakisH:
Actually, since now every component uses its own list of arguments (and might be called from arbitrary repositories), I needed to add a `*_PR` for every `*_REF` in https://github.com/precice/tutorials/commit/ae87db64bf577d1c2815d521406ecdf99fdb752d.

PR #931, opened 2026-09-21
Add alya-small to the system tests

MakisH:
Adds alya-small to the system tests, as an example of a restricted external code and case.

How the code and case are handled as sensitive:

- Manually, the code is copied to the authorized runner
- The case itself is extracted from the code, extended with a `metadata.yaml`
- The Docker Compose file mounts the code directory as additional build context (only accessing the code it needs)
- The case is served as an archive via a local file URL
- The reference results are stored in a runner-specific directory. The `generate-reference-results.yaml` workflow will generate them but fail to upload to the tutorials repository (which is fine, as there is nothing to upload).
- The GitHub Actions workflows:
    - In the beginning, make the files available by moving them to the `tests/` directory
    - In the end, it excludes the respective directories from the uploaded build artifact. The preCICE log files are still uploaded, but for more details, one would need to connect to the runner.

Testing in https://github.com/precice/tutorials/actions/runs/35691184247

Note that the case is partitioned to use 4+3 MPI ranks, but the current runner can only offer 4 (I will try to get a bigger one). I am running the case with `--oversubscribe`, which makes everything significantly slower: the case needs ~30min for three coupling time windows.

We could also set up a dedicated runner for restricted cases, but that would raise the complexity a bit (we would need yet another GHA workflow that picks the specific runner).

## Checklist


uekerman (approved):
Looks good to me. You now don't check against the provided iterations and convergence file, but these are generated by the system tests in a first run. Correct?

MakisH:
The iterations file is checked bitwise against the generated one for these three coupling time windows. The exported interface is also checked, with a tolerance.

uekerman:
:+1:  But not the convergence file? For three time windows, the iterations file might not tell the complete story, but maybe sth for a different PR.

MakisH:
Opened an issue: https://github.com/precice/tutorials/issues/933

